Issue #250, opened 2023-03-23
[QA] Current estimates for gas payment are showing as Not Available


AigaD:
Noticed that for some transactions the gas payment fee estimates are showing as "Not available", not sure of what triggers this.

Preconditions:
Wallet is connected

Environment:
MacBook Pro, macOS Ventura 13.0.1, Firefox version 109.0.1 (64-bit)

Steps:
1. Prepare to send 0.213 CELO from Alfajores to BSC
2. Observe the Gas Payment section

Actual result: Occasionally the Gas payment section shows as Not Available where the current estimate should be calculated. Under the preview, the source chain gas estimate also is shown as Not Available.

Expected result: Source fee gas estimate is calculated and shown to the user.

Comments: Relevant GitHub ticket - https://github.com/wormhole-foundation/wormhole-connect/issues/7#issue-1559176429

Attachment:

https://user-images.githubusercontent.com/91047061/227193403-cdabb232-3e3d-41c6-b410-5396523576b1.mov

<img width="1426" alt="Screenshot 2023-03-23 at 13 41 56" src="https://user-images.githubusercontent.com/91047061/227193443-64bfc0b2-2846-49d5-a275-54202d546a17.png">
<img width="718" alt="Screenshot 2023-03-23 at 13 47 35" src="https://user-images.githubusercontent.com/91047061/227194544-5ee264ad-dbbb-48cd-b1d8-2baa7c1d59dc.png">

PR #442, opened 2023-04-26
Add gas estimates for mainnet

M-Picco:
I used the testnet estimates as a base, and updated according to the txs I saw on the mainnet explorers for each chain.


anondev2323 (approved):

anondev2323:
Closed #250

Issue #376, opened 2023-04-19
[bug] register token link


anondev2323:
When no tokens appear in the list (I believe this should only happen on Solana if the user has no balance in any of the tokens), it shows "Register a token" and the link doesn't work.  I don't think a link to token registration is the right prompt here

anondev2323:
Removed in #397

Issue #352, opened 2023-04-18
[QA] Sometimes Assets list is not shown if user with slow internet connection connects wallet and selects Network


tsadovska:
#### Scenario 1:
1. Connect Phantom wallet
2. Select Solana network as fast as possible
3. Open Assets list as fast as possible
4. Observe that Assets list is not loaded
5. Click Close button
6. Open Assets list again

#### Scenario 2:
Repeat Scenario 1 with slow internet connection. It is not necessary to do the steps quickly.

#### Actual results:
Text “No results” is displayed when the user reopens the Assets list. 

#### Expected results:
Assets list is shown.

#### Reproducibility:
2/6

#### Tested device:
MacBook Pro, macOS Monterey 12.6.5, Google Chrome Version 112.0.5615.121 (Official Build) (x86_64)

#### Attachments:
* Scenario 1

https://user-images.githubusercontent.com/10746654/232710216-77c34c15-c5e6-4296-8963-20381dabbfd7.mov

* Scenario 2 (Recording)

https://user-images.githubusercontent.com/10746654/232710255-604790c7-1eb6-4817-8c21-6ba35e4dcf6f.mov

* Scenario 2 (Network settings)

![Network settings](https://user-images.githubusercontent.com/10746654/232710508-20b86250-3d1e-4db6-b692-d815a0cdd0f0.png)

Issue #258, opened 2023-03-24
[QA] Balance is shown incorrectly in Assets list (Moonbase)


tsadovska:
#### Steps:
1. Connect Metamask wallets in ‘From’ and ‘To’ sections
2. Select Moonbase network in ‘From’ section
3. Select Alfajores network in ‘To’ section
4. Open Assets list
5. Observe CELO balance in Assets list
6. Select CELO asset
7. Observe CELO balance on Bridge screen
8. Open Assets list again

#### Actual results:
* Balance is shown incorrectly in Step 5. 
* Balance is shown correctly in Step 8. 

#### Expected results:
Balance is shown correctly in Step 5 and 8.

#### Tested device:
MacBook Pro, macOS Monterey 12.6.1, Chrome Version 111.0.5563.110 (Official Build) (x86_64)

#### Attachment:

https://user-images.githubusercontent.com/10746654/227503344-24cedeeb-b010-4105-ba22-5d019a0042d2.mov

Issue #242, opened 2023-03-21
[QA] Balance does not revert to 0 when switching to a different network


AigaD:
Preconditions: Wallet is connected

Environment:
MacBook Pro, macOS Ventura 13.0.1, Firefox version 109.0.1 (64-bit)

Steps:
1. Select Network at "From" section (e.g. Mumbai)
2. Select Asset (E.g. MATIC)
3. Observe the balance
4. Open the Select Network again and with to a different Network (e.g. Fantom)
5. Observe balance

Actual result: When switching away from network were the previously chosen Asset is not available, the asset window resets, but the balance is shown for the main token of the chosen Network.

Expected result: Asset and Balance windows reverts to empty when Network is switched.

Attachment:

https://user-images.githubusercontent.com/91047061/226525870-e4f3dab7-1816-4976-82c0-86c9be08aca3.mov

PR #451, opened 2023-04-27
[397] disable token modal and filter tokens

### LINKED ISSUE #397: [Product] Prevent opening asset list unless wallet is connected
If no wallet is connected in the FROM section, the asset list should not open when the user taps the Asset component.

When the list is open, it should show the tokens in the wallet.

To discuss: how should tokens that are not present in the config be displayed, if in any way? should we only display the tokens in the list?

For now: just prevent opening the asset list as decribed above and expand the list of supported tokens with USDC and others defined by @aakashthumaty

anondev2323:
Closes #397

### DIFF
--- a/wormhole-connect/src/views/Bridge/Modals/TokensModal.tsx
+++ b/wormhole-connect/src/views/Bridge/Modals/TokensModal.tsx
@@ -263,7 +263,7 @@ function TokensModal() {
     }
     const filtered = networkTokens.filter((t) => {
       const b = tokenBalances[t.symbol];
-      return b !== undefined && b !== null && b !== '0';
+      return b !== null && b !== '0';
     });
     setTokens(filtered);
   }, [fromNetwork, tokenBalances, networkTokens, walletAddr]);
--- a/wormhole-connect/src/views/Bridge/Modals/TokensModal.tsx
+++ b/wormhole-connect/src/views/Bridge/Modals/TokensModal.tsx
@@ -222,7 +222,7 @@ function TokensModal() {
     const getBalances = async (
       tokens: TokenConfig[],
       walletAddr: string,
-      chain: ChainName | ChainId,
+      chain: ChainName,
     ) => {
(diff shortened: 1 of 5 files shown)


wormhole-connect/src/views/Bridge/Modals/TokensModal.tsx
```diff
@@ -363,13 +362,6 @@ function TokensModal() {
           </Collapse> */}
         </div>
       </Scroll>
-
-      <div className={classes.register}>
-        <div className={classes.registerText}>Don't see your token?</div>
-        <a href="#" className={classes.registerLink}>
-          Register token
-        </a>
-      </div>
```
anondev2323, line 372: Really shouldn't prompt user to register tokens here, kind of a dumb ux. Can just show the "no results" for tokens with balances

wormhole-connect/src/views/Bridge/Modals/TokensModal.tsx
```diff
@@ -255,19 +255,18 @@ function TokensModal() {
...
-              tokenBalances[t.symbol] !== null,
-          )
-        : networkTokens;
+    if (!fromNetwork || !walletAddr) {
+      setTokens(networkTokens);
+      return;
+    }
+    const filtered = networkTokens.filter((t) => {
+      const b = tokenBalances[t.symbol];
+      return b !== null && b !== '0';
```
anondev2323, line 266: Should not filter out balances of `undefined`.  This means it just hasn't gotten the balance yet and will display a spinner next to that token while it is fetching.

M-Picco (approved):

anondev2323:
Closed #376

anondev2323:
I think this resolves #352 also

anondev2323:
Closes #258

anondev2323:
Closes #242

anondev2323:
Seems like clearing the balances on network change has a lot of downstream affects/improvements

PR #461, opened 2023-04-27
add: option for integration to specify default bridge parameters

### LINKED ISSUE #394: [Feature] Allow integrator to pre-fill bridge parameters
This is a request from a user (via @aakashthumaty).  They would like to specify the destination chain and address.  I think this is a pretty useful feature, and I would suspect other users might want to specify source/destination networks and token as well.

anondev2323:
Partially resolves #394

### DIFF
--- a/wormhole-connect-loader/package-lock.json
+++ b/wormhole-connect-loader/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@wormhole-foundation/wormhole-connect",
-  "version": "0.0.1-beta.3",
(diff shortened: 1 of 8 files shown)


M-Picco:
What should we do if the user specifies a token that is not supported by the source network (e.g. `ETH` on Avalanche)?

anondev2323:
It would be weird of an integrator to do that, but is a possibility.  I'll add some kind of check for that

M-Picco (approved):

PR #470, opened 2023-04-28
[#333] Disable controls when a transaction is in progress

### LINKED ISSUE #333: [QA] User should not be able to change transaction details after transaction was approved
#### Steps:

1. Connect Metamask wallets in ‘From’ and ‘To’ sections
2. Prepare to send 0.01 ETH from Goerli to Mumbai
3. Click “Approve and proceed with transaction” button
4. Change Asset/Network or Disconnect wallet

#### Actual result:
User is able to change transaction details after transaction was approved.

#### Expected result:
User should not be able to change transaction details after transaction was approved.

#### Tested device:
MacBook Pro, macOS Monterey 12.6.1, Version 112.0.5615.49 (Official Build) (x86_64)

#### Attachment:

https://user-images.githubusercontent.com/10746654/232028951-53fef871-fe66-4532-8632-c6382955c2f8.mov

M-Picco:
https://user-images.githubusercontent.com/11276821/235186336-f5c0b68b-ea25-4171-828e-b6f796f8b6cf.mov

Closes #333

### DIFF
--- a/wormhole-connect/src/views/Bridge/Inputs.tsx/From.tsx
+++ b/wormhole-connect/src/views/Bridge/Inputs.tsx/From.tsx
@@ -28,6 +28,7 @@ function FromInputs() {
     validations,
     fromNetwork,
     token,
+    isTransactionInProgress,
   } = useSelector((state: RootState) => state.transfer);
   const tokenConfig = token && TOKENS[token];
 
@@ -80,7 +81,7 @@ function FromInputs() {
       selected={selectedToken}
       error={!!(showErrors && validations.token)}
       onClick={openTokensModal}
-      editable
+      editable={!isTransactionInProgress}
       disabled={!fromNetwork || !wallet.address}
     />
   );
@@ -89,7 +90,7 @@ function FromInputs() {
   const amountInput = (
     <Input
       label="Amount"
-      editable
+      editable={!isTransactionInProgress}
       error={!!(showErrors && validations.amount)}
       onClick={focusAmt}
     >
@@ -102,6 +103,7 @@ function FromInputs() {
           step={0.1}
           onChange={handleAmountChange}
           onPause={validateAmount}
+          disabled={isTransactionInProgress}
         />
       ) : (
         <div>-</div>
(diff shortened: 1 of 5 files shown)


wormhole-connect/src/views/Bridge/Inputs.tsx/Inputs.tsx
```diff
@@ -181,7 +185,7 @@ function Inputs(props: Props) {
                       selected={selectedNetwork}
                       error={!!(showErrors && props.networkValidation)}
                       onClick={props.onNetworkClick}
-                      editable
+                      editable={!isTransactionInProgress}
```
anondev2323: Leave editable as true, and use `disabled={!isTransactionInProgress}`.  Maybe we should change `editable` prop to `filled` or `colorFilled`...

anondev2323 (changes requested):

wormhole-connect/src/views/Bridge/Inputs.tsx/From.tsx
```diff
@@ -82,17 +82,17 @@ function FromInputs() {
...
   );
 
   // amount input jsx
   const amountInput = (
     <Input
       label="Amount"
-      editable={!isTransactionInProgress}
       error={!!(showErrors && validations.amount)}
       onClick={focusAmt}
+      editable={!isTransactionInProgress}
```
anondev2323: Same here, leave editable and set `disabled={!isTransactionInProgress`

anondev2323 (changes requested):

wormhole-connect/src/views/Bridge/Inputs.tsx/From.tsx
M-Picco: The inputs would show like this (I've also set the editable from the token picker to always true:

image

I changed it initially, but then I thought it would cause more confusion than anything.

anondev2323: I want to keep the color fill in both asset and amount on the "From" section

anondev2323 (approved):

PR #473, opened 2023-04-28
[344] fix solana transfer token error

### LINKED ISSUE #344: [QA] “WalletSendTransactionError” appears when sending wrapped asset from Solana
#### Steps:

1. Connect Phantom wallet in 'From' section and Metamask wallet in 'To' section
2. Prepare to send 0.01 CELO from Solana to Mumbai
3. Click “Approve and proceed with transaction”

#### Actual results:

- Transaction is not started
- “Error sending transfer, please try again” is displayed for the user
- Error in Console:

```
WalletSendTransactionError: failed to send transaction: Transaction simulation failed: Error processing Instruction 1: custom program error: 0x0
    at u.<anonymous> (main.ebfc4165.js:2:16924541)
    at l (main.ebfc4165.js:2:15854034)
    at Generator.<anonymous> (main.ebfc4165.js:2:15855395)
    at Generator.throw (main.ebfc4165.js:2:15854397)
    at n (main.ebfc4165.js:2:15847811)
    at c (main.ebfc4165.js:2:15848050)
```

#### Expected results:
Transaction is processed without errors.

#### Notes:
Issue is also reproducible when:

- User sends WGLMR from Solana no Alfajores
- User sends CELO from Solana to Fuji

Issue is not reproducible when:

- User sends CELO from Solana to Alfajores
- User sends WGLMR from Solana to Moonbase

#### Tested device:
MacBook Pro, macOS Monterey 12.6.5, Google Chrome Version 112.0.5615.121 (Official Build) (x86_64)

#### Attachments:
* Recording

https://user-images.githubusercontent.com/10746654/232504831-7bcb8075-63ac-4607-8ad2-d894491260fe.mov

* Screenshot

Screenshot 2023-04-17 at 16 58 06

anondev2323:
Closes #344

Need to verify 1 small thing first.  I think whenever a token originated on solana you need to use `transferNativeSol`, not just with SOL)

### DIFF
--- a/sdk/src/contexts/solana/index.ts
+++ b/sdk/src/contexts/solana/index.ts
@@ -276,7 +276,7 @@ export class SolanaContext<
   }
 
   private async transferFromSolana(
-    senderAddress: string,
+    senderAddress: PublicKeyInitData,
     amount: bigint,
     recipientChain: ChainId | ChainName,
     recipientAddress: Uint8Array | Buffer,
@@ -352,7 +352,7 @@ export class SolanaContext<
     const relayerFeeBN = relayerFee ? BigInt(relayerFee) : undefined;
     const amountBN = BigNumber.from(amount).toBigInt();
 
-    if (token === NATIVE) {
+    if (token === NATIVE || token.chain === 'solana') {
       return await this.transferNativeSol(
         senderAddress,
         amountBN,
@@ -363,12 +363,8 @@ export class SolanaContext<
         'finalized',
       );
     } else {
-      const destTokenAddr = await destContext.mustGetForeignAsset(
-        token,
-        recipientChain,
-      );
       const formattedTokenAddr = arrayify(
-        destContext.formatAddress(destTokenAddr),
+        destContext.formatAddress(token.address),
       );
       const solTokenAddr = await this.mustGetForeignAsset(
         token,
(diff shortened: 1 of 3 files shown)


sdk/src/contexts/solana/index.ts
```diff
@@ -352,7 +352,7 @@ export class SolanaContext<
     const relayerFeeBN = relayerFee ? BigInt(relayerFee) : undefined;
     const amountBN = BigNumber.from(amount).toBigInt();
 
-    if (token === NATIVE) {
+    if (token === NATIVE || token.chain === 'solana') {
```
M-Picco: Should this condition also check that `token.tokenId` is not set?

M-Picco:
> Need to verify 1 small thing first. I think whenever a token originated on solana you need to use `transferNativeSol`, not just with SOL)

I think `transferNativeSol` is only for SOL. Compared to `transferFromSolana` it doesn't accept a `mintAddress` (i.e. token address) as an argument (see [here](https://github.com/wormhole-foundation/wormhole/blob/main/sdk/js/src/token_bridge/transfer.ts#L352))

anondev2323:
Yeah good point.  I vaguely remembered a conversation I had about `transferNativeSol` so I wanted to check it out.  But you're right, it doesn't even take in the parameter so I'm sure that's not the case

M-Picco (approved):

PR #475, opened 2023-04-28
Started Sui support

kev1n-peters:


sdk/src/contexts/eth/index.ts
```diff
@@ -297,7 +297,7 @@ export class EthContext<T extends WormholeContext> extends RelayerAbstract {
       const tokenAddr = await this.mustGetForeignAsset(token, sendingChain);
       await this.approve(sendingChain, bridge.address, tokenAddr, amountBN);
       const v = await bridge.transferTokensWithPayload(
-        destContext.parseAddress(tokenAddr),
+        destContext.parseAddress(tokenAddr), // TODO: shouldn't this be the sending context (possible bug)?
```
anondev2323: Yes, that's right actually. This method isn't in use right now, I think last time we talked, we discussed not having a separate send/sendWithPayload.  Should just add a payload parameter back to `send`

sdk/src/contexts/abstracts/tokenBridge.ts
```diff
@@ -38,6 +38,9 @@ export abstract class TokenBridgeAbstract {
   protected abstract formatAddress(address: string): any;
   protected abstract parseAddress(address: any): string;
 
+  protected abstract formatAssetAddress(address: string): Promise<any>;
+  protected abstract parseAssetAddress(address: any): Promise<string>;
```
anondev2323, line 42: Is it different for account addresses?

wormhole-connect/src/store/transfer.ts
```diff
@@ -14,7 +14,12 @@ export const formatBalance = (
   token: TokenConfig,
   balance: BigNumber | null,
 ) => {
-  const decimals = chain === 'solana' ? token.solDecimals : token.decimals;
+  const decimals =
```
M-Picco, line 17: We should change the decimals to a map indexed by chain or context type

sdk/src/wormhole.ts
```diff
@@ -190,7 +196,7 @@ export class WormholeContext extends MultiProvider<Domain> {
     recipientAddress: string,
     relayerFee?: string,
     payload?: any,
-  ): Promise<ContractReceipt | Transaction> {
+  ): Promise<ContractReceipt | Transaction | TransactionBlock> {
```
anondev2323, line 199: Really should implement the type parameters, like Martin showed us in the wallet aggregator sdk

wormhole-connect/src/config/mainnet.ts
```diff
@@ -92,6 +92,17 @@ export const MAINNET_NETWORKS: NetworksConfig = {
...
+  sui: {
+    ...chains.sui!,
+    displayName: 'Sui',
+    explorerUrl: 'https://explorer.sui.io/',
+    explorerName: 'Sui Explorer',
+    gasToken: 'SUI',
+    chainId: 0,
+    icon: Icon.SUI,
+    automaticRelayer: false, // TODO: true
+    maxBlockSearch: 0, // TODO: what is this?
```
anondev2323: It's for searching for the Redeem event for automatic transfers.  RPCs usually have a limit of how far back you can search

wormhole-connect/src/config/mainnet.ts
```diff
@@ -103,6 +114,7 @@ export const MAINNET_TOKENS: { [key: string]: TokenConfig } = {
     color: '#62688F',
     decimals: 18,
     solDecimals: 8,
+    suiDecimals: 8,
```
anondev2323, line 118: I know you brought this up earlier, maybe refactor to:
```
decimals: {
  default: 18,
  solana: 8
  sui: 8
}
```
Or rather... maybe we should add a method that figures this out programmatically based on network and token

anondev2323: I like the second option

wormhole-connect/src/sdk/index.ts
```diff
@@ -164,32 +170,45 @@ export const sendTransfer = async (
...
+      wh.registerProviders();
+      return solTx;
+    } else if (fromChainId === MAINNET_CHAINS.sui) {
+      const response = await signSuiTransaction(
+        tx as unknown as TransactionBlock,
+        TransferWallet.SENDING,
+      );
+      wh.registerProviders();
+      return response;
+    } else {
```
anondev2323: We should figure out a system here.  It's going to get bloated as we add in more contexts

sdk/src/contexts/abstracts/tokenBridge.ts
kev1n-peters: Yes, Sui account addresses are 32 byte hex strings so we don't even need to parse/format them since they're already "formatted". But assets/tokens are different - e.g. the gas token on Sui is `0x2::sui::SUI` and to convert it to a wormhole address (and vice versa) we have to look it up from the Sui token bridge contract unfortunately (that's why these methods are `async`)

wormhole-connect/src/sdk/index.ts
```diff
@@ -133,11 +133,11 @@ export const getRelayerFee = async (
   destChain: ChainName | ChainId,
   token: string,
 ) => {
-  const EthContext: any = wh.getContext(destChain);
+  const context: any = wh.getContext(sourceChain);
```
anondev2323, line 136: Are you certain this should be the source chain?

kev1n-peters: Yeah, we should be getting the relayer fee from the source chain since it's what gets set on the VAA

wormhole-connect/src/store/transfer.ts
kev1n-peters: Agreed, we will refactor this in the future.

wormhole-connect/src/sdk/index.ts
kev1n-peters: Agreed, we will refactor this in the future.

anondev2323 (approved):
🌈

PR #479, opened 2023-04-30
Sui wallet initialization

anondev2323:


wormhole-connect/src/views/WalletModal.tsx
```diff
@@ -89,20 +89,41 @@ const WALLETS: { [key: string]: WalletData } = {
     isReady: getReady(wallets.solana.solflare),
   },
 };
-const getWalletOptions = async (chain: ChainConfig) => {
+
+let suiWalletOptions = {};
+
+const fetchSuiOptions = async () => {
+  const suiWallets = await getSuiWallets({ timeout: 0 });
```
anondev2323, line 96: I thought about checking if this is already defined. However, I think a user might try to install a wallet if nothing pops up and come back

M-Picco: If I'm not mistaken, most wallets inject their scripts when the page loads, so the user would have to reload anyways. We can keep it though, since there might be other cases though (some extension scripts can be loaded at install or runtime time if I recall correctly).

kev1n-peters (approved):

PR #493, opened 2023-05-01
Fix wallet detection for uninstalled wallets

M-Picco:
The check missed the `NotDetected` (i.e. not installed) state

### DIFF
--- a/wormhole-connect/src/views/WalletModal.tsx
+++ b/wormhole-connect/src/views/WalletModal.tsx
@@ -53,7 +53,7 @@ const useStyles = makeStyles()((theme) => ({
 
 const getReady = (wallet: Wallet) => {
   const ready = wallet.getWalletState();
-  return ready !== WalletState.Unsupported;
+  return ready !== WalletState.Unsupported && ready !== WalletState.NotDetected;
 };
 
 type WalletData = {


wormhole-connect/src/views/WalletModal.tsx
```diff
@@ -53,7 +53,7 @@ const useStyles = makeStyles()((theme) => ({
 
 const getReady = (wallet: Wallet) => {
   const ready = wallet.getWalletState();
-  return ready !== WalletState.Unsupported;
+  return ready !== WalletState.Unsupported && ready !== WalletState.NotDetected;
```
anondev2323, line 56: If I remember correctly, sometimes this caused Solflare and Phantom to not show up in the list.  I think if you weren't signed in they wouldn't show up

anondev2323: Maybe I was using the wrong value though

M-Picco: There are 4 states: `Installed`, `Loadable`, `NotDetected`, `Unsupported`. There's an additional layer beyond installed/not installed; for example: browser wallets would not be supported in other environments other than a browser, so they would yield Unsupported instead of NotDetected.

M-Picco:
Closed since PR #497 will cover this

PR #495, opened 2023-05-01
[387] update error message to get user unstuck

### LINKED ISSUE #387: [enhancement] Switch to Manual for low-value transfers
Switch to Manual if amount is too low to cover automatic fees.  The user can get stuck if they enter an amount that is too small for automatic but don't have a high enough balance to cover fees

anondev2323:
Closes #387

### DIFF
--- a/wormhole-connect/src/utils/transferValidation.ts
+++ b/wormhole-connect/src/utils/transferValidation.ts
@@ -81,8 +81,6 @@ export const validateAmount = (
   if (paymentOption === PaymentOption.MANUAL) return '';
   if (!minAmt) return '';
   if (amount < minAmt) return `Minimum amount is ${minAmt}`;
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/utils/transferValidation.ts
```diff
@@ -81,8 +81,6 @@ export const validateAmount = (
   if (paymentOption === PaymentOption.MANUAL) return '';
   if (!minAmt) return '';
   if (amount < minAmt) return `Minimum amount is ${minAmt}`;
-  if (amount + minAmt > b)
-    return 'Amount plus estimated fees exceeds the wallet balance';
```
anondev2323, line 85: Double check, but I think this is just wrong.  It takes fees out of the amount, not in addition

M-Picco (approved):

PR #497, opened 2023-05-01
[WIP, do not merge] Add solana wallets supported by portal bridge

### LINKED ISSUE #378: [enhancement] hide wallets that could not be used
For example, if an integration doesn't use Solana we should not show solana wallets in the modal. The same would be true for other network/context specific wallets.

M-Picco:
Also, use icons supplied by the wallet aggregator sdk wallets.

Also closes #378 

image


wormhole-connect/src/icons/WalletIcons.tsx
```diff
@@ -16,241 +17,23 @@ const useStyles = makeStyles<{ size: number }>()((theme, { size }) => ({
...
-    }
-    default: {
-      return <div style={{ height: `${size}px`, width: `${size}px` }}></div>;
-    }
-  }
+  const { wallet } = props;
+
+  return wallet ? (
+    <div className={classes.container}>
+      <img className={classes.icon} src={wallet.getIcon()} alt={wallet.getName()} />
```
anondev2323: Could actually just do:
```
return (
  <div className={classes.container}>
    {wallet && <img className={classes.icon} src={wallet.getIcon()} alt={wallet.getName()} />}
  </div>
)
```

wormhole-connect/src/views/WalletModal.tsx
```diff
@@ -62,40 +62,28 @@ type WalletData = {
...
-    isReady: getReady(wallets.solana.solflare),
-  },
+
+const mapWallets = (wallets: Record<string, Wallet>, type: WalletType): WalletData[] => {
+  return Object.values(wallets).map((wallet) => ({
+    name: wallet.getName(),
+    wallet,
+    type,
+    isReady: getReady(wallet),
+  }));
```
anondev2323, line 72: Nice, much cleaner/easy to maintain

anondev2323:
#378

wormhole-connect/src/icons/WalletIcons.tsx
M-Picco: I noticed the component is always used with a wallet prop, so no need to consider this case inside this component I think

anondev2323:
Screen Shot 2023-05-02 at 2 50 18 PM

anondev2323 (approved):

anondev2323:
Changes merged in #562

PR #504, opened 2023-05-01
fix: bug with alert component

anondev2323:
Alert was being cleared and not reset.  I found there is an `unmountOnExit` option which seems to do the trick


wormhole-connect/src/components/AlertBanner.tsx
```diff
@@ -35,18 +35,14 @@ function AlertBanner(props: Props) {
   const { classes } = useStyles();
   const [alertContent, setAlertContent] = useState(props.content);
```
M-Picco: Would this state hook still be needed? You could use `props.content` directly now that you don't need to clear the value.

kev1n-peters (approved):

PR #520, opened 2023-05-02
refactor: transaction signing

anondev2323:


anondev2323:
Cherry-picked onto a fresh branch #522

PR #536, opened 2023-05-03
Remove hard coded gasLimit and apply user supplied overrides

### LINKED ISSUE #533: [Bug] Hard coded gas limit of 250K is insufficient for some transfers

 - https://sui-mainnet--wormhole-connect.netlify.app/ as of 5/2 9:24PM PT
 - Transfer SUI from Sui -> Avalanche on mainnet with manual transfer
 - Avalanche side claim fails (see that wallet is set to 250K gas limit):
   - https://snowtrace.io/tx/0x9937a62a515d87bc69258d446c1232f9a63968564bda8cc2ce37808593a19ad1
   - https://snowtrace.io/tx/0x0dfc15264b1c30ccb28be28d315c0c209c1c8f8f6be30949e9d560ee31b2c07f
 - When gas limit is increased to 500K, claim succeeds: https://snowtrace.io/tx/0x5585ce53538a375d396ea65f9b88a6b99eb9f2ed7777bb3ff16aff6f0eb81815

- Gas limit should be inferred by wallet since claims can require a >250K gas limit transaction

M-Picco:
Closes #533

### DIFF
--- a/sdk/src/contexts/eth/index.ts
+++ b/sdk/src/contexts/eth/index.ts
@@ -114,7 +114,7 @@ export class EthContext<T extends WormholeContext> extends RelayerAbstract {
     contractAddress: string,
     token: string,
     amount?: BigNumberish,
-    overrides?: any,
+    overrides: PayableOverrides & { from?: string | Promise<string> } = {},
   ): Promise<ethers.ContractReceipt | void> {
     const signer = this.context.getSigner(chain);
     if (!signer) throw new Error(`No signer for ${chain}`);
@@ -136,7 +136,7 @@ export class EthContext<T extends WormholeContext> extends RelayerAbstract {
       const tx = await tokenImplementation.approve(
         contractAddress,
         approveAmount,
-        // overrides,
+        overrides
       );
       await tx.wait();
     }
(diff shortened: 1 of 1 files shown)


sdk/src/contexts/eth/index.ts
```diff
@@ -114,7 +114,7 @@ export class EthContext<T extends WormholeContext> extends RelayerAbstract {
     contractAddress: string,
     token: string,
     amount?: BigNumberish,
-    overrides?: any,
+    overrides: PayableOverrides & { from?: string | Promise<string> } = {},
```
anondev2323, line 117: Are we passing in any gas limits? These send functions fail pretty often with "Cannot estimate gas" if it's not set

kcsongor: should we just set it to 400k instead?

evan-gray: I don't see why it should fail. I've not seen that.

M-Picco: If the aim is for this to be an sdk which other developers eventually use, shouldn't overrides be left for the app/dev to set?

anondev2323: Yes, I'm saying we should set this from the app `wormhole-connect` not inside the sdk

anondev2323: I don't know why we run into this but you don't, but the only fix I've found is to supply a gas limit.  Maybe it's an rpc thing?

PR #566, opened 2023-05-04
Generic types for sdk contexts

M-Picco:
Add generic types to contexts. Infer result types for the send, sendWithRelay, and redeem operations instead of manually defining it in the base wormhole context.

### DIFF
--- a/sdk/src/contexts/abstracts/relayer.ts
+++ b/sdk/src/contexts/abstracts/relayer.ts
@@ -2,7 +2,10 @@ import { BigNumber, BigNumberish } from 'ethers';
 import { TokenId, ChainName, ChainId } from '../../types';
 import { TokenBridgeAbstract } from './tokenBridge';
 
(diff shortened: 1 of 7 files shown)


sdk/src/contexts/abstracts/relayer.ts
```diff
@@ -2,7 +2,10 @@ import { BigNumber, BigNumberish } from 'ethers';
 import { TokenId, ChainName, ChainId } from '../../types';
 import { TokenBridgeAbstract } from './tokenBridge';
 
-export abstract class RelayerAbstract extends TokenBridgeAbstract {
+export abstract class RelayerAbstract<
+  SendResult,
+  RedeemResult,
```
anondev2323: Can we just have 1 type `TransactionResult`? I think it's the same type for both

M-Picco: You are right, it doesn't make sense to have a different one right now. I'll update it.

anondev2323 (approved):

PR #584, opened 2023-05-08
isTransferLimitedWarning

anondev2323:
(Ignore that the details in the error don't match the transaction, I overrode the values to get the error to appear)

Screen Shot 2023-05-08 at 10 18 11 AM
Screen Shot 2023-05-08 at 10 18 26 AM


kev1n-peters:
Should we format the notional values to be displayed in USD? https://stackoverflow.com/a/16233919

anondev2323:
Oh, definitely, let's do that

anondev2323:
Discussed with @kev1n-peters that we don't want to use the IsTransferLimitedWarning here.  He's going to open another PR that would add an alert with a link to Discord support

PR #591, opened 2023-05-08
Removed SUI coingecko fallback price

kev1n-peters:


kev1n-peters:
No longer necessary since sui is now listed on CoinGecko

anondev2323 (approved):

PR #597, opened 2023-05-09
fix: handle network and token search better

### LINKED ISSUE #241: [QA] [Mainnet] Previously searched networks are only shown next time when Selecting Network
Preconditions: None

Environment: 
MacBook Pro, macOS Ventura 13.0.1, Chrome Version 108.0.5359.98
MacBook Pro, macOS Ventura 13.0.1, Firefox version 109.0.1 (64-bit)

Steps:
1. Click on Select Network
2. Type in the Search field letter "u" for example
3. Observe search results
4. Click on "X" to exit Select Network window
5. Click again on Select Network
6. Observe

Actual result: Second time opening the Select Network window only previously searched options are shown in the list, user has to refresh the page for the list to fully show again or make a new search and then delete the search.

Expected result: When closing the Select Network with search results, full list should be shown next time opening Select Network.

Attachment:

https://user-images.githubusercontent.com/91047061/226522256-84a7e4f6-dcfc-4f87-bf0d-b839d0196a2e.mov

anondev2323:
Closes #241


wormhole-connect/src/views/Bridge/Modals/TokensModal.tsx
```diff
@@ -245,28 +238,28 @@ function TokensModal() {
...
-    // filter only when a wallet is connected AND a network is selected
     if (!fromNetwork || !walletAddr) {
-      setTokens(networkTokens);
+      setTokens(TOKENS_ARR);
       return;
     }
-    const filtered = networkTokens.filter((t) => {
+
+    // get tokens that exist on the chain and have a balance greater than 0
+    const filtered = TOKENS_ARR.filter((t) => {
```
M-Picco, line 254: You could merge this two filters into a single one

wormhole-connect/src/views/Bridge/Modals/TokensModal.tsx
```diff
@@ -170,37 +165,35 @@ function TokensModal() {
...
-    },
-    [networkTokens],
-  );
+      setSearch(lowercase);
+    } else {
+      setSearch('');
+    }
+  };
+
+  const tokensSearch = () => {
```
M-Picco: You can maybe use `useMemo(() => { ... }, [ tokens, search ]`

M-Picco (approved):
The fix works. Just some comments on things that could be improved for readability.

M-Picco:
Something I noticed related to the network search: it seems we search for a matching `key` when we should check for the `displayName` instead. For example: when looking for `al`, it shows Moonbase, due to the key being `moonbasealpha`.

image

PR #598, opened 2023-05-10
WIP: Batch balance requests

anondev2323:
So far, just have batching for EVM


anondev2323:
Wondering if we should split some of these methods into a more general `utils` class/folder

anondev2323:
Gonna come back to this later

PR #614, opened 2023-05-17
refactor: get redeem from global-tx endpoint if available

anondev2323:
- Checks if the redeem tx is available using the `global-tx` endpoint
 - Waits until the number of block confirmations has passed before querying the api

### DIFF
--- a/wormhole-connect/src/utils/events.ts
+++ b/wormhole-connect/src/utils/events.ts
@@ -1,25 +1,48 @@
 import { BigNumber } from 'ethers';
 import {
-  ChainId,
   ChainName,
-  TokenId,
   EthContext,
   SuiContext,
   WormholeContext,
 } from '@wormhole-foundation/wormhole-connect-sdk';
-import { wh } from '../sdk';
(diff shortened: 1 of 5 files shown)


wormhole-connect/src/utils/events.ts
```diff
@@ -1,25 +1,48 @@
...
+import { ParsedMessage, ParsedRelayerMessage, PaymentOption, wh } from '../sdk';
 import { fromNormalizedDecimals } from '.';
 import { CHAINS } from '../config';
 import { arrayify } from 'ethers/lib/utils.js';
+import { fetchGlobalTx, getEmitterAndSequence } from './vaa';
+
+export const fetchRedeemTx = async (
+  txData: ParsedMessage | ParsedRelayerMessage,
+): Promise<{ transactionHash: string } | null> => {
+  try {
```
M-Picco, line 18: I think it might be best for readability to only wrap the `fetchGlobalTx` in the try/catch and avoid throwing the Error, which could be a `console.warn` instead.

M-Picco (approved):

PR #632, opened 2023-05-18
Aptos manual transfer support

### LINKED ISSUE #602: Implement aptos bridging without gas dropoff

kev1n-peters:


wormhole-connect/src/config/mainnet.ts
```diff
@@ -116,6 +127,7 @@ export const MAINNET_TOKENS: { [key: string]: TokenConfig } = {
     decimals: 18,
     solDecimals: 8,
     suiDecimals: 8,
+    aptosDecimals: 8,
```
kev1n-peters, line 130: Maybe it's time to refactor this. We could have a single `decimals` property that's the token's decimals on its native chain. Then we could have a function that returns the token's decimals for a given chain.

anondev2323: How do Aptos decimals work? solana and sui are 9 for native and max 8 for everything else right?

sdk/src/wormhole.ts
```diff
@@ -245,7 +250,7 @@ export class WormholeContext extends MultiProvider<Domain> {
     const context = this.getContext(sendingChain);
 
     // TODO: generalize to multiple chains
-    if (context.type === Context.SOLANA) {
+    if (context.type === Context.SOLANA || context.type === Context.APTOS) {
```
anondev2323: Can check if the relayer contract is defined. 
```
const relayer = context.getTokeBridgeRelayer(sendingChain);
if (!relayer) {
...
}
```

wormhole-connect/src/sdk/index.ts
```diff
@@ -184,6 +184,8 @@ export const sendTransfer = async (
       : '0';
     if (fromChainId === MAINNET_CHAINS.solana) {
       throw new Error('solana send with relay not supported');
+    } else if (fromChainId === MAINNET_CHAINS.aptos) {
+      throw new Error('aptos send with relay not supported');
```
anondev2323: Could check if the `sendWithRelay` method exists, rather than checking per chain

anondev2323: Also if the relayer contract is configured

wormhole-connect/src/utils/index.ts
```diff
@@ -107,6 +120,8 @@ export function getTokenDecimals(
       ? 9
       : chain === MAINNET_CHAINS.sui
       ? 9
+      : chain === MAINNET_CHAINS.aptos
+      ? 8
       : 18;
```
anondev2323, line 125: Should implement a `nativeDecimals` field on the Network config

wormhole-connect/src/views/Redeem/SendTo.tsx
```diff
@@ -217,6 +245,7 @@ function SendTo() {
     if (
       vaa &&
       txData.toChain !== 'solana' &&
+      txData.toChain !== 'aptos' &&
       txData.payloadID === PaymentOption.AUTOMATIC
```
anondev2323: We can just check if it's an automatic payment

wormhole-connect/src/config/mainnet.ts
kev1n-peters: Yeah, for Aptos the decimals are 8 for native and capped at 8 for everything else

anondev2323:
Just a note, it looks like the image for "Fletch" wallet isn't working.  Also, the Nightly logo for Aptos is different @M-Picco

wormhole-connect/src/config/mainnet.ts
kev1n-peters: We decided we'll do this in a separate PR to avoid any merge conflicts with some of the outstanding larger PRs

sdk/src/contexts/aptos/contracts.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  mustGetCore(chain: ChainName | ChainId): string {
+    const core = this.getCore(chain);
+    if (!core) throw new Error(`Core contract for domain ${chain} not found`);
+    return core;
+  }
+
+  getBridge(chain: ChainName | ChainId): string | undefined {
+    return this.getContracts(chain)?.token_bridge;
```
anondev2323, line 48: I would prefer to provide the contract interfaces from these methods, as opposed to just the address.  I know the sdk methods tend to just take an address, which is probably why you did it this way

anondev2323 (approved):

sdk/src/contexts/aptos/contracts.ts
kev1n-peters: For Aptos and Sui there isn't anything similar to typechain to generate interfaces unfortunately if that's what you're referring to

anondev2323: Oh, got it. It all looks good then!

Issue #675, opened 2023-06-07
[Enhancement] Add CI job to update OFAC block list


tonyjin:
## Describe the issue in detail

Part 2 of #673 

Add CI or other job that regularly pulls the OFAC list from the source of truth and updates hard coded list from #674 on a regular basis.

## What is the expected/desired behavior

- OFAC Treasury SDN site: https://ofac.treasury.gov/specially-designated-nationals-list-data-formats-data-schemas
- Pull `SDN.csv` file: https://www.treasury.gov/ofac/downloads/sdn.csv
- Filter for addresses (see #674)
- Update OFAC block list const file with any new addresses
- Build CI job that performs this on a regular basis - you could consider running this as part of a production build CI job, or alternatively on every CI run (which may cause flakey CI runs if the Treasury site is unavailable)

heyitaki:
Closed by #688

PR #688, opened 2023-06-13
Hardcode OFAC sanctioned wallets

### LINKED ISSUE #674: [Enhancement] Hard-code OFAC sanctioned wallet list for screening

Part 1 of #673

Fetch the OFAC sanctioned wallet list directly from the Treasury website, hard code those wallets (<300) into a config or const file, and screen connected wallets against that list. We see other projects adopt a similar approach (e.g. https://github.com/osmosis-labs/osmosis/blob/main/app/blocked.go).

- OFAC Treasury SDN site: https://ofac.treasury.gov/specially-designated-nationals-list-data-formats-data-schemas
- Sample Python code to parse addresses from SDN.csv file:

```
import requests
import re

r = requests.get('https://www.treasury.gov/ofac/downloads/sdn.csv')

addr = re.findall(r"Digital\sCurrency\sAddress\s\-\s[^\s]+\s([0-9a-zA-Z]+);", r.text)

for match in set(addr):
    print(match)
```
- Sample code above extracts all digital currency addresses, but you could potentially optimize to only extract valid EVM addresses given the other address types are for native assets we don't support (BTC, LTC, XMR, ZEC, DASH, BTG, BCH)
- Put addresses in a config or TS file maintaining constants
- Compare all connected wallet with blocked addresses and prevent user from proceeding if there is a match

heyitaki:
This PRs adds two scripts that 1) fetch the SDN list and 2) fetch and check the SDN list for updates, and throws if there are any updates. A contributor would resolve this by looking at the output of `checkSdnListForUpdates.ts` or running `getSdnList.ts` manually and updating `src/consts/wallets.ts`.

Resolves #674 and #675


wormhole-connect/scripts/ofac/utils.ts
```diff
@@ -0,0 +1,14 @@
+export const getSanctionedAddresses = async (): Promise<string[]> => {
```
anondev2323, line 1: Can we consolidate these files a little bit?  I think this could be moved into `getSdnList.ts` and exported from there

wormhole-connect/scripts/ofac/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+  const data = await readFileFromUrl(OFAC_SDN_LIST_URL);
+  const ethAddresses = data
+    .match(/Digital Currency Address - [a-zA-Z0-9]+ (0x)?([a-fA-F0-9]{40})/g)
+    ?.map((m) => ensure0x(m.split(' ').at(-1) ?? ''));
+  return [...new Set(ethAddresses)];
+};
+
+const OFAC_SDN_LIST_URL = 'https://www.treasury.gov/ofac/downloads/sdn.csv';
+
+const ensure0x = (str: string) => (str.startsWith('0x') ? str : `0x${str}`);
```
anondev2323: What about addresses that don't start with `0x` (i.e. Solana)?

wormhole-connect/src/consts/wallet.ts
```diff
@@ -0,0 +1,40 @@
...
+    '0x23773E65ed146A459791799d01336DB287f25334',
+    '0x22aaA7720ddd5388A3c0A3333430953C68f1849b',
+    '0x03893a7c7463AE47D46bc7f091665f1893656003',
+    '0x2717c5e28cf931547B621a5dddb772Ab6A35B701',
+    '0xD21be7248e0197Ee08E0c20D4a96DEBdaC3D20Af',
+    '0x39D908dac893CBCB53Cc86e0ECc369aA4DeF1A29',
+    '0x4f47bc496083c727c5fbe3ce9cdf2b0f6496270c',
+    '0x38735f03b30FbC022DdD06ABED01F0Ca823C6a94',
+    '0x97b1043abd9e6fc31681635166d430a458d14f9c',
+    '0xb6f5ec1a0a9cd1526536d3f0426c429529471f40',
```
anondev2323: This seems to be fewer than what I was thinking.  Maybe that's all though

wormhole-connect/src/utils/transferValidation.ts
```diff
@@ -99,42 +99,8 @@ export const validateAmount = (
...
-    }
-    return false;
-  }
-
-  const data = await res.json();
-  trmCache[address] = data[0];
-  return data[0].isSanctioned;
-}
+const checkAddressIsSanctioned = (address: string): boolean =>
+  SANCTIONED_WALLETS.has(address) || SANCTIONED_WALLETS.has('0x' + address);
```
anondev2323, line 103: I think the address would already be prefixed with `0x`, unless it's a Solana address or doesn't use that formatting

wormhole-connect/scripts/ofac/utils.ts
heyitaki: `getSdnList` itself is a script, so if we import something from there, there would be a side effect since the script would run.

wormhole-connect/src/consts/wallet.ts
heyitaki: This is it for EVM addresses; check out Tony's comment here https://github.com/wormhole-foundation/wormhole-connect/issues/674#issuecomment-1581099300

wormhole-connect/src/utils/transferValidation.ts
heyitaki: Imo better to not pass that assumption to the caller, and just check it here.

wormhole-connect/scripts/ofac/utils.ts
heyitaki: This is a good point, do you think we should create a single list of all sanctioned addresses that we compare against? Or would it be better to create an object/Map of form `chainId => Array<sanctionedAddresses>`

anondev2323: I think a single list is fine, it's not that long of a list.  We don't just support EVM, so we should account for other types of addresses also

heyitaki: Done

wormhole-connect/scripts/ofac/getSdnList.ts
```diff
@@ -0,0 +1,10 @@
+import { getSanctionedAddresses } from './utils';
+
+const getSdnList = async () => {
+  console.log('Parsing OFAC SDN list...');
+  const addresses = await getSanctionedAddresses();
+  console.log(JSON.stringify(addresses, null, 2));
```
anondev2323, line 6: Can we write to the file here? Instead of just logging the addresses?

wormhole-connect/src/consts/wallet.ts
```diff
@@ -0,0 +1,233 @@
+export const SANCTIONED_WALLETS = Object.freeze(
```
anondev2323, line 1: Should add a quick note in CONTRIBUTING.md about how to update this file

anondev2323 (approved):

wormhole-connect/src/consts/wallet.ts
heyitaki: Done!

wormhole-connect/scripts/ofac/getSdnList.ts
heyitaki: Agreed offline that this may entail more work that it's worth

PR #703, opened 2023-06-15
Docs: Adding typedoc for SDK

barnjamin:
Adding a script to run typedoc against the SDK which can be served easily with github pages.

This PR also removes the current `/docs` directory content in favor of typedoc generated content.

Draft until there is some consensus that this is a good thing to do. 

Example of what this might look like:
https://barnjamin.github.io/wormhole-connect/


anondev2323:
I like it.  A couple thoughts:
- The Readme is the root readme and not the sdk Readme
- It would nice to group this by root code (i.e. WormholeContext, VAA utils, etc) and then by chain.  You can specify a Category like so https://typedoc.org/tags/category/

anondev2323 (approved):

PR #710, opened 2023-06-21
Routes refactor

### LINKED ISSUE #676: Standardized routing options
Come up with a standard way to introduce more routes. It should be compatible with Hasflow and CCTP initially, but should be easily expandable.

anondev2323:
Separates the bridge input from the route logic. Implements a standard set of methods for each route.

TODO:
- [x] cannot input amount 0.01
- [x] destination token should not filter out tokens without a balance (I just noticed there's a new "Show assets with 0 balance" checkbox.  Maybe consider having a "My Tokens" and "All Tokens" tab?)
- [x] if there is only 1 option for receiving token, auto fill that value
- [ ] if there is only 1 network supported by a wallet, auto fill that network
- [x] separate route logic from preview component? Pass in the `rows` object to be rendered
- [ ] switch network button is not positioned like it is in the designs (should be inline with the "To" text and Connect Wallet button)
- [x] Send transaction block explorer link not working (Look up this tx from Goerli and then click on "Etherscan") https://goerli.etherscan.io/tx/0xf863a7ddefe83f86862f3cf42d005776cb59d066510582cef2789cc887210d9e


sdk/src/config/MAINNET.ts
```diff
@@ -174,12 +174,13 @@ const MAINNET_CONFIG: WormholeConfig = {
...
     sui: 'https://rpc.mainnet.sui.io',
     aptos: 'https://fullnode.mainnet.aptoslabs.com/v1',
-    sei: '', // TODO: fill in
+    sei: '', // TODO: fill in,
   },
   rest: {
     sei: '',
   },
   chains: MAINNET,
+  wormholeHosts: ['https://wormhole-v2-mainnet-api.certus.one'],
```
kev1n-peters: I think we should add more guardians in case the jump guardian is down: https://github.com/wormhole-foundation/wormhole/blob/5ce1a42c0b22cd6fd959aebea8fa8d027209b0bf/sdk/js/src/relayer/consts.ts#L197-L203

anondev2323: Sorry I missed this.  Just updated

M-Picco (approved):

PR #751, opened 2023-07-07
documentation

anondev2323:
TODO:
- [ ] <s>get `@category` tag working</s>

Note: Category doesn't seem to work as well as I imagined.  I would like to have a sidebar ordered like so, will probably need to revisit and handle in another PR
- SDK
    - Base
    - EVM
    - Solana
         - Utils
    - Aptos
    - Sui
    - Sei
- Utils
    - VAA
    - Address
    - etc...


sdk/src/contexts/abstracts/tokenBridge.ts
```diff
@@ -7,13 +7,36 @@ import {
...
+   * @see {@link ContractsAbstract}
+   */
   protected abstract contracts: AnyContracts;
+  protected abstract context: WormholeContext;
 
   /**
-   * These operations have to be implemented in subclasses.
+   * Send a Token Bridge transfer
+   *
+   * @dev This _must_ be claimed on the destination chain, see `redeem`
```
M-Picco: Is there a `@link` attribute to reference the `redeem` symbol?

sdk/src/contexts/abstracts/tokenBridge.ts
```diff
@@ -36,51 +73,118 @@ export abstract class TokenBridgeAbstract<TransactionResult> {
...
+  /**
+   * Parse a token address from a 32-byte Wormhole address which is the standard format for a given chain
+   *
+   * @param address The 32-byte wormhole address
+   * @returns The token address in the blockchain specific format
+   */
+  protected abstract parseAssetAddress(address: any): Promise<string>;
 
+  /**
+   * Fetches the address for a token on any chain (These are the Wormhole token addresses, not necessarily the cannonical version of that token)
```
M-Picco: Maybe add that the address is for a token "representation"

sdk/src/contexts/eth/index.ts
```diff
@@ -74,6 +77,13 @@ export class EthContext<
     return addr;
   }
 
+  /**
+   * Returns the decimals for a token
```
M-Picco: `fetchTokenDecimals` seems to be present on all contexts. We should move the definition up in the hierarchy to `TokenBridgeAbstract`.

sdk/src/contexts/abstracts/tokenBridge.ts
```diff
@@ -36,51 +73,118 @@ export abstract class TokenBridgeAbstract<TransactionResult> {
   ): Promise<TransactionResult>;
 
   /**
-   * Format an address to a 32-byte array, padded with zeros if needed
-   * @param address The address to format
-   * @returns A 32 byte array of the address, padded with zeros if needed
+   * Format an address to 32-bytes which can be utilized by the Wormhole contracts
```
M-Picco: I don't know the current state of the wh sdk rewrite discussion, but if I remember correctly there was a proposal to adopt the term "canonical address" for a blockchain specific address (e.g: 40 byte hex encoded for EVMs, 32 byte base58 encoded for Solana, etc) and a "universal address" which are 32 byte address which wormhole uses (which may be the same as a canonical address or not).

Perhaps we could use this, or whatever other terms we see fit, in order to create a clear division of the two concepts.

We should also state that `formatAddress`/`parseAddress` are specifically to deal with normal account addresses, and not with asset addresses (the async methods are for that).

sdk/src/contexts/abstracts/tokenBridge.ts
```diff
@@ -36,51 +73,118 @@ export abstract class TokenBridgeAbstract<TransactionResult> {
...
 
-  protected abstract formatAssetAddress(address: string): Promise<Uint8Array>;
-  protected abstract parseAssetAddress(address: string): Promise<string>;
+  /**
+   * Format a token address to 32-bytes which can be utilized by the Wormhole contracts
+   *
+   * @param address The token address as a string
+   * @returns The token address as a 32-byte Wormhole address
+   */
+  protected abstract formatAssetAddress(address: string): Promise<any>;
```
M-Picco, line 99: Perhaps we should clear up why these two methods exists, which is basically that converting some assets to a universal representation might require querying a registry first. I recall getting confused at first when they were introduced for the Sui integration, so clearing it up for other developers might be worth it.

sdk/src/contexts/abstracts/tokenBridge.ts
```diff
@@ -36,51 +73,118 @@ export abstract class TokenBridgeAbstract<TransactionResult> {
...
     chain: ChainName | ChainId,
   ): Promise<BigNumber | null>;
 
+  /**
+   * Redeems funds for a token bridge transfer on the destination chain
+   *
+   * @param destChain The destination chain name or id
+   * @param signedVAA The Signed VAA bytes
+   * @param overrides Optional overrides, varies between chains
+   * @param payerAddr Optional, Solana specific. The address that pays for the redeem transaction
```
M-Picco: The `payerAddr` might not be specific to Solana. We can just specify that it's the address that will send the transaction/pay the fees

M-Picco (approved):
Some small suggestions, but it looks good.

I love the work @barnjamin and you put into the docs for the project 👏

sdk/src/contexts/eth/index.ts
anondev2323: I couldn't get the Abstract contracts to generate the ts doc or anything else to inherit the documentation from the abstract.  Maybe need to look into it a bit more

PR #757, opened 2023-07-11
Routes: Tokens modal

anondev2323:
- Don't filter out tokens without a balance for the destination token selection
- Add "Available tokens/All tokens" tabs

### DIFF
--- a/wormhole-connect/src/components/TokensModal.tsx
+++ b/wormhole-connect/src/components/TokensModal.tsx
@@ -5,7 +5,7 @@ import { useMediaQuery } from '@mui/material';
 import { makeStyles } from 'tss-react/mui';
 import { ChainName } from '@wormhole-foundation/wormhole-connect-sdk';
 import { RootState } from '../store';
-import { CHAINS } from '../config';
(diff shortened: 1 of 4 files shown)


wormhole-connect/src/components/TokensModal.tsx
```diff
@@ -272,16 +334,24 @@ function TokensModal(props: Props) {
...
-    setShowNullBalances(!showNullBalances);
-  };
+  const tabs = [
+    {
+      label: 'Available Tokens',
+      panel: <DisplayTokens tokens={displayedTokens} balances={tokenBalances} walletAddress={walletAddress} network={network} selectToken={selectToken} loading={loading} search={search} />
+    },
+    {
+      label: 'All Tokens',
+      panel: <DisplayTokens tokens={TOKENS_ARR} balances={tokenBalances} walletAddress={walletAddress} network={network} selectToken={selectToken} loading={loading} search={search} />
```
M-Picco: Should it use the same `selectToken` handler? It lets me pick other chains' native assets. For example, I set Fuji as network but I can still pick ETH or SOL.

image

PR #758, opened 2023-07-12
Routes: move Preview data building to the routes

M-Picco:


wormhole-connect/src/views/Bridge/Preview.tsx
```diff
@@ -182,8 +81,34 @@ function Preview(props: { collapsed: boolean }) {
     receiveNativeAmt,
     gasEst,
     dispatch,
+    relayerFee,
   ]);
 
+  useEffect(() => {
+    const computeRelayerFee = async () => {
```
anondev2323, line 88: Wondering if this should be in the main bridge component...

anondev2323 (approved):

wormhole-connect/src/views/Bridge/Preview.tsx
M-Picco: I thought so as well, now that it is decoupled from the row building

PR #765, opened 2023-07-14
add missing apostrophe in docs

arihantbansal:


aadam-10:
hi @arihantbansal, it seems like many of the files were accidentally changed as part of this pr, would you mind resubmitting with only the appropriate changes?

PR #769, opened 2023-07-17
adding method on wormhole to get a VAA

barnjamin:


sdk/src/wormhole.ts
```diff
@@ -365,6 +368,48 @@ export class WormholeContext extends MultiProvider<Domain> {
...
+   * @param msg The MessageIdentifier used to fetch the VAA
+   * @returns The ParsedVAA if available
+   */
+  async getVAA(msg: MessageIdentifier): Promise<ParsedVaa | undefined> {
+    const { emitterChain, emitterAddress, sequence } = msg;
+    const url = `${this.conf.api}/api/v1/vaas/${emitterChain}/${emitterAddress}/${sequence}`;
+    const response = await axios.get(url);
+
+    // TODO: raise exception? wait?
+    if (!response.data.data) return;
```
M-Picco, line 406: You could try using `getSignedVaaWithRetry` ([source code](https://github.com/wormhole-foundation/wormhole/blob/8e0cf4c31f39b5ba06b0f6cdb6e690d3adf3d6a3/sdk/js/src/rpc/getSignedVAAWithRetry.ts)) to query the guardian rpc as a fallback

barnjamin: should I? What is the expected behavior for this fn? that it waits until its available or returns immediately if not found?

anondev2323:
I moved these changes over to the sdk repo, so this can be closed now.  Soon, I will update Connect to use the published sdk packages

PR #792, opened 2023-07-27
wip: Cosmos Gateway integration

M-Picco:


derpy-duck (approved):
Looks good to me

sdk/src/config/DEVNET.ts
```diff
@@ -0,0 +1,95 @@
...
+export type DevnetChainId = (typeof DEVNET_CHAINS)[DevnetChainName];
+/**
+ * chain name to contracts mapping
+ */
+export type ChainContracts = {
+  [chain in DevnetChainName]: Contracts;
+};
+
+const DEVNET: { [chain in DevnetChainName]: ChainConfig } = {
+  ethereum: {
```
anondev2323, line 30: Why is Ethereum in Devnet?  Also, Solana is a devnet chain, does it belong in here?  What's the purpose of creating a devnet file instead of using testnet?

anondev2323 (commented):
Looking good!  Really only had the one question about devnet

sdk/src/config/DEVNET.ts
kev1n-peters: I think this is for the Tilt environment from the wormhole monorepo

M-Picco: As @kev1n-peters mentioned, there was a point I needed to set up the tilt environment locally, which required defining new chains and RPCs. I can remove it if you think it muddles up the plugin.

Re: Solana, their testnet is called "devnet"

PR #795, opened 2023-08-02
Added Base support

kev1n-peters:


wormhole-connect-loader/src/types.ts
```diff
@@ -1,4 +1,4 @@
-import { Theme } from './theme';
+import {Theme} from "./theme";
```
anondev2323, line 1: Hm, why these lint changes? Not consistent with the rest of the repo

anondev2323 (approved):
Looks good to me!

PR #806, opened 2023-08-04
Routes more abstractions

derpy-duck:
The goal with this PR is to remove things from the React code that assume the token bridge route or token bridge with relay route.

For example, 'isTransferCompleted' was added to the route interface, and the concept of storing a VAA in the react state was replaced by storing a boolean 'isRedeemReady'

I tested this with the ‘Bridge’ route and the 'Relay' route (as well as testing pasting a transaction while its in progress and seeing it live update to complete)

### DIFF
--- a/wormhole-connect/src/utils/routes/bridge.ts
+++ b/wormhole-connect/src/utils/routes/bridge.ts
@@ -261,6 +261,13 @@ export class BridgeRoute extends RouteAbstract {
     return adaptParsedMessage(message);
   }
 
+  public async isTransferCompleted(
+    destChain: ChainName | ChainId,
+    signedVaa: string,
+  ): Promise<boolean> {
+    return await wh.isTransferCompleted(destChain, signedVaa);
+  }
+
   public async getPreview({
     destToken,
     sourceGasToken,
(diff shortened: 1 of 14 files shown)


wormhole-connect/src/utils/routes/relay.ts
```diff
@@ -229,6 +233,15 @@ export class RelayRoute extends BridgeRoute {
     };
   }
 
+  public async isTransferCompleted(
```
M-Picco, line 236: RelayRoute extends from BridgeRoute, so this method definition can be omitted

wormhole-connect/src/utils/routes/bridge.ts
```diff
@@ -226,13 +226,21 @@ export class BridgeRoute extends RouteAbstract {
...
 
-  async redeem(
-    destChain: ChainName | ChainId,
-    vaa: Uint8Array,
-    payer: string,
-  ): Promise<string> {
+  public async readyForRedeem(
+    txData: ParsedMessage | ParsedRelayerMessage,
+  ): Promise<boolean> {
+    const parsedVaa = await fetchVaa(txData);
```
M-Picco, line 232: I noticed that the change introduced 4 calls to re-fetch the vaa.

Perhaps the method can be changed to receive an object from a union type (e.g: `type MessageInfo = VaaInfo | CCTPAttestationInfo`) and leave it to the route to pick which one to use. What do you think?

M-Picco (commented):
I think it looks ok. However, I do think we should discuss the point on the re-fetching of the vaa/tx information in general.

derpy-duck:
Addressed with a different PR

PR #839, opened 2023-08-22
Add cctp rebased

derpy-duck:


wormhole-connect/src/components/Options.tsx
```diff
@@ -47,15 +47,11 @@ type Option = {
...
   children: Option[];
   active?: number;
+  onSelect: (key: any) => void;
 };
 
 function Options(props: Props) {
   const { classes } = useStyles();
-  // dispatch selectOption event
-  const emitSelectOption = (key: any) => {
-    const event = new CustomEvent('selectOption', { detail: key });
```
anondev2323, line 56: There was something listening to this event in `GasOptions`, did you check that it isn't affected by this deletion?

wormhole-connect/src/utils/gasEstimates.ts
```diff
@@ -253,6 +253,16 @@ const getGasFeeFallback = async (
       );
     const gasFees = BigNumber.from(gasEst).mul(gasPrice);
     return toFixedDecimals(utils.formatEther(gasFees), 6);
+  } else if (route === Route.CCTPRelay) {
+    const gasEst = gasEstimates.sendCCTPWithRelay;
```
anondev2323: Maybe we can clean up the gas estimate logic a bit by creating a mapping of route to gas estimates.  Something like
```
const gasEstimates = {
  [Route.BRIDGE]: {
    send: ...,
    sendNative: ...,
    claim: ...,
  },
  [Route.RELAY]: {
    send: ...,
    sendNative: ...,
  },
  etc...
}
```

wormhole-connect/src/utils/routes/cctpManual.ts
```diff
@@ -0,0 +1,612 @@
...
+      return isMainnet ? 'avalanche' : 'fuji';
+  }
+  throw Error('Invalid CCTP domain');
+}
+function getDomainCCTP(chain: ChainName | ChainId): number {
+  const chainId = wh.toChainId(chain);
+  switch (chainId) {
+    case 2:
+      return 0;
+    case 6:
```
anondev2323: Not a big deal now since there's only a few chains supported, but maintaining these switch cases with growing chains can be a bit of a headache.  Consider adding a CCTP domain on the chain config?

wormhole-connect/src/utils/routes/cctpManual.ts
```diff
@@ -0,0 +1,612 @@
...
+
+    // approve
+    await chainContext.approve(
+      sendingChain,
+      circleTokenMessenger.address,
+      tokenAddr,
+      parsedAmt,
+    );
+
+    console.log('About to send 2');
```
anondev2323: Clean up console logs

wormhole-connect/src/utils/routes/cctpManual.ts
```diff
@@ -0,0 +1,612 @@
...
+        to: messageInfo.mintRecipient,
+        relayerFee: messageInfo.relayerFee,
+        toNativeTokenAmount: messageInfo.toNativeTokenAmount,
+      };
+    }
+    return {
+      sendTx: messageInfo.transactionHash,
+      sender: messageInfo.depositor,
+      amount: messageInfo.amount.toString(),
+      payloadID: PayloadType.MANUAL,
```
anondev2323, line 353: I'm not sure this is correct.  payloadID is the first byte which identifies the type of transfer. Maybe we can add another field for Automatic vs Manual

wormhole-connect/src/views/Bridge/GasOptions.tsx
```diff
@@ -165,26 +165,44 @@ function GasOptions(props: { disabled: boolean }) {
...
+  const { token, fromNetwork, toNetwork, automaticRelayAvail, gasEst, route } =
     useSelector((state: RootState) => state.transferInput);
+  const automaticOrManual =
+    route === Route.BRIDGE
+      ? Route.BRIDGE
+      : route === Route.RELAY
+      ? Route.RELAY
+      : route === Route.CCTPManual
+      ? Route.BRIDGE
+      : Route.RELAY;
```
anondev2323, line 178: Can we change this variable to `isAutomatic` as a boolean?  Could probably add this value on the route as well for easy access

wormhole-connect/src/views/Bridge/Send.tsx
```diff
@@ -190,6 +196,8 @@ function Send(props: { valid: boolean }) {
     const valid = isTransferValid(validations);
     if (!valid) return;
 
+    // TODO: should use a different enum than Route
+    // which is either one transaction or two
```
anondev2323, line 200: Agree

wormhole-connect/src/views/Redeem/SendTo.tsx
```diff
@@ -126,12 +126,11 @@ function SendTo() {
       ? inProgress && !transferComplete
       : !transferComplete;
   const manualClaimText =
-    transferComplete || txData.payloadID === Route.RELAY
+    transferComplete || txData.payloadID === Route.RELAY // todo: should be the other enum, should be named better than payload id
```
anondev2323, line 128: Per earlier comment.  Yes, payload ID is different.  Can make a `txData.isAutomatic` value

anondev2323 (commented):
Great progress!

wormhole-connect/src/components/Options.tsx
derpy-duck: Oh - rather than use a listener, I just passed down a function as props from GasOptions into Options - so there is no more listener there

wormhole-connect/src/utils/routes/cctpRelay.ts
```diff
@@ -0,0 +1,696 @@
...
+      (log) =>
+        log.topics[0] ===
+        '0x8c5261668696ce22758910d05bab8f186d6eb247ceac2af2e82c7dc17669b036',
+    )[0];
+
+    const message = new utils.Interface([
+      'event MessageSent(bytes message)',
+    ]).parseLog(messageLog).args.message;
+
+    const messageHash = utils.keccak256(message);
```
barnjamin, line 475: does this work? 

Locally I had to convert the message to bytes _then_ hash it _then_ apply `0x` prefix to get the actual Attestation from circle



```ts
    const [messageLog] = messageLogs;
    // Need to get the message (0xdeadbeef...) to bytes prior to hashing
    const msgBytes = new Uint8Array(
      Buffer.from((messageLog.args.message as string).slice(2), 'hex'),
    );
    const messageHash = `0x${Buffer.from(keccak256(msgBytes)).toString('hex')}`;
```

PR #863, opened 2023-08-25
Initial Builder

evan-gray:


builder/src/consts.ts
```diff
@@ -0,0 +1,193 @@
...
+    mainnet: "base",
+  },
+  {
+    name: "Osmosis",
+    testnet: "osmosis",
+    mainnet: "osmosis",
+  },
+];
+
+// TODO: move the connect config to a shared sdk
```
anondev2323, line 76: Agree, we should make it so we can generate these lists

anondev2323 (approved):

PR #864, opened 2023-08-28
Cctp bugfixes

derpy-duck:
TLDR: Should have thoroughly tested before merging in the CCTP changes into development :'(

Changes:
- add 'nativeTokenAmount' and 'maxSwapAmount' to the routes abstract class, because these both are different for CCTP than for the token bridge routes.
- edit the 'amount' check in transferValidation.ts to not immediately say 'amount' is allowed if it is greater than balance - it should still check that it is >= 'minAmt'
- in Bridge.tsx, when computing the 'supportedSourceTokens' and 'supportedDestTokens', search over *all routes* not just the current route! 
- use the wormscan URL instead of wormhole explorer in <Tag>
- Have the supportedSourceTokens/supportedDestTokens functions take source and dest chains as (optional) inputs (needed for USDC because we need to have the supported tokens be *native* USDC
- Only show native USDC in the tokens modal
- Don't wait for the signed attestation before proceeding to the redeem page (accomplished by adding a 'unsigned' optional parameter to getMessageInfo)

### DIFF
--- a/wormhole-connect/src/config/index.ts
+++ b/wormhole-connect/src/config/index.ts
@@ -30,7 +30,6 @@ export const isMainnet = config && config.env === 'mainnet';
 export const CONFIG = isMainnet ? CONF.MAINNET : CONF.TESTNET;
 export const ENV = isMainnet ? 'MAINNET' : 'TESTNET';
 
-export const WORMHOLE_EXPLORER = 'https://wormhole.com/explorer/';
 export const WORMSCAN = 'https://wormholescan.io/#/';
 export const WORMHOLE_API = isMainnet
   ? 'https://api.wormholescan.io/'
--- a/wormhole-connect/src/config/index.ts
+++ b/wormhole-connect/src/config/index.ts
@@ -31,6 +31,7 @@ export const CONFIG = isMainnet ? CONF.MAINNET : CONF.TESTNET;
 export const ENV = isMainnet ? 'MAINNET' : 'TESTNET';
 
 export const WORMHOLE_EXPLORER = 'https://wormhole.com/explorer/';
+export const WORMSCAN = 'https://wormholescan.io/#/';
 export const WORMHOLE_API = isMainnet
   ? 'https://api.wormholescan.io/'
   : 'https://api.testnet.wormholescan.io/';
--- a/wormhole-connect/src/utils/routes/bridge.ts
+++ b/wormhole-connect/src/utils/routes/bridge.ts
@@ -35,6 +35,8 @@ interface TransferDestInfoParams {
 }
 
 export class BridgeRoute extends BaseRoute {
+  NATIVE_GAS_DROPOFF_SUPPORTED = false;
+
   async isRouteAvailable(
     sourceToken: string,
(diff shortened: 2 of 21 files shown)


wormhole-connect/src/config/index.ts
```diff
@@ -31,6 +31,7 @@ export const CONFIG = isMainnet ? CONF.MAINNET : CONF.TESTNET;
 export const ENV = isMainnet ? 'MAINNET' : 'TESTNET';
 
 export const WORMHOLE_EXPLORER = 'https://wormhole.com/explorer/';
+export const WORMSCAN = 'https://wormholescan.io/#/';
```
anondev2323, line 33: Is this the same for mainnet/testnet?

anondev2323: Could probably remove the WORMHOLE_EXPLORER const now

wormhole-connect/src/utils/routes/bridge.ts
```diff
@@ -322,4 +322,21 @@ export class BridgeRoute extends BaseRoute {
...
     );
   }
+
+  async nativeTokenAmount(
+    destChain: ChainName | ChainId,
+    token: TokenId,
+    amount: BigNumber,
+    walletAddress: string,
+  ): Promise<BigNumber> {
+    throw new Error('Not implemented');
```
anondev2323, line 334: 'Not supported'

anondev2323: Can we add a const at the top of the class that says whether native gas dropoff is a supported feature of that route?

wormhole-connect/src/config/index.ts
derpy-duck: it is the same, but testnet has a '?network=TESTNET' flag that I put in the code (branches on 'isMainnet')

wormhole-connect/src/utils/routes/bridge.ts
derpy-duck: yup done!

anondev2323:
Couple issues when spinning up this branch
1. No token balances found on the source chain
2. The destination token modal shouldn't filter for tokens with balance, should show any tokens available from any of the routes
3. Gas Payment section was empty when selecting Goerli -> Mumbai (I think because there are no routes available for this token pair)

https://github.com/wormhole-foundation/wormhole-connect/assets/122566365/4ca00978-7d72-4d0d-bf85-e127ffb9563d

derpy-duck:
> Couple issues when spinning up this branch
> 
> 1. No token balances found on the source chain
> 2. The destination token modal shouldn't filter for tokens with balance, should show any tokens available from any of the routes
> 3. Gas Payment section was empty when selecting Goerli -> Mumbai (I think because there are no routes available for this token pair)
> 
>  Screen.Recording.2023-08-28.at.10.42.09.AM.mov

Addressed all of these except for (3) - trying to figure this out!

M-Picco (approved):

anondev2323 (approved):

PR #870, opened 2023-08-28
Routes component rebased

anondev2323:
- Icon logic was driving me crazy, so I cleaned that up a bit
- Added a combined "routes" component (instead of Route + Gas Options components)
- Added AUTOMATIC_DEPOSIT constant on Route classes

TODO:
- [x] Check route availability and hide unavailable routes
- [x] Best route logic
- [x] Calculate fees/USD
- [x] Fix native gas slider

https://github.com/wormhole-foundation/wormhole-connect/assets/122566365/a250a879-ce55-4f80-8be8-6869566bce41


wormhole-connect/src/components/Options.tsx
```diff
@@ -1,17 +1,18 @@
...
-const useStyles = makeStyles()((theme) => ({
+const useStyles = makeStyles()((theme: any) => ({
   options: {
     backgroundColor: theme.palette.card.background,
     borderBottomLeftRadius: '8px',
     borderBottomRightRadius: '8px',
   },
   option: {
-    position: 'relative',
+    // position: 'relative',
```
M-Picco: Remove commented props

derpy-duck (approved):
This is huge! Thanks for the catch on 'relayerFee' in the CCTPRelay route. Love how this looks.. makes a lot more sense than the dual 'route'/'gasoption' setup. 

Lmk if you want me to test it out at all - but the PR looks great

wormhole-connect/src/components/Options.tsx
anondev2323: I actually just cleaned this up 😅

M-Picco (approved):

PR #879, opened 2023-08-29
Adding Optimism and Arbitrum support

derpy-duck:
One note that I want some feedback on @anondev2323  @M-Picco :

One minor change I made was that I made the ‘relay.ts’ route’s ‘isSupportedDestToken’ and ‘isSupportedSourceToken’ functions not be promises (return immediately) - meaning they don’t check against the relayer contract to see if they are supported - they return true as long as the ‘manual’ route supports the token

This improved efficiency of the 'find all supported source tokens' function (and reduces rpc calls) - because it makes it so that now every route's 'isSupportedDestToken' and 'isSupportedSourceToken's are not promises but immediate functions -  but maybe is undesirable - if you think we should revert this specific change I can do that!


anondev2323:
Hey, regarding your question about `isSupportedDestToken` and `isSupportedSourceToken`.  I think that we actually should revert that change.  That method should check if a token is supported by that route specifically, because it can be used other places such as input validation.  Also, it needs to be async for Hashflow as well.

I actually noticed that this was doing a lot of unnecessary calls, so I changed the logic to check if that token is already supported via another route before querying. We could probably further optimize by sorting the bridge route to the beginning and then querying.  You can see the change [here](https://github.com/wormhole-foundation/wormhole-connect/blob/4b9cdd859fb7cb04afd19e2f704795f200d4bf31/wormhole-connect/src/utils/routes/operator.ts#L156-L206) included in [this PR ](https://github.com/wormhole-foundation/wormhole-connect/pull/908)

anondev2323 (approved):

PR #880, opened 2023-08-29
Cosmos gateway integration

M-Picco:
Squashed version of #792


derpy-duck (approved):

kev1n-peters:
Should we update the docs here to add the Osmosis network (along w/ the OSMO token in the Tokens section)? https://github.com/wormhole-foundation/wormhole-connect/blob/6be9d50b059a1a4785808bb37b11e134d32ddae4/wormhole-connect-loader/README.md#L30

kev1n-peters:
Should we add Osmosis here? (although there isn't a chainID, so maybe this chain integration is fundamentally different than the others) https://github.com/wormhole-foundation/wormhole-connect/blob/6be9d50b059a1a4785808bb37b11e134d32ddae4/wormhole-connect-loader/src/types.ts#L2

PR #884, opened 2023-08-30
Add foreign asset address cache

M-Picco:
Create a cache object which `WormholeContext` passes down to the specific contexts. The aim of this cache is to prevent repeating `getForeignAsset` queries when the information will be the same.


derpy-duck (approved):

kev1n-peters:
lgtm, nice use of caching

kev1n-peters (approved):

PR #902, opened 2023-08-31
Fixed guardian RPCs from being called too frequently

kev1n-peters:


wormhole-connect/src/views/Redeem/Redeem.tsx
```diff
@@ -17,131 +17,157 @@ import Spacer from '../../components/Spacer';
...
+      <GovernorEnqueuedWarning
+        show={!messageInfo && isVaaEnqueued}
+        chain={txData.fromChain}
+      />
+      <Stepper />
+    </div>
+  );
 }
 
 function mapStateToProps(state: RootState) {
```
M-Picco, line 166: Is this stuff at the bottom of the file still needed? The redux state and dispatch can now be used directly inside the Redeem component

M-Picco (approved):

wormhole-connect/src/views/Redeem/Redeem.tsx
evan-gray: We were going for the minimally invasive fix for this, but agree we should switch to the hooks (and functional components) across the board.

PR #907, opened 2023-09-01
Solana testnet automatic redeem support

kev1n-peters:


kev1n-peters:
The contract interfaces may change after an internal audit, so putting this back into draft mode.

wormhole-connect/src/utils/routes/bridge.ts
```diff
@@ -29,9 +29,18 @@ import {
   TransferDestInfoBaseParams,
   TransferInfoBaseParams,
 } from './types';
-import { hexlify } from 'ethers/lib/utils.js';
 import { isCosmWasmChain } from '../cosmos';
 import { fetchVaa } from '../vaa';
+import { hexlify } from 'ethers/lib/utils.js';
+
+export interface BridgePreviewParams {
```
M-Picco: Should this interface be here? I don't see it used anywhere else

wormhole-connect/public/index.html
```diff
@@ -15,7 +15,7 @@
   </head>
   <body style="margin: 0">
     <noscript>You need to enable JavaScript to run this app.</noscript>
-    <!-- <div id="wormhole-connect" config='{"networks": ["goerli", "mumbai"], "tokens": ["ETH", "WETH", "MATIC", "WMATIC"], "theme": "light"}'></div> -->
-    <div id="wormhole-connect"></div>
+    <div id="wormhole-connect" config='{"env": "testnet", "networks": ["fuji", "solana", "bsc", "goerli"], "tokens": ["AVAX", "WAVAX", "SOL", "WSOL", "BNB", "WBNB", "ETH", "WETH"], "theme": "light"}'></div>
+    <!-- <div id="wormhole-connect"></div> -->
```
M-Picco: Should this be commented out?

sdk/src/contexts/solana/relayer.ts
```diff
@@ -0,0 +1,186 @@
...
+      this.getRedeemerConfig(),
+    ]);
+    const swapRatePrecisionBN = new BN(swapRatePrecision);
+    const nativeSwapRate = this.calculateNativeSwapRate(
+      swapRatePrecisionBN,
+      solSwapRate,
+      swapRate,
+    );
+    const maxSwapAmountIn =
+      decimals > SOL_DECIMALS
```
M-Picco, line 80: Does it need to be decouple in two cases? If `decimals < SOL_DECIMALS` the exponent `decimals - SOL_DECIMALS` will be negative, which is the same as dividing by `TEN.pow(SOL_DECIMALS - decimals)`. Perhaps there's something from the BN lib I'm missing

sdk/src/contexts/solana/relayer.ts
```diff
@@ -0,0 +1,186 @@
...
+        this.getRedeemerConfig(),
+      ]);
+    const swapRatePrecisionBN = new BN(swapRatePrecision);
+    const nativeSwapRate = this.calculateNativeSwapRate(
+      swapRatePrecisionBN,
+      solSwapRate,
+      swapRate,
+    );
+    const swapAmountOut =
+      decimals > SOL_DECIMALS
```
M-Picco, line 108: Same here

sdk/src/contexts/solana/relayer.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    return BigInt(swapAmountOut.toString());
+  }
+
+  async fetchSwapEvent(signature: string): Promise<SwapEvent | null> {
+    const transaction = await this.connection.getParsedTransaction(signature);
+    if (transaction) {
+      const logMessages = transaction.meta?.logMessages || [];
+      for (const msg of logMessages) {
+        const matches =
```
M-Picco, line 124: Is there any other way to retrieve this data or is using a regex the only "elegant" way?

M-Picco (commented):
Didn't notice anything out of place. Just some minor comments.

wormhole-connect/src/utils/routes/bridge.ts
kev1n-peters: This shouldn't be here, removed

wormhole-connect/public/index.html
kev1n-peters: This shouldn't be changed, reverted

sdk/src/contexts/solana/relayer.ts
kev1n-peters: Unfortunately this is how we have to do it on Solana. The contract calls `msg!` which is just a macro that logs data in the transaction.

sdk/src/contexts/solana/relayer.ts
kev1n-peters: Good catch, we can simplify this

kev1n-peters: Actually, I changed this and it broke the gas slider. Here's a little example that shows that the two expressions aren't equivalent:

```
const TEN = new BN(10);
const SWAP_RATE_PRECISION = new BN(100_000_000);
const SOL_DECIMALS = 9;

function calculateSwapAmount(
  decimals: number,
  maxNativeSwapAmount: BN,
  nativeSwapRate: BN,
  firstCaseOnly: boolean = false
) {
  return decimals > SOL_DECIMALS || firstCaseOnly
    ? maxNativeSwapAmount
        .mul(nativeSwapRate)
        .mul(TEN.pow(new BN(decimals - SOL_DECIMALS)))
        .div(SWAP_RATE_PRECISION)
    : maxNativeSwapAmount
        .mul(nativeSwapRate)
        .div(TEN.pow(new BN(SOL_DECIMALS - decimals)).mul(SWAP_RATE_PRECISION));
}

  console.log(calculateSwapAmount(8, new BN(1000000), new BN(100000)));
  console.log(calculateSwapAmount(8, new BN(1000000), new BN(100000), true));

// prints
<BN: 64>
<BN: 2710>
```

kev1n-peters: So after some digging I've discovered that bn.js simply doesn't support `pow()` using negative numbers, e.g.
```
> new BN(10).pow(new BN(-2)).toString()
'100'
```

M-Picco: No choice but to leave it as is then

M-Picco (dismissed):

M-Picco (approved):

PR #908, opened 2023-09-01
feature: configurable routes

anondev2323:
- [x] Add routes as a configurable option
       - Check if a route is in the list before performing actions
- [x] Add `getMinAmount` onto routes
- [x] Clean up Route.Bridge/Route.Relay checks across application
       - Left over from when there were only the 2 routes.  Now we check each route dynamically
- [x] Add logic for determining the default route (if bridge is not available)
- [x] Filter networks by available routes
- [x] Only query available routes for available tokens, do not query for a token if it's already in the supported list
- [x] Add validation for route

Bonus:
- [x] Add `.tsx` files to `tsconfig.json` and fix type errors
- [x] Clean up config folder, how we define the constants in `config/index.ts` and validation checks
       - Rename "network" to "chain". Network refers to mainnet, testnet and devnet and chain refers to the individual blockchains
       - Split up chains/tokens/gasEstimates into their own files so they're easier to locate
- [x] Clean up the gas estimation logic, it had gotten kind of out of hand
       - Gas fallback can be accessed using the chainName and route
       - Gas estimates were moved to the chain contexts so that we got rid of all of the `if/else`s
       - If gas cannot be estimated, throw an error.  Catch the error and grab the fallback estimate
- [x] Optimize `allSupportedSourceTokens` and `allSupportedDestTokens`.  Don't need to query every token on every route.  Only need to query if we know it's not supported by another route.
- [x] Fix gas formatting bug which was causing EVM gas estimates to appear as 0


wormhole-connect/src/utils/routes/cctpManual.ts
```diff
@@ -623,7 +634,8 @@ export class CCTPManualRoute extends BaseRoute {
   }: TransferDestInfoParams): Promise<TransferDisplayData> {
     const token = TOKENS[txData.tokenKey];
     const { gasToken } = CHAINS[txData.toChain]!;
-    const gas = await calculateGas(txData.toChain, Route.CCTPManual, receiveTx);
+    // const gas = await calculateGas(txData.toChain, Route.CCTPManual, receiveTx);
+    const gas = 'TODO';
```
anondev2323: Missed this

wormhole-connect/src/utils/routes/cosmosGateway.ts
```diff
@@ -734,11 +741,12 @@ export class CosmosGatewayRoute extends BaseRoute {
...
-      txData.toChain,
-      Route.COSMOS_GATEWAY,
-      receiveTx,
-    );
+    // const gas = await calculateGas(
+    //   txData.toChain,
+    //   Route.CosmosGateway,
+    //   receiveTx,
+    // );
+    const gas = 'TODO';
```
anondev2323: Missed this

wormhole-connect/src/utils/routes/operator.ts
```diff
@@ -141,13 +137,90 @@ export default class Operator {
...
+  async allSupportedDestTokens(
+    sourceToken: TokenConfig | undefined,
+    sourceChain?: ChainName | ChainId,
+    destChain?: ChainName | ChainId,
+  ): Promise<TokenConfig[]> {
+    const supported: { [key: string]: TokenConfig } = {};
+    for (const route of ROUTES) {
+      for (const key in TOKENS) {
+        const alreadySupported = supported[key];
+        if (!alreadySupported) {
```
anondev2323, line 191: We were querying all tokens for every route.  Tried to cut that down a bit by checking if they were already on the list

wormhole-connect/src/views/Bridge/Inputs/AmountInput.tsx
```diff
@@ -23,8 +23,14 @@ function AmountInput(props: Props) {
...
-  function handleAmountChange(event) {
-    let value = event.target.value;
+  function handleAmountChange(
+    e:
+      | React.ChangeEvent<HTMLInputElement>
+      | React.ChangeEvent<HTMLTextAreaElement>
+      | undefined,
+  ) {
+    let value = e!.target.value;
+    // let value = event.target.value;
```
anondev2323: Need to delete random comment

derpy-duck:
' Add logic for determining the default route (if bridge is not available)

Where is this change done?

anondev2323:
> ' Add logic for determining the default route (if bridge is not available)
> 
> Where is this change done?

So it just sets the default to the first route (`ROUTES[0]`).  And then it relies on `establishRoute` in Bridge.tsx.  I also added a config validation to ensure that at least 1 route is enabled.  It was too much of a headache to set the route to a falsey value

anondev2323:
Just realizing, the gas estimate when sending is 0, due to a bug with formatting the amount.  Looks like this was present before, not anything introduced by the refactor.  Will push a fix in a moment

PR #909, opened 2023-09-03
[#903] Don't wait to reaching finality before moving on to the Redeem page

### LINKED ISSUE #903: [Bug] Recovery should work on valid transaction prior to signedVAA being available

- make a transaction from a slow finality chain (like Eth)
- copy the transaction hash
- refresh the page
- from the menu, choose "Resume transaction"
- paste in the tx
- you get an error about that transaction not being a valid bridge transaction

- you go to the redeem flow which will show you the progress of your transaction

## Details

You used to be able to "resume" a transaction that didn't have a VAA yet (the search would call parseMessageFromTx) but now it requires the messageInfo which is basically the signedVAA. I think this functionality likely needs to be added back in, otherwise users cannot get to the screen to wait for perfectly valid transactions that haven't reached finality or ones that are held in the governor (and they won't see that).

### LINKED ISSUE #912: [QA] Redeem screen does not open in 40 min although transaction is successful
Redeem screen does not open in 40 min although transaction is successful.

Checked from https://wormhole-connect.netlify.app/ [Testnet] on macOS version 12.6.7 with Google Chrome Version 116.0.5845.140 (Official Build) (x86_64)

Steps:
1.  Connect Metamask wallets in From and To sections
2. Prepare to send 0.01 WFTM from Base Goerli to Moonbase
3. Click "Approve and proceed with transaction" butoon
4. Observe if transaction is displayed as successful in Metamask wallet
5. Observe if Redeem screen is opened

Send from tx: [BaseScan](https://goerli.basescan.org/tx/0xebacb1587689bf239c5e0d8fd7cd7789e675f3453cf50634415c878bbfada946)

https://github.com/wormhole-foundation/wormhole-connect/assets/10746654/20e9777e-70ff-4bb6-a414-c1f3975c9359

Redeem screen is opened.

M-Picco:
Should address #903.

Revert some changes made during the [routes refactor](https://github.com/wormhole-foundation/wormhole-connect/pull/710) which caused the Bridge UI to wait for a VAA to be emitted before proceeding to the Redeem stage (in some chains it'd take a long time, for example 30 minutes on Ethereum).

The split to `parseMessageFromTx` into `getVaa` and `parseMessage` was reverted back, now under the name `getMessage`.

Adapted the logic and interfaces to include the CCTP routes, which were developed after the routes refactor.

### DIFF
--- a/sdk/src/contexts/aptos/context.ts
+++ b/sdk/src/contexts/aptos/context.ts
@@ -8,36 +8,26 @@ import {
   ParsedMessage,
   Context,
   ParsedRelayerPayload,
-  VaaInfo,
 } from '../../types';
 import { WormholeContext } from '../../wormhole';
 import { TokenBridgeAbstract } from '../abstracts/tokenBridge';
 import { AptosContracts } from './contracts';
 import { AptosClient, CoinClient, Types } from 'aptos';
 import {
-  CHAIN_ID_APTOS,
   getForeignAssetAptos,
   getIsTransferCompletedAptos,
-  getSignedVAAWithRetry,
   getTypeFromExternalAddress,
   hexToUint8Array,
   isValidAptosType,
   parseTokenTransferPayload,
-  parseVaa,
   redeemOnAptos,
   transferFromAptos,
 } from '@certusone/wormhole-sdk';
-import {
-  arrayify,
-  hexZeroPad,
-  hexlify,
-  stripZeros,
-  zeroPad,
-} from 'ethers/lib/utils';
+import { arrayify, hexlify, stripZeros, zeroPad } from 'ethers/lib/utils';
 import { sha3_256 } from 'js-sha3';
 import { MAINNET_CHAINS } from '../../config/MAINNET';
 import { SolanaContext } from '../solana';
-import { ForeignAssetCache, stripHexPrefix } from '../../utils';
+import { ForeignAssetCache } from '../../utils';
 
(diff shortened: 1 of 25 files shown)


sdk/src/contexts/aptos/context.ts
```diff
@@ -264,46 +254,11 @@ export class AptosContext<
...
-    const { vaaBytes } = await getSignedVAAWithRetry(
-      this.context.conf.wormholeHosts,
-      CHAIN_ID_APTOS,
-      emitter,
-      sequence,
-      undefined,
-      undefined,
-      this.context.conf.wormholeHosts.length,
+    const parsed = parseTokenTransferPayload(
+      Buffer.from(payload.slice(2), 'hex'),
```
kev1n-peters, line 260: In the future we should probably be passing `Uint8Array` for binary data so we don't have to do this slicing business (which can be easy to forget, not obvious when it's needed), I know that it was like this before, but just want to jot down my thoughts

kev1n-peters:
Tested resuming a TX from Goerli and it works, nice!

sdk/src/contexts/aptos/context.ts
M-Picco: I agree, it looks kinda flaky. I think `arrayify(payload, { allowMissingPrefix: true })` would be better

kev1n-peters (approved):

Issue #914, opened 2023-09-04
[Bug] WH Connect shows the transfer as complete but the funds have not yet reached the wallet


M-Picco:
There might be a window of time between the time the transfer is redeemed on wormchain and the IBC transfer that eventually deposits the funds on the destination gets completed.

PR #915, opened 2023-09-04
[#913][#914] Add logic to fetch the redeem transaction for cosmos gateway

### LINKED ISSUE #913: [Bug] Missing explorer link on destination chain for gateway transfer
Currently there's no logic to retrieve the final tx on the destination chain (e.g. osmosis)

M-Picco:
This PR adds the logic to retrieve the destination IBC transfer tx given a wh transfer. It also modifies the `isTransferComplete` method to check for this transaction to exist to avoid cases in which the user might receive the funds instantly when the wormchain redeem is completed.

Closes #913, #914

Depends on #909

### DIFF
--- a/wormhole-connect/src/utils/routes/cosmosGateway.ts
+++ b/wormhole-connect/src/utils/routes/cosmosGateway.ts
@@ -638,9 +638,7 @@ export class CosmosGatewayRoute extends BaseRoute {
     };
   }
 
-  async fetchRedeemedEvent(
-    message: SignedMessage,
-  ): Promise<string | null> {
+  async fetchRedeemedEvent(message: SignedMessage): Promise<string | null> {
     if (!isSignedWormholeMessage(message)) {
       throw new Error('Signed message is not for gateway');
     }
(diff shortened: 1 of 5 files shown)


wormhole-connect/src/utils/routes/cosmosGateway.ts
```diff
@@ -559,15 +567,19 @@ export class CosmosGatewayRoute extends BaseRoute {
...
-    return isCosmWasmChain(destChain)
-      ? wh.isTransferCompleted(CHAIN_ID_WORMCHAIN, hexlify(message.vaa))
-      : new BridgeRoute().isTransferCompleted(destChain, message);
+
+    if (!isCosmWasmChain(destChain)) {
+      return new BridgeRoute().isTransferCompleted(destChain, message);
+    }
+
+    const destTx = await this.fetchRedeemedEvent(message);
+    return !!destTx;
```
anondev2323, line 530: Is finding the redeem event the only way to tell if a gateway transaction has been completed?

M-Picco: After the transfer is completed in wormchain, there's a time window until the IBC transfer to the destination cosmos chain completes. Until then, the user won't have the funds available in their wallet. I thought it'd be a "harder" criteria than just checking the redeem on wormchain.

Perhaps it's a bit much and the time window can be negligible. But some QA and user reports showed that it can happen.

kev1n-peters (approved):

PR #917, opened 2023-09-04
Add missing connection swap for wallets

M-Picco:
Add missing call to #901 that swaps the wallet aggregator connections


tonyjin:
Successfully tested on testnet:
1. 0.0001 AVAX Fuji -> Osmosis transaction
2. "Bridge more assets"
3. Swapped source/destination
4. 0.0001 AVAX Osmosis -> Fuji

kev1n-peters (approved):

PR #941, opened 2023-09-08
Refactor to move some logic from the CosmosGateway route to the cosmos context

M-Picco:
Start with `getForeignAsset`


sdk/src/contexts/cosmos/context.ts
```diff
@@ -434,19 +529,45 @@ export class CosmosContext<
...
+  private async getTmClient(
+    chain: ChainId | ChainName,
+  ): Promise<Tendermint34Client | Tendermint37Client> {
     const name = this.context.toChainName(chain);
     if (CosmosContext.CLIENT_MAP[name]) {
       return CosmosContext.CLIENT_MAP[name];
     }
 
-    const rpc = this.context.conf.rpcs[name];
+    const rpc = this.context.conf.rpcs[this.context.toChainName(chain)];
```
anondev2323: Can just use name variable defined above

anondev2323 (approved):

sdk/src/contexts/cosmos/context.ts
M-Picco: Missed that. Thanks!

PR #974, opened 2023-09-13
custom page header

derpy-duck:


wormhole-connect/src/config/index.ts
```diff
@@ -31,6 +31,7 @@ const getEnv = () => {
 export const ENV = getEnv();
 export const isMainnet = ENV === 'MAINNET';
 export const sdkConfig = WormholeContext.getConfig(ENV);
+export const pageHeader = config.pageHeader || 'Bridge';
```
evan-gray: 🤔 should we allow integrators to remove the page header?

evan-gray (commented):
Reminder after this is in a release to update the builder.

wormhole-connect/src/config/index.ts
evan-gray: e.g. 
```
export const pageHeader = config.pageHeader === undefined ? 'Bridge' :  config.pageHeader;
```

M-Picco: Maybe leave it as undefined by default and render it only if it is set?

derpy-duck: Just allowed the empty string - I think undefined as default would modify behavior for existing integrators right (perhaps they want 'Bridge' as the title?)

derpy-duck: JK - now the default is the empty string

evan-gray: Agree this is a reasonable default change - not sure anyone _wanted_ "Bridge" but they can set it if they particularly enjoyed it, I guess.

evan-gray (approved):

Issue #976, opened 2023-09-14
[QA] Source chain gas estimate is shown in ETH instead of ETHoptimism


tsadovska:
## Describe the current behavior in detail
Source chain gas estimate is shown in ETH instead of ETHoptimism.

Checked from https://cctp-only--wormhole-connect.netlify.app/ [Mainnet] with macOS version 12.6.7 and Google Chrome Version 116.0.5845.179 (Official Build) (x86_64).

Steps:
1. Select Optimism network in From section
2. Select Ethereum network in To section
3. Connect Metamask wallets in From and To section
4. Prepare to send 0.001 USDC
5. Observe Source chain gas estimate value in Preview section

Note: Gas fee is shown in ETHoptimism on Redeem screen.

<img width="479" alt="Screenshot 2023-09-14 at 10 04 34" src="https://github.com/wormhole-foundation/wormhole-connect/assets/10746654/d44224b9-402d-437c-87ce-6fdca88874a9">

## What is the expected/desired behavior
Source chain gas estimate is shown in ETHoptimism.

derpy-duck:
cc @tonyjin I see why this can be confusing haha, thoughts?

tsadovska:
Fee is shown in ETHarbitrum/ ETHoptimism in "Total fee estimates" field and in ETH in "Source chain gas estimate" field.
Build: https://cctp-only--wormhole-connect.netlify.app/ [Mainnet]


![Screenshot 2023-09-14 at 18 41 18](https://github.com/wormhole-foundation/wormhole-connect/assets/10746654/fe7940e5-ea05-43b0-829a-e30ed85d6436)


![Screenshot 2023-09-14 at 20 18 42](https://github.com/wormhole-foundation/wormhole-connect/assets/10746654/1cb9117c-4799-456b-ad6e-c6784dd2a082)

PR #981, opened 2023-09-14
Fix for #976: clarify which ETH/WETH

derpy-duck:
Addresses #976


wormhole-connect/src/config/mainnet/tokens.ts
```diff
@@ -395,7 +395,7 @@ export const MAINNET_TOKENS: TokensConfig = {
   },
   ETHarbitrum: {
     key: 'ETHarbitrum',
-    symbol: 'ETH',
+    symbol: 'ETH (Arbitrum)',
```
anondev2323, line 398: Symbols can't be more than 6 characters.  You'll need to add a displayName field

PR #991, opened 2023-09-15
Fixes #985 - Fix floating point math for 'receive' calculations

### LINKED ISSUE #985: [QA] Many extra decimals are shown in amount in To section
[QA] Many extra decimals are shown in amount in To section.

Checked from https://cctp-only--wormhole-connect.netlify.app/ [Mainnet] with macOS 12.6.7 and Google Chrome Version 116.0.5845.179 (Official Build) (x86_64).

Steps:
1. Select Ethereum in From section
2. Select Arbitrum in To section
3. Connect Metamask wallets
4. Enter amount 1.12
5. Observe Amount in To section (actual result: 0.1200000000000001)

https://github.com/wormhole-foundation/wormhole-connect/assets/10746654/3c0a6fb2-b4e4-4bd9-b836-a3f8e2b7c705

Correct Amount is displayed in To section.

derpy-duck:
Fixes #985

### DIFF
--- a/wormhole-connect/src/utils/routes/relay.ts
+++ b/wormhole-connect/src/utils/routes/relay.ts
@@ -180,11 +180,11 @@ export class RelayRoute extends BridgeRoute {
     const { toNativeToken, relayerFee } = routeOptions;
 
     // floating point math
-    const DECIMALS = 10 ** 18;
+    const DECIMALS = 10 ** 8;
     return (
-      (sendAmount * DECIMALS -
-        (toNativeToken || 0) * DECIMALS -
-        (relayerFee || 0) * DECIMALS) /
+      (Math.round(sendAmount * DECIMALS) -
+        Math.round((toNativeToken || 0) * DECIMALS) -
+        Math.round((relayerFee || 0) * DECIMALS)) /
       DECIMALS
     );
   }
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/utils/routes/relay.ts
```diff
@@ -176,10 +176,16 @@ export class RelayRoute extends BridgeRoute {
     routeOptions: RelayOptions,
   ): Promise<number> {
     if (!sendAmount) return 0;
+
+    const { toNativeToken, relayerFee } = routeOptions;
+
+    // floating point math
+    const DECIMALS = 10 ** 8;
```
evan-gray, line 183: Is this always 8 decimals? Seems dangerous to hardcode.

evan-gray (changes requested):
This seems like a good time to add some unit tests for these calculations. I’m also concerned by hardcoding the decimals, though maybe could be convinced if there’s a very good reason it’s okay.

wormhole-connect/src/utils/routes/relay.ts
derpy-duck: This is the token bridge route - which only supports up to 8 decimals right (this serves as an upper bound on # of decimals)

evan-gray: ah, I see the division at the bottom (hence why you mentioned "up to").

evan-gray (approved):
As discussed with @derpy-duck, I am firmly against using `number` for these sorts of calculations, but I agree that change will require a larger re-write.

PR #992, opened 2023-09-15
nit: close modal when user clicks outside content

### LINKED ISSUE #920: Nicer UX for exiting network/asset selection popups
It's a little annoying to have to click the "x" in the top right of the screen when selecting network/asset in connect. The UX I'd expect is being able to exit the selection popup by clicking anywhere outside of the main selection UI (so anywhere in the dark background).

For example, in the below screen for selecting a network, a user can only exit the selection popup by clicking the "x" in the top right. Most UIs allow users to exit popups by clicking anywhere outside of the popup content.

connect_13

anondev2323:
Quick fix for #920

### DIFF
--- a/wormhole-connect/src/components/Modal.tsx
+++ b/wormhole-connect/src/components/Modal.tsx
@@ -61,10 +61,16 @@ function Modal({ open, width, closable, children, onClose }: Props) {
   return (
     <Dialog
       open={open}
-      onClose={onClose}
+      onClose={(event, reason) => {
+        console.log(reason);
+        if (reason === 'backdropClick') {
+          onClose();
+        }
+      }}
       sx={{ borderRadius: 8 }}
       fullWidth
       fullScreen
+      onClick={onClose}
       // maxWidth={width}
       // fullScreen={fullScreen}
     >
--- a/wormhole-connect/src/components/Modal.tsx
+++ b/wormhole-connect/src/components/Modal.tsx
@@ -26,7 +26,6 @@ const useStyles = makeStyles<{ width: number }>()((theme: any, { width }) => ({
       margin: '20px auto',
       padding: '24px 12px',
     },
-    background: 'rgba(255,255,255,.02)',
     maxHeight: 'calc( 100vh - 80px )',
   },
   close: {
--- a/wormhole-connect/src/components/Modal.tsx
+++ b/wormhole-connect/src/components/Modal.tsx
@@ -1,5 +1,5 @@
 import { makeStyles } from 'tss-react/mui';
-import React from 'react';
+import React, { useCallback } from 'react';
 import { Dialog, ScopedCssBaseline } from '@mui/material';
 // import { useTheme } from '@mui/material/styles';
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/components/Modal.tsx
```diff
@@ -61,10 +61,16 @@ function Modal({ open, width, closable, children, onClose }: Props) {
...
+      onClose={(event, reason) => {
+        console.log(reason);
+        if (reason === 'backdropClick') {
+          onClose();
+        }
+      }}
       sx={{ borderRadius: 8 }}
       fullWidth
       fullScreen
+      onClick={onClose}
```
evan-gray: onClose is fired on click?

wormhole-connect/src/components/Modal.tsx
```diff
@@ -61,10 +61,16 @@ function Modal({ open, width, closable, children, onClose }: Props) {
   return (
     <Dialog
       open={open}
-      onClose={onClose}
+      onClose={(event, reason) => {
+        console.log(reason);
```
evan-gray: Probably can remove this console log

wormhole-connect/src/components/Modal.tsx
```diff
@@ -61,10 +61,16 @@ function Modal({ open, width, closable, children, onClose }: Props) {
   return (
     <Dialog
       open={open}
-      onClose={onClose}
+      onClose={(event, reason) => {
```
evan-gray: Might want to consider use callback to avoid re-render / re-attaching the event every render.

evan-gray (changes requested):
Is it worth considering simply removing `fullScreen`?

wormhole-connect/src/components/Modal.tsx
evan-gray: @anondev2323 confirmed, this causes the dialog to close even when you click on, like, the search box or the heading, etc.

wormhole-connect/src/components/Modal.tsx
evan-gray: This event never seems to fire for me (after removing the onClick)

evan-gray (dismissed):
Made changes to prevent the onClick from closing when clicking inside the content. Also fixed a double-scrollbar issue.

wormhole-connect/src/components/Modal.tsx
```diff
@@ -26,6 +26,8 @@ const useStyles = makeStyles<{ width: number }>()((theme: any, { width }) => ({
       margin: '20px auto',
       padding: '24px 12px',
     },
+    background: 'rgba(255,255,255,.02)',
```
anondev2323: Maybe it's related to this line.  But I noticed an ever so slight difference in color between the inner/outter modal content.  Should all be the same color

evan-gray: ah, did that to discern the difference (of when clicking would close vs not), but agree with leaving as is for now. can always revisit.

evan-gray (approved):

PR #994, opened 2023-09-15
Fix for 952 - glitch with warning messages

derpy-duck:
This was a tricky one!

Fixed by adding undefined value for available routes

### DIFF
--- a/wormhole-connect/src/utils/transferValidation.ts
+++ b/wormhole-connect/src/utils/transferValidation.ts
@@ -143,7 +143,7 @@ export const validateToNativeAmt = (
 
 export const validateRoute = (
   route: Route | undefined,
-  availableRoutes: string[],
+  availableRoutes: string[] | undefined,
 ): ValidationErr => {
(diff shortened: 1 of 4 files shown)


wormhole-connect/src/utils/transferValidation.ts
```diff
@@ -255,7 +256,8 @@ export const validate = async (dispatch: Dispatch<AnyAction>) => {
     transferInput.token &&
     transferInput.destToken &&
     transferInput.amount &&
-    Number.parseFloat(transferInput.amount) >= 0
+    Number.parseFloat(transferInput.amount) >= 0 &&
+    transferInput.availableRoutes !== undefined
```
anondev2323, line 260: Ah I see. Hm, could you just check the length rather than make it optionally undefined?

anondev2323: Wait, ignore that.  I looked at the issue a bit more closely.  This makes sense

anondev2323 (approved):

PR #995, opened 2023-09-15
Cosmos to cosmos transfers for gateway

M-Picco:
* Support for cosmos to cosmos transfers for gateway enabled chains
 * Upgrade `@certusone/wormhole-sdk` to `v0.10.3`
 * Add chains:
    *  CosmosHub testnet
    * Evmos mainnet/testnet

Depends on #915


wormhole-connect/src/config/mainnet/tokens.ts
```diff
@@ -594,4 +594,19 @@ export const MAINNET_TOKENS: TokensConfig = {
       default: 5,
     },
   },
+  EVMOS: {
+    key: 'EVMOS',
+    symbol: 'EVMOS',
+    nativeChain: 'evmos',
+    tokenId: {
+      chain: 'evmos',
+      address: 'atevmos',
```
mdfalcon104: it should be`aevmos` not `atevmos`

wormhole-connect/src/config/mainnet/tokens.ts
```diff
@@ -594,4 +594,19 @@ export const MAINNET_TOKENS: TokensConfig = {
...
+    },
+    icon: Icon.EVMOS,
+    coinGeckoId: 'evmos',
+    color: '#ed4e33',
+    decimals: {
+      default: 6,
```
mdfalcon104: The decimal should be 18.

wormhole-connect/src/config/mainnet/tokens.ts
M-Picco: Thanks for catching this!

mdfalcon104:
@M-Picco Is the token received on evmos_9001-2 an `erc-20 `token not IBC token?  I tried to bridge USDC to USDC.wh (ibc/0C19171CDC59451F91D2749CDEA63355532DCD5D8904CCBAC4953290E16AB8FD) but my evmos ibc is zero while evmos-erc20 (0x271c84755f1e35549cf4b2511bb56833d8ec9d86) token has the landed here. 

https://escan.live/tx/88B7E39B835A6BA3937A491A3FDA39F03CFACC0BD0C89AC770CD8C867C6D7FB4

Screenshot 2023-09-26 at 14 22 58

M-Picco:
@dlpigpen It might be the case that the IBC denom received through gateway is converted to an ERC20 on Evmos (see the related [protocol docs](https://docs.evmos.org/protocol/modules/erc20) on this). The linked transaction shows that the recipient account received 5 IBC transfer.

mdfalcon104:
> @dlpigpen It might be the case that the IBC denom received through gateway is converted to an ERC20 on Evmos (see the related [protocol docs](https://docs.evmos.org/protocol/modules/erc20) on this). The linked transaction shows that the recipient account received 5 IBC transfer.

Thanks for your answer. How to bridge back from EVMOS evm to Ethereum. Do we need to convert EVMOS erc-20 tokens to IBC token before transferring.

mdfalcon104:
How about this fee for evmos ibc: https://github.com/wormhole-foundation/wormhole-connect/blob/3b5f19813494063533a85fa898e7ee2146239255/wormhole-connect/src/utils/routes/cosmosGateway.ts#L387
I am thinking of the unit fee here: `uosmo`. It should be `aevmos` if the source chain sending is from `EVMOS` Also, the tx does not work on evmos ibc using kelp. Here is the log:

`
Tx running error: Error: Unsupported type: '/ethermint.types.v1.EthAccount'
    at SigningStargateClient.accountFromAny [as accountParser] (accounts.js:92:1)
    at SigningStargateClient.getAccount (stargateclient.js:180:1)
    at async SigningStargateClient.getSequence (stargateclient.js:191:1)
    at async SigningStargateClient.sign (signingstargateclient.js:285:1)
    at async SigningStargateClient.signAndBroadcast (signingstargateclient.js:235:1)
    at async runKeplrTx (keplr.tsx:175:1)
    `
    
The tx generated on evmos seems does work with cosmosjs, with osmosis it works well.

M-Picco:
@dlpigpen First of all thanks for giving a try!

Working on that issue. It seems the cosmjs libraries are not fully compatible with Ethermint chains.

Re: fee denom. I slipped there, going to fix it. However, Keplr/Leap override the fee and set it to the corresponding chain, that's why it hasn't come up earlier.

derpy-duck (dismissed):

M-Picco:
@dlpigpen Hey there. Sorry for the late update. I think the issue with transfers out of Evmos should be solved now, in case you wanted to try again.

derpy-duck (dismissed):

derpy-duck (approved):

PR #1003, opened 2023-09-18
Refactor: get redeem tx logic

anondev2323:
Pending #915 
Originally, the motivation was to use the new `/relays` endpoint from the wormholescan api.  However, I wonder if that might have been for generic relays, because I discovered that the regular endpoint actually now picks up token bridge relays as well.  Still, I think the refactor makes sense to move more of this logic into the routes


anondev2323:
Looking at what's on development, I don't think we need to change anything anymore

PR #1008, opened 2023-09-19
Solana: Create WSOL ATA if it doesn't exist for auto transfers of SOL

kev1n-peters:
The relayer contract expects the WSOL associated token account to exist when transferring native SOL.

### DIFF
--- a/sdk/src/contexts/solana/context.ts
+++ b/sdk/src/contexts/solana/context.ts
@@ -17,6 +17,8 @@ import {
   getAccount,
   createAssociatedTokenAccountInstruction,
   Account,
+  getAssociatedTokenAddressSync,
+  TokenAccountNotFoundError,
 } from '@solana/spl-token';
 import {
   clusterApiUrl,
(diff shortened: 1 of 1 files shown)


sdk/src/contexts/solana/context.ts
```diff
@@ -928,10 +935,28 @@ export class SolanaContext<
...
       const wrapToken = token === NATIVE;
+      if (wrapToken) {
+        const ata = getAssociatedTokenAddressSync(
+          NATIVE_MINT,
+          new PublicKey(senderAddress),
+        );
+        try {
+          await getAccount(this.connection, ata);
+        } catch (e) {
+          const createAccountInst = createAssociatedTokenAccountInstruction(
```
M-Picco: Would it be possible to add a check to make sure the error is due to an inexistent ATA and not some other thing?

kev1n-peters: I agree that this would be more robust. Unfortunately I don't think there's an error code to check against, but we can compare against the error string

kev1n-peters: We could check `e instanceof TokenAccountNotFoundError` but I think that could have issues if the error were thrown from different libraries, going to leave it for now

M-Picco (dismissed):

M-Picco (approved):

PR #1010, opened 2023-09-19
hack: batch evm provider

evan-gray:
testing improvements from switching to JsonRpcBatchProvider


evan-gray:
better handled in #1050

PR #1020, opened 2023-09-21
Disable ATA warning for auto transfers to Solana

kev1n-peters:
The relayer will create the ATA for us

### DIFF
--- a/wormhole-connect/src/store/transferInput.ts
+++ b/wormhole-connect/src/store/transferInput.ts
@@ -80,7 +80,6 @@ export type TransferValidations = {
   route: ValidationErr;
   toNativeToken: ValidationErr;
   foreignAsset: ValidationErr;
-  associatedTokenAccount: ValidationErr;
 };
 
 export interface TransferInputState {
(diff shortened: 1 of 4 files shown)


anondev2323 (dismissed):

kev1n-peters:
Automatic Deposit (one transaction) when associated token account (ATA) doesn't exist - the relayer contract will create it for us:
image
Bridge (two transactions) when ATA doesn't exist - notice how we can still switch between routes and the approve button is not rendered)
image

kev1n-peters:
Previously if the Solana associated token account (ATA) didn't exist and you selected the bridge route (instead of the default, relay route), the ATA warning would appear and you wouldn't be able to change the route without refreshing the page. `RouteOptions` wasn't being rendered, because `isTransferValid` was returning `false` (and it should, because the ATA must exist for manual transfers, but we still want to render the `RouteOptions` in this case). `isTransferValid` no longer checks for the ATA to exist - instead now that takes place in `Bridge.tsx`. This allows us to display the `RouteOptions` even if the ATA doesn't exist and, importantly, disable the approve & send button if the ATA doesn't exist.

M-Picco (approved):

PR #1034, opened 2023-09-26
pre-populate foreign asset cache

evan-gray:
This PR adds `foreignAsset` to the `tokensConfig` in order to avoid looking up wrapped asset addresses we can pre-calculate. It also includes a check which ensures there are no errors in the `foreignAsset` data and alerts when updates can be made to the config.

The check is run in CI, but can be run manually with the following command (please be patient, as it makes requests slowly in order to avoid getting rate limited).
```
cd wormhole-connect && npx ts-node scripts/checkForeignAssetsConfig.ts
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -33,6 +33,18 @@ jobs:
       - run: npm ci
       - run: npm run build -w sdk
       - run: npm test -w wormhole-connect
+  check-token-list:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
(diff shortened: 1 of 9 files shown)


anondev2323 (approved):
Looks great! Tested out a couple scenarios and worked as expected.  Quick question, what happens if the script is rate limited?

evan-gray:
> Looks great! Tested out a couple scenarios and worked as expected. Quick question, what happens if the script is rate limited?

Should throw an error (and fail the action) 🙂 if the script ends up having a problem we can revisit it. it might be worth working each rpc in parallel with a timeout (this is how the monitor works, for example), but 4m for CI check doesn't seem too bad considering. Can play with it later.

Issue #1030, opened 2023-09-26
[QA] Incorrect asset name is displayed on Redeem screen


tsadovska:
## Describe the current behavior in detail
WETH (Optimism) is displayed instead of WETH (Base) on Redeem screen when sending 0.0001 ETH (Base) from Base (Metamask) to Ethereum (Metamask). User receives the correct asset, ETH (Base). 

Checked on https://deploy-preview-1029--wormhole-connect.netlify.app/ [Mainnet] with macOS version 12.6.7 and Google Chrome Version 117.0.5938.92 (Official Build) (x86_64).

Steps:
1. Prepare to send 0.0001 ETH (Base) from Base (Metamask) to Ethereum (Metamask)
2. Click "Approve and proceed with transaction" button
3. Click "Claim" button on Redeem screen

Note:
* Send from tx: [Basescan](https://basescan.org/tx/0x1f23a7e4117217613f66327577f30d6608dcab486b64089def62ee5cabb31d55)
* Send to tx: [Etherscan](https://etherscan.io/tx/0xbf1fad0b0d211d3a689cfadeda21a9e81c0ff7732ed0178eb2c8616b5bda1af6)

![Screenshot 2023-09-26 at 10 55 20](https://github.com/wormhole-foundation/wormhole-connect/assets/10746654/a91cd676-df7d-45fe-a3b9-a1f8d40de736)

## What is the expected/desired behavior
Correct asset name is displayed on Redeem screen.

tsadovska:
Issue is not reproducible on https://wormhole-connect-mainnet.netlify.app/ [Mainnet]. Correct asset name is displayed on Redeem screen. 

Send from tx: [Basescan](https://basescan.org/tx/0xfa0d4ff9dabf3b372c1347370c385eac0e97d7f1097deec2ab2301d29fd2841f)
Send to tx: [Etherscan](https://etherscan.io/tx/0x713796c2afd996505fe5d77061a65fce8da60287368eff79631afeb178f94f3c)

![Screenshot 2023-10-03 at 16 47 41](https://github.com/wormhole-foundation/wormhole-connect/assets/10746654/38573e64-556a-4771-b39c-7e02b96c1d64)


Ticket is closed.

PR #1035, opened 2023-09-26
getTokenById should check chain matches

kev1n-peters:
Addresses #1030

### DIFF
--- a/wormhole-connect/src/utils/index.ts
+++ b/wormhole-connect/src/utils/index.ts
@@ -85,12 +85,13 @@ export function getWrappedTokenId(token: TokenConfig): TokenId {
   return wrapped.tokenId!;
 }
 
-export function getTokenById(tokenId: TokenId): TokenConfig | void {
-  return TOKENS_ARR.filter(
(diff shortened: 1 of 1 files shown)


kev1n-peters:
QA transferred WETH (Base) to Ethereum and the redeem page was showing the redeemed token as WETH (Optimism). getTokenById was returning the wrong token, because the addresses of WETH (Base) and WETH (Optimism) are the same (and getTokenById was only checking if the addresses matched, not also the chains)

wormhole-connect/src/utils/index.ts
```diff
@@ -89,6 +89,7 @@ export function getTokenById(tokenId: TokenId): TokenConfig | void {
   return TOKENS_ARR.filter(
```
evan-gray: Not directed at this change, but I might have used `.find` here.

evan-gray: Also isn't the type `TokenConfig | undefined`?

evan-gray (dismissed):

evan-gray (approved):

PR #1036, opened 2023-09-26
change exports for types to be correct

elee1766:


evan-gray (approved):
thanks for the contribution! confirmed that the types do seem to be in the types folder 🙃

PR #1044, opened 2023-09-28
more consistent spacing

evan-gray:
## without errors

before:
image

after:
image

## with errors

before:
image

after:
image

### DIFF
--- a/wormhole-connect/src/components/InputContainer.tsx
+++ b/wormhole-connect/src/components/InputContainer.tsx
@@ -6,7 +6,7 @@ type StyleProps = { bg?: string };
 const useStyles = makeStyles<StyleProps>()((theme: any, { bg }) => ({
   input: {
     width: '100%',
-    padding: '16px',
+    padding: theme.spacing(1.5),
     borderRadius: '8px',
     backgroundColor: bg || theme.palette.card.background,
     boxShadow: theme.palette.card.elevation,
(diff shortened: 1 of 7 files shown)


wormhole-connect/src/components/InputContainer.tsx
```diff
@@ -6,7 +6,7 @@ type StyleProps = { bg?: string };
 const useStyles = makeStyles<StyleProps>()((theme: any, { bg }) => ({
   input: {
     width: '100%',
-    padding: '16px',
+    padding: theme.spacing(1.5),
```
tonyjin, line 9: How'd you decide between when to set specific values vs when to use `theme.spacing()`?

evan-gray: I used the theme spacing for gap/padding/margin anywhere it was available.

tonyjin: Thinking down the line - might it be tricky for someone to figure out what's controlled via `theme.spacing()` vs `px`. The latter is annoying to track, but I'm thinking about someone who changes theme.spacing(), sees that it doesn't change everything they want, and needing to go through all the `px` ones anyways.

If you think there's a way to get to relying on `theme` everywhere down the line then this makes sense.

tonyjin (approved):
Added a comment about `px` vs `theme.spacing()`. Don't feel too strongly about it so lgtm based on screenshots. Thanks for making this!

wormhole-connect/src/components/InputContainer.tsx
evan-gray: Sorry, should have added that to the description - I think we should consider moving all the spacing to `theme.spacing` or the `sx` prop so that we can support "comfortable" vs "compact" with 1 setting. all of these being explicitly set like this would never have been properly customizable by integrators.

PR #1063, opened 2023-10-04
improve aptos getTokenBalances

evan-gray:
This PR leverages the Aptos indexer to query coin balances.

Most of this code is the plumbing to add a new config for aptos indexer. I don't expect integrators to have to override the default.

### DIFF
--- a/sdk/src/contexts/aptos/context.ts
+++ b/sdk/src/contexts/aptos/context.ts
@@ -28,9 +28,18 @@ import { sha3_256 } from 'js-sha3';
 import { MAINNET_CHAINS } from '../../config/MAINNET';
 import { SolanaContext } from '../solana';
 import { ForeignAssetCache } from '../../utils';
+import axios from 'axios';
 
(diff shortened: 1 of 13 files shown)


sdk/src/contexts/aptos/context.ts
```diff
@@ -355,15 +364,25 @@ export class AptosContext<
...
+    const limit = 100;
+    while (true) {
+      const result = await this.fetchCurrentCoins(walletAddress, offset, limit);
+      coinBalances = [...coinBalances, ...result.data.current_coin_balances];
+      if (result.data.current_coin_balances.length < limit) {
+        break;
+      }
+      offset += result.data.current_coin_balances.length;
+    }
+    console.log(coinBalances);
```
anondev2323: Remove log

anondev2323 (dismissed):
Looking good!

sdk/src/contexts/aptos/context.ts
evan-gray: oops! thanks 🙂

anondev2323 (approved):

PR #1082, opened 2023-10-05
solana gas dropoff rent min to fresh wallets

kev1n-peters:
A fresh wallet cannot be transferred less SOL than the minimum rent exemption. The relayer encounters an error when the UI allowed for a gas dropoff which was too low to a fresh account. Now, the UI will enforce the rent exemption minimum if the wallet does not exist 🙂👍

image


evan-gray (approved):

evan-gray:
idk what GH's deal is, but this is #1075's replacement

PR #1086, opened 2023-10-05
[Refactor] Optimizations for cosmos getTokenBalances

M-Picco:
* Reduce the amount of calls to fetch balances to just one for all instead of one per asset
 * Fix issue where the foreign asset cache would not be populated for other chains than wormchain
 * Reduce the overhead created by multiple promises trying to create a Tendermint Client when doing it once would be enough. Leverage the single-threaded nature of JS to control the access to the creation logic through a hacky pseudo-condvar and a static variable.


evan-gray (approved):
nice. definitely reduced that Osmosis balance spam!

PR #1087, opened 2023-10-05
builder: automatic updates

evan-gray:


anondev2323:
Nit: the widget goes over the tab on the right hand corner
Screenshot 2023-10-06 at 1 52 50 PM

anondev2323:
How come only the pinned version has `integrity` and `crossorigin` values? Is that intentional or should they all have it? 
Screenshot 2023-10-06 at 1 54 56 PM
Screenshot 2023-10-06 at 1 55 00 PM

evan-gray:
> How come only the pinned version has `integrity` and `crossorigin` values? Is that intentional or should they all have it? Screenshot 2023-10-06 at 1 54 56 PM Screenshot 2023-10-06 at 1 55 00 PM

This is required, as (by its nature) the integrity of anything pulled by tag _will_ change - if the integrity was set, as soon as there was a new release, the browser would no longer load the resource. It was suggested by one of the security-conscious contributors to set the integrity for the pinned version so the browser would enforce that it is, indeed, always the same version.

Setting `integrity` requires explicitly setting `crossorigin` which is why I set it in this case.

anondev2323 (approved):

PR #1089, opened 2023-10-06
Update RPC warning and documentation

tonyjin:
1) Update custom RPC console.error warning -> console.info 
2) Update NPM README for custom RPCs
3) Fix inconsistent rpcs/rpc configs and types

Reviewers:
- Please check & provide feedback for updated language
- Please check whether I changed `rpcs` -> `rpc` in the right locations. Since our public config interface uses `rpc`, I updated `rpcs` -> `rpc` where it made sense.
- How can we only show the informational console log during development and not in production builds? IMO, we should only show log critical warnings in production builds and avoid surfacing integrator feedback to end users

### DIFF
--- a/wormhole-connect/src/config/types.ts
+++ b/wormhole-connect/src/config/types.ts
@@ -57,7 +57,7 @@ export interface BridgeDefaults {
 // TODO: move to a shared package
 export interface WormholeConnectConfig {
   env?: 'mainnet' | 'testnet' | 'devnet';
-  rpcs?: ChainResourceMap;
+  rpc?: ChainResourceMap;
   rest?: ChainResourceMap;
   networks?: ChainName[];
   tokens?: string[];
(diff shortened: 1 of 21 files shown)


wormhole-connect/src/config/types.ts
```diff
@@ -57,7 +57,7 @@ export interface BridgeDefaults {
 // TODO: move to a shared package
 export interface WormholeConnectConfig {
   env?: 'mainnet' | 'testnet' | 'devnet';
-  rpcs?: ChainResourceMap;
+  rpc?: ChainResourceMap;
```
evan-gray, line 60: This would be a breaking change

evan-gray (commented):
What was the motivation for changing rpcs to rpc?

evan-gray (changes requested):
Updating to request changes. As written this would break existing integrators. Throughout the code `rpcs` is correctly used, whereas one line of the readme is incorrect. I suggest we fix the one line in the readme instead of making a breaking change.

evan-gray:
agree with switching the error to info. unfortunately (for this case), the build pulled by integrators is pre-built - meaning that it is already in "production" mode. Short of some pass-through hackery in the loader react component, this won't work like a typical react library.

also, confirmed that the working flag is, indeed, `rpcs`. I'll open a PR to address the README.md
image
image
image

PR #1091, opened 2023-10-06
Aptos rpc errors

anondev2323:
We were calling `registerProvider` on the Aptos RPC upon initialization.  This attempted to create an ethers JSON rpc provider and call `detectNetwork`.  This resulted in a 405 error because it is not compatible with the Aptos rpc.  Resolved by only registering EVM chain providers.

TODO:
- [ ] Still investigating CORS errors


evan-gray (approved):

evan-gray:
@anondev2323 I think this is a good change as is since the hitting endpoints that aren't evm was bothering me 😅 feel free to open another to investigate errors.

PR #1093, opened 2023-10-06
refactor: organize routes

anondev2323:
The Routes code has grown a lot, so trying to organize it a bit more logically.

1. Move it outside of utils.  It's a major component, so I think it deserves a top-level spot
2. Create a folder for each route
3. Separate RouteAbstract into RouteAbstract and RelayAbstract
4. Separate types, utility methods and methods that don't need to be on the route class


wormhole-connect/src/routes/abstracts/relayAbstract.ts
```diff
@@ -0,0 +1,27 @@
...
+    walletAddress: string,
+  ): Promise<BigNumber>;
+
+  abstract maxSwapAmount(
+    destChain: ChainName | ChainId,
+    token: TokenId,
+    walletAddress: string,
+  ): Promise<BigNumber>;
+
+  abstract tryFetchRedeemTx(
```
evan-gray: The other two functions were removed from routeAbstract, but this one is in both - is that intentional?

wormhole-connect/src/routes/bridge/bridge.ts
```diff
@@ -382,22 +382,22 @@ export class BridgeRoute extends BaseRoute {
...
-  }
-
-  async maxSwapAmount(
-    destChain: ChainName | ChainId,
-    token: TokenId,
-    walletAddress: string,
-  ): Promise<BigNumber> {
-    throw new Error('Not implemented');
-  }
+  // async nativeTokenAmount(
```
evan-gray: can these be deleted instead of commented out?

wormhole-connect/src/routes/cctpManual/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './cctpManual';
+export * from './utils';
```
evan-gray, line 2: One thing I worry about with always using `export *` is that it won't be clear when identically named exports will get clobbered lower down (like in `routes/index.ts`).

wormhole-connect/src/routes/cosmosGateway/utils/fromCosmos.ts
```diff
@@ -0,0 +1,96 @@
...
+      sender: senderAddress,
+      receiver: getTranslatorAddress(),
+      token: coin,
+      timeoutTimestamp: millisToNano(Date.now() + IBC_TIMEOUT_MILLIS),
+      memo,
+    }),
+  };
+
+  const tx: CosmosTransaction = {
+    fee: calculateFee(1000000, '1.0uosmo'),
```
evan-gray: just flagging that this will probably need to be extended with future chain support. don't have to solve now though

evan-gray (changes requested):

wormhole-connect/src/routes/bridge/bridge.ts
anondev2323: Missed this thanks

wormhole-connect/src/routes/abstracts/relayAbstract.ts
anondev2323: Good catch, should be removed

wormhole-connect/src/routes/cctpManual/index.ts
anondev2323: fair, not sure how I'd want to tackle that.  Gonna resolve for now and we can come back to it later

evan-gray (approved):

PR #1095, opened 2023-10-06
import ethers contracts from esm instead of cjs to reduce bundle size

kev1n-peters:
Bundle size before:
Screenshot from 2023-10-06 11-22-48
Bundle size after:
Screenshot from 2023-10-06 11-23-03
We shave off ~500KB in gzipped bundle size

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -44,7 +44,7 @@ jobs:
           cache: "npm"
       - run: npm ci
       - run: npm run build -w sdk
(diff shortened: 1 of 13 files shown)


anondev2323 (dismissed):

kev1n-peters:
bundle size is down to 3.6MB from 5MB 
image

evan-gray (approved):
Confirmed - the new package in preview is 3.5 MB transferred, 16.7 MB resource size. Old was 4.9 MB / 23.9 MB.
Nice work!

PR #1097, opened 2023-10-08
Fix route for Sei

M-Picco:
Sei should go through the bridge route. Also update the wallet aggregator sei dependency


M-Picco:
Closing this in favor of #1109 which addresses the root issue

PR #1109, opened 2023-10-12
Refactor: introduce isGatewayChain in favor of isCosmWasmChain

M-Picco:
This will prevent conflicts for future cosmos chains that should use the bridge route.


evan-gray (approved):
Nice change! Should fix some bugs

evan-gray (approved):
Nice change! Should fix some bugs

PR #1132, opened 2023-10-18
Switch cosmos chain when changing the from/to network

M-Picco:
Cosmos accounts have different address format for each chain (e.g. `osmo1lez...`, `evmos1q9x...`) derived from a public key and a prefix. When changing the network, the asset balances are fetched for the new chain, but requests will fail due to the invalid format of the queried account (e.g. Evmos will return an error when querying balances for `osmo1lez...`).

This PR makes it so that when changing cosmos networks it will request the wallet for a network switch to the new chain and update the store with the new address. I tried decoding the address back to the public key and reencoding for the new chain, but evm cosmos chains use a different account scheme so re-encoding is not correct for some transitions (e.g. Osmosis to Evmos).

Should address #1136

### DIFF
--- a/wormhole-connect/src/store/wallet.ts
+++ b/wormhole-connect/src/store/wallet.ts
@@ -87,11 +87,12 @@ export const walletSlice = createSlice({
       const { type, error } = payload;
       state[type].error = error;
     },
-    setCurrentAddress: (
+    setAddress: (
       state: WalletState,
       { payload }: PayloadAction<{ type: TransferWallet; address: string }>,
     ) => {
       const { type, address } = payload;
+      state[type].address = address;
       state[type].currentAddress = address;
     },
     clearWallets: (state: WalletState) => {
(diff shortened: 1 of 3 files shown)


wormhole-connect/src/store/wallet.ts
```diff
@@ -87,11 +87,12 @@ export const walletSlice = createSlice({
       const { type, error } = payload;
       state[type].error = error;
     },
-    setCurrentAddress: (
+    setAddress: (
       state: WalletState,
       { payload }: PayloadAction<{ type: TransferWallet; address: string }>,
     ) => {
       const { type, address } = payload;
+      state[type].address = address;
```
kev1n-peters, line 95: What is the difference between `address` and `currentAddress` in the store?

M-Picco: I think it's for cases where the wallet may have different available addresses (I saw some validations using it). However, I only see it set once when connecting and it's never modified again. I didn't want to remove it in case it breaks something, but it may be that the validation is no longer being performed and it's become redundant.

kev1n-peters (approved):

PR #1155, opened 2023-10-24
token approval checks

### LINKED ISSUE #1101: [Enhancement] Show warning when ERC20 token allowance is too low

We should address https://github.com/XLabs/portal-bridge-ui/issues/461 in Connect.

Repro:
 1. Attempt to send a token from an EVM network that you have not previously set a token allowance for (or "spending cap" as MetaMask calls it). You can view your current allowances at revoke.cash and entering in your address
 2. Initiate the transfer
 3. Note that your wallet will pop up with an approval to set a spending cap
 4. Set this spending cap to below what the default amount is (which will be prefilled with the amount you're trying to send from the first step)
 5. Once that transaction completes, you'll see "Error sending transfer, please try again", and the debug console will show an error like
 ```
 main.e527c64d.js:137 Error: call revert exception; VM Exception while processing transaction: reverted with reason string "insufficient allowance" [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="transferTokens(address,uint256,uint16,bytes32,uint256,uint32)", data="0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000016696e73756666696369656e7420616c6c6f77616e636500000000000000000000", errorArgs=["insufficient allowance"], errorName="Error", errorSignature="Error(string)", reason="insufficient allowance", code=CALL_EXCEPTION, version=abi/5.7.0)
 ```
Screenshot 2023-10-09 at 12 26 12 PM

Error message should say: "Error due to insufficient token allowance, please fix and try again"
Screenshot 2023-10-09 at 12 33 21 PM

I'm also proposing that we change the default message to "Error with transfer, please try again". Sending + transfer seems redundant.

evan-gray:
Metamask allows users to change the token allowance. This may result in users who end up setting the allowance smaller than the transfer. This change explicitly checks that the allowance is at least large enough for the transfer, otherwise it will show a more specific error.

image

My only hesitancy about performing the check directly after the `.wait()` is that it is theoretically possible to hit a node 
(via the UI) on the subsequent request that is behind the node (via the wallet) which confirmed the tx. In this case, the error may be displayed when the user had approved the correct amount. Still, I didn't run into this given a few tests and I'm not sure the likeliness vs resolving the identified issue.

### DIFF
--- a/sdk/src/contexts/eth/context.ts
+++ b/sdk/src/contexts/eth/context.ts
@@ -34,6 +34,7 @@ import { arrayify } from 'ethers/lib/utils';
 import { ForeignAssetCache, chunkArray } from '../../utils';
 
 export const NO_VAA_FOUND = 'No message publish found in logs';
+export const INSUFFICIENT_ALLOWANCE = 'Insufficient token allowance';
 
 export class EthContext<
   T extends WormholeContext,
@@ -288,6 +289,15 @@ export class EthContext<
         overrides,
       );
       await tx.wait();
+      // Metamask allows users to set a different amount than specified above
+      // Check to make sure that the amount set was at least the requested amount
+      const nowApproved = await tokenImplementation.allowance(
+        senderAddress,
+        contractAddress,
+      );
+      if (nowApproved.lt(approveAmount)) {
+        throw new Error(INSUFFICIENT_ALLOWANCE);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


barnjamin (dismissed):

sdk/src/contexts/eth/context.ts
```diff
@@ -288,6 +289,13 @@ export class EthContext<
         overrides,
       );
       await tx.wait();
+      const nowApproved = await tokenImplementation.allowance(
```
barnjamin, line 294: can you add a comment here on why this check is performed?

barnjamin (approved):

barnjamin (approved):

M-Picco (approved):

Issue #952, opened 2023-09-12
[QA]Glitch with warning messages


AigaD:
Preconditions: 
Metamask wallet connected

Environment: MacBook Pro, macOS Ventura 13.0.1, Chrome Version 115.0.5790.170

Tested from: https://feature-configurable-routes--wormhole-connect.netlify.app/ 

Scenario 1
Steps:
1. Connect Fuji Network at sending From section
2. Connect Moonbase Network at sending To section
3. Click on Assets and select Avax
4. In the amount field start typing 0.02
5. Observe the warning message shown below sending To section

Actual result: For a brief moment message "No bridge or swap route available for selected tokens" is shown

Expected result: No warning message is shown for tokens that have available swap route

Attachment:

<img width="1183" alt="Screenshot 2023-09-12 at 12 09 27" src="https://github.com/wormhole-foundation/wormhole-connect/assets/91047061/530ee341-8efa-4cfb-8cd3-38f310471a60">

https://github.com/wormhole-foundation/wormhole-connect/assets/91047061/689c144b-4b06-4872-a630-264d672f6be3

Scenario 2
Steps:
1. Connect Fantom Network at sending From section
2. Connect Sui Network at sending To section
3. Click on Assets and select FTM
4. In the amount field start typing 0.02
5. Observe the warning message regarding minimum amount

Actual result: Minimum amount warning message glitches

Expected result: Minimum amount warning message is shown and stays until user takes next action

Attachment:

https://github.com/wormhole-foundation/wormhole-connect/assets/91047061/f3c6e4fb-cc5b-4ed4-992e-1169c8264d11

AigaD:
Reproducible from: https://cctp-only--wormhole-connect.netlify.app/#
When sending USDC asset from Arbitrum message "No bridge or swap route available for selected tokens" is shown
And https://latest--wormhole-connect.netlify.app/ when 0.01 MATIC was sent from Polygon to BSC
Attachment:
<img width="1162" alt="Screenshot 2023-09-14 at 11 02 58" src="https://github.com/wormhole-foundation/wormhole-connect/assets/91047061/c8b76c08-65aa-46db-b0df-daa52b04d857">

AigaD:
Minimum amount warning message shown for a brief moment for TX from Sei to Fantom
Tested from: https://latest--wormhole-connect.netlify.app/#
Attachment:

[...]

PR #1162, opened 2023-10-25
connect: better state syncing / debouncing

### LINKED ISSUE #868: [QA] Neither Approve button nor token registration warning are present when transaction details are entered
"Approve and proceed with transaction" button is not present when sending asset from Osmosis to Sei.

Checked from https://feat-cosmos-gateways--wormhole-connect.netlify.app/ [Testnet] with macOS version 12.6.7 and Google Chrome Version 116.0.5845.110 (Official Build) (x86_64).
 
Steps:
1. Select Osmosis network in "From" section
2. Select Sei network in "To" section
3. Select Leap for Cosmos wallet in "From" section
4. Select Leap for Sei wallet in "To" section
5. Prepare to send 0.1 CELO
6. Observe if "Approve and proceed with transaction" button is present 

https://github.com/wormhole-foundation/wormhole-connect/assets/10746654/6be933ed-ef7e-49d5-b558-e3a5b47229ce

Note:
Issue is also reproducible when sending 0.01 WMATIC from Osmosis (Leap) to Sei (Compass).

 
image

"Approve and proceed with transaction" button is present.

evan-gray:
This PR addresses several bugs related to the state syncing of validation checks and state-based calculations.

## Change 1 - Consistent Validation

After Initially debugging #868, I found the cause was due to `validate` not getting triggered after `setAvailableRoutes`. The `availableRoutes` dependency was not included in the dependency array, but furthermore the `route` would not be "established" until `validate` was triggered with "all of the fields filled out". For now, I kept the one-shot triggering of showing validation errors (where the errors won't show until after the first time the form is filled out), but switched it to take the state dependencies as parameters and created a more appropriate hook to trigger it. This hook debounces the relevant state so that, in a future PR, we can hide errors or display a loader while the validation is waiting to be executed.

### Before

`validate` was not firing after the resolution of the available routes.

Screenshot from 2023-10-24 21-17-21

### Proper `useValidate`, without debounce

Setting the correct dependencies in the hook caused several more updates, but fired correctly.

Screenshot from 2023-10-24 21-20-11

### `useValidate` with debounce

Debouncing the state used by the `validate` method reduced the calls back to the previous number + 1, as expected. Now that this debouncing was in place and handled properly calling `validate` on any dependent state updates, the `onPause` effect of the amount input was no longer needed.

Screenshot from 2023-10-24 21-30-14

## Change 2 - Simplify `establishRoute`

Selection of a route was handled by a hook with an overly broad dependency array. Furthermore, the only effective change was when the available routes changed. To avoid two renders, it appeared more effective to move this calculation into the state update of `setAvailableRoutes`.

## Change 3 - Reduce network spam on amount change

As I was testing, I was typing in the amount and every key press was firing off network calls. I identified two culprits:

1. `setSendingGas` - the source chain gas estimation
2. Calculations in the native gas slider

Switching both to a debounced amount cleared up the network calls.

## Future improvements

I believe this PR is an improvement over the previous situation, in that the state more consistently syncs / validates / fires effects and more of the network-call-inducing effects are debounced. However, the user is still subjected to every intermediate state / error while these resolve. Ideally, these intermediate state transitions / errors would be hidden from the user until all of the calls are settled. This could be done by debouncing the errors that are displayed, setting some loading flags, comparing the debounced and non-debounced values, or some combination of these as appropriate.

### DIFF
--- a/wormhole-connect/src/store/transferInput.ts
+++ b/wormhole-connect/src/store/transferInput.ts
@@ -87,7 +87,7 @@ export type TransferValidations = {
 };
 
 export interface TransferInputState {
-  validate: boolean;
+  showValidationState: boolean;
   validations: TransferValidations;
   availableRoutes: string[] | undefined;
   fromChain: ChainName | undefined;
@@ -112,7 +112,7 @@ export interface TransferInputState {
 }
 
 const initialState: TransferInputState = {
-  validate: false,
+  showValidationState: false,
   validations: {
     fromChain: '',
     toChain: '',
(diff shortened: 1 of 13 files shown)


wormhole-connect/src/store/transferInput.ts
```diff
@@ -182,28 +182,62 @@ const performModificationsIfToChainChanged = (state: TransferInputState) => {
   }
 };
 
+// This function was relocated from a useEffect in the Bridge component to avoid
+// the visual delay of two store updates.
+// At the time, the dependency array included:
+// [showValidationState, availableRoutes, fromChain, toChain, token, destToken, amount]
```
M-Picco, line 188: Good catch. I'd guess the reason this deps array included these things beyond `availableRoutes` must be to trigger the route selection whenever they are changed, but these same dependencies are already on the hook that sets `availableRoutes`, so it was kind of redundant.

M-Picco (approved):
Nice work. Thanks for the thorough description.

Nothing to add beyond the intermediate errors showing in succession (I triggered it by setting an amount for a relay transfer and then changing it to below the threshold), but you already mention the debouncing for a future PR.

evan-gray:
> Nice work. Thanks for the thorough description.
> 
> Nothing to add beyond the intermediate errors showing in succession (I triggered it by setting an amount for a relay transfer and then changing it to below the threshold), but you already mention the debouncing for a future PR.

Definitely, hope to tackle #952 soon!

PR #1173, opened 2023-10-26
connect: update foreign assets cache and script to include gateway chains

M-Picco:
* Update foreign asset caches for some tokens
 * Add some checks to the `checkForeignAssetsConfig` script to skip errors when querying gateway chains

### DIFF
--- a/wormhole-connect/scripts/checkForeignAssetsConfig.ts
+++ b/wormhole-connect/scripts/checkForeignAssetsConfig.ts
@@ -60,7 +60,9 @@ const checkEnvConfig = async (
             } catch (e: any) {
               if (
                 e?.message === '3104 RPC not configured' ||
-                e?.message === 'wormchain RPC not configured'
+                e?.message === 'wormchain RPC not configured' ||
+                e?.message ===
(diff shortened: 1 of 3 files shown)


wormhole-connect/scripts/checkForeignAssetsConfig.ts
```diff
@@ -60,24 +60,39 @@ const checkEnvConfig = async (
             } catch (e: any) {
               if (
                 e?.message === '3104 RPC not configured' ||
-                e?.message === 'wormchain RPC not configured'
+                e?.message === 'wormchain RPC not configured' ||
```
sebastianscatularo: maybe you could move these checks out to a function to make evident why those errors got ignored and make it maintainable

```suggestion
/**
 * check if the give error message match a well know wormchain error message
 */
function isWormchainError({ message }: Error) {
  const messages = ['3104 RPC not configured', 'wormchain RPC not configured', 'wormchain RPC not configured'];
  return messages.includes(message);
}
```

sebastianscatularo (approved):

PR #1188, opened 2023-10-30
#1183 - add config option to mount to search tx view

### LINKED ISSUE #1183: [connect > portal] - Add option to mount directly to resume tx view

sebastianscatularo:


wormhole-connect/src/config/index.ts
```diff
@@ -19,7 +19,13 @@ const el = document.getElementById('wormhole-connect');
 if (!el)
   throw new Error('must specify an anchor element with id wormhole-connect');
 const configJson = el.getAttribute('config');
-export const config: WormholeConnectConfig = JSON.parse(configJson!) || {};
+export const config: WormholeConnectConfig = JSON.parse(configJson!) || {
+  searchTx: {
```
kev1n-peters: I think we want to remove this, because when we open up the netlify preview it brings us to the Resume transaction page

kev1n-peters (commented):
I'm a little confused about the intention of this PR. I think instead of specifying the search tx in the config it should be passed as an argument in the URL. Sorry if I'm misunderstanding something here

wormhole-connect/src/config/index.ts
evan-gray: Explicitly requesting changes on this. Similar to my comment on the other PR, I don't think we want to be changing the default config for everyone like this.

evan-gray (dismissed):
Is the desire here to have the widget load to the search tx view? I'm not sure we'll want to have the widget mess with the URL considering we don't know how the outer page handles routing. Maybe there's a more direct way to have the widget hook into that.

wormhole-connect/src/config/types.ts
```diff
@@ -74,8 +74,36 @@ export interface WormholeConnectConfig {
...
+export type SearchTxConfig = {
+  txHash?: string;
+  chainName?: ChainName;
+};
+export type MoreTokenConfig = {
+  label: string;
+  href: string;
+};
+
+export type ExtraChainConfig = {
```
M-Picco: Should this extra chain/token types be on this PR?

wormhole-connect/src/config/index.ts
M-Picco: Perhaps it was a left over from testing and was comitted without intention

sebastianscatularo: yes, this needs to be removed before we merge to development, it was for local testing and some integration test that I'm trying to archive, I'll remove it once I'm done

wormhole-connect/src/config/types.ts
sebastianscatularo: oh, good call, thanks

wormhole-connect/src/config/index.ts
sebastianscatularo: removed

sebastianscatularo:
> I'm a little confused about the intention of this PR. I think instead of specifying the search tx in the config it should be passed as an argument in the URL. Sorry if I'm misunderstanding something here

The idea is to allow the outer container to mount directly to the search view, as Connect is a widget, not sure if we want to add URL interactions, but, I think that could be another solution / approach

kev1n-peters (approved):

PR #1195, opened 2023-11-02
[bugfix] Address failure to fetch balances when the config sets a custom tokens array

M-Picco:
Use the TOKENS_ARR to compute available tokens so that the resulting array is filtered for the specific configured tokens.

Setting a custom config with a tokens array (e.g: `{ "tokens": [ "WETH", "WETHoptimism", "WETHarbitrum", "USDCarbitrum", "AVAX", "WAVAX" ] }` would cause issues with fetching balances since the behaviour was to fetch for all tokens and not only those configured. Since part of the fetching process would involve looking up the `TOKENS_ARR` (done [here](https://github.com/wormhole-foundation/wormhole-connect/blob/21ba498910922b4468f155d3668a530ce525f89f/wormhole-connect/src/utils/index.ts#L91)), it would fail due to not finding the token config for it.

### DIFF
--- a/wormhole-connect/src/routes/operator.ts
+++ b/wormhole-connect/src/routes/operator.ts
@@ -6,7 +6,7 @@ import {
 } from '@wormhole-foundation/wormhole-connect-sdk';
 import { BigNumber } from 'ethers';
 
-import { CHAINS, ROUTES, TOKENS } from 'config';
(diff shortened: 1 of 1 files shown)


derpy-duck (approved):
looks good to me

PR #1198, opened 2023-11-02
Gas dropoff slider conversion rate fix

### LINKED ISSUE #1076: [Bug] The gas dropoff slider conversion ratio doesn’t match the on-chain ratio
It seems like the conversion ratio for the slider preview is calculated based on coingecko prices but the real conversion calculation is on-chain in the relay contract, which is used after the slider is released. This leads to the numbers significantly changing after the slider is released. The on chain calculation should be used for both.

kev1n-peters:
Previously the gas dropoff slider was setting the native gas amount using the conversion rate from CoinGecko. However the conversion rate from CoinGecko may not be the same as the conversion rate used by the relayer contract. This commit fixes the gas dropoff slider to use the conversion rate from the relayer contract instead. A side effect of this change is that the native gas amount is not displayed in the UI until the conversion rate is fetched from the relayer contract, after the debounced value from the slider changes.

Addresses #1076

### DIFF
--- a/wormhole-connect/src/views/Bridge/NativeGasSlider.tsx
+++ b/wormhole-connect/src/views/Bridge/NativeGasSlider.tsx
@@ -282,11 +282,9 @@ function GasSlider(props: { disabled: boolean }) {
   const handleChange = (e: any) => {
     if (!amountNum || !state.conversionRate) return;
     const value = e.target.value < state.min ? 0 : e.target.value;
-    const newGasAmount = value * state.conversionRate;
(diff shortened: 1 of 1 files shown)


kev1n-peters:
Still want to test some more and make sure the tradeoff is okay (gas dropoff value doesn't change while moving the slider but waits until after the debounced value fires off an effect to fetch the conversion from the relayer contract).

Also note that fetching the native gas amount from the relayer contract was already happening - this PR just removes the intermediate calculation that uses the conversion rate from CoinGecko.

PR #1236, opened 2023-11-16
connect/gateway: fix message/vaa decoding when there are multiple ibc transfers

### LINKED ISSUE #1231: [QA] [Mainnet] Cannot claim WFTM, tx Osmosis -> Ethereum
Checked on https://deploy-preview-1174--wormhole-connect-mainnet.netlify.app/ [Mainnet] with macOS 12.7 and Google Chrome Version 119.0.6045.123 (Official Build) (x86_64).

Steps:
1. Prepare to send 0.001 WFTM from Osmosis (Leap) to Ethereum (Metamask)
2. Click "Approve and proceed with transaction" button
3. Click "Claim" button

Send from tx: [Mintscan](https://www.mintscan.io/osmosis/tx/A8120D4207C99648E69325C96E6C150BBE30397528CB2472A03869283D5A98CF)

Error:
```
Error: call revert exception; VM Exception while processing transaction: reverted with reason string "invalid EVM address" [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="completeTransfer(bytes)", data="0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000013696e76616c69642045564d206164647265737300000000000000000000000000", errorArgs=["invalid EVM address"], errorName="Error", errorSignature="Error(string)", reason="invalid EVM address", code=CALL_EXCEPTION, version=abi/5.7.0)
    at f.makeError (main.119ba056.js:2:948592)
    at f.throwError (main.119ba056.js:2:948712)
    at A.decodeFunctionResult (main.119ba056.js:2:810718)
    at Object.<anonymous> (main.119ba056.js:2:846706)
    at Generator.next (<anonymous>)
    at a (main.119ba056.js:2:842193) 
```

image

Transaction is successfully completed.

M-Picco:
It turns out that sometimes IBC transfers can get queued up on the source chain and, when the relayer eventually picks them up, they all get delivered on a single transaction on wormchain. So for `n` transactions on the source chain, there will be only one transaction on wormchain, which emits the `n` observations, each as a msg in the transaction.

The current implementation assumed there was only one transaction per IBC transfer, so it would extract the first vaa present in the logs, yielding an incorrect VAA. The changes in this PR involve using the IBC transfer information (src and dst channel, packet sequence) to find the exact entry in the wormchain transaction's logs for that specific IBC transfer.

Should address #1231

### DIFF
--- a/sdk/src/contexts/cosmos/context.ts
+++ b/sdk/src/contexts/cosmos/context.ts
@@ -9,17 +9,21 @@ import {
   parseVaa,
 } from '@certusone/wormhole-sdk';
 import { CosmWasmClient } from '@cosmjs/cosmwasm-stargate';
-import { EncodeObject, decodeTxRaw } from '@cosmjs/proto-signing';
+import { EncodeObject } from '@cosmjs/proto-signing';
 import {
   BankExtension,
   Coin,
   IbcExtension,
   QueryClient,
   calculateFee,
-  logs as cosmosLogs,
   setupBankExtension,
   setupIbcExtension,
 } from '@cosmjs/stargate';
+import {
+  Tendermint34Client,
+  Tendermint37Client,
+  TendermintClient,
+} from '@cosmjs/tendermint-rpc';
 import { MsgExecuteContract } from 'cosmjs-types/cosmwasm/wasm/v1/tx';
 import { BigNumber, utils } from 'ethers';
 import {
@@ -27,7 +31,6 @@ import {
   base58,
   base64,
   hexStripZeros,
-  hexlify,
   keccak256,
   zeroPad,
 } from 'ethers/lib/utils';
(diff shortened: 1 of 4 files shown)


sdk/src/contexts/cosmos/context.ts
```diff
@@ -473,64 +471,10 @@ export class CosmosContext<
     id: string,
     chain: ChainName | ChainId,
   ): Promise<ParsedMessage | ParsedRelayerMessage> {
-    const client = await this.getCosmWasmClient(chain);
```
kev1n-peters, line 476: Is throwing here going to cause an issue in the resume transaction flow for cosmos TXs? I see that context.getMessage gets called in `getRouteFromTx` in `operator.ts`

M-Picco: `getRouteFromTx` checks whether the source chain is a gateway chain before trying to retrieve the message (see [here](https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/routes/operator.ts#L65)), so this method should never be called. Maybe I can mention that on the error message.

kev1n-peters (approved):

PR #1251, opened 2023-11-21
#1221 - adjust size for bottom menu

sebastianscatularo:
Adjust bottom nav bar size

image

### DIFF
--- a/wormhole-connect/src/components/FooterNavBar.tsx
+++ b/wormhole-connect/src/components/FooterNavBar.tsx
@@ -16,11 +16,12 @@ const useStyles = makeStyles()((theme) => ({
     flexDirection: 'row',
     gap: '8px',
     padding: '8px',
-    width: '85%',
+    width: '97%',
   },
   menuItem: {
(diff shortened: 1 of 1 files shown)


kev1n-peters:
Can we change this label: `      "label": "Advance Tools",` to `Advanced Tools`?

kev1n-peters (dismissed):

wormhole-connect/src/components/FooterNavBar.tsx
```diff
@@ -16,11 +16,12 @@ const useStyles = makeStyles()((theme) => ({
     flexDirection: 'row',
     gap: '8px',
     padding: '8px',
-    width: '85%',
+    width: '97%',
```
evan-gray: wondering why this uses width instead of being width 100% with a horizontal margin or something.

sebastianscatularo: you are right, I set it to 100% with margin auto

M-Picco (approved):

Issue #1233, opened 2023-11-15
[connect > portal] - add warning message for all the transfer of USDC that does not use CCTP


sebastianscatularo:
Add a warning message for users that attempt to transfer USDC using a different corridor than CCTP

- [ ] add a warning message for users that attempt to transfer USDC using a different corridor than CCTP
- [ ] add an option to enable this warning
- [ ] (optional) add an option to redirect the user to another site, for instance [USDC Bridge](https://www.portalbridge.com/usdc-bridge/)

mliu-xlabs:
Additional requirements:
- User should still be able to natively use CCTP in supported chain pairs in the main flow

See warning below:

![Image](https://github.com/wormhole-foundation/wormhole-connect/assets/130170527/7c666803-e27d-494e-a318-385d7552e360)

- The warning should appear in the same location as the warning shown above.
- There are 2 other checks that will happen for potential warning messages that will appear here.
- List of priority of the checks:

1. Check whether the asset is registered on the dest chain
2. Check whether the user has an associated token account created for that asset
3. Check whether the chain pair supports CCTP (USDC only)


warning message as follows:

This transaction will transfer wrapped USDC (wUSDC) to the destination chain. If you want to transfer native USDC on chains supported by Circle's CCTP, use the [USDC Bridge](https://www.portalbridge.com/usdc-bridge/).

PR #1258, opened 2023-11-22
[connect > portal] - add warning message for all the transfer of USDC that does not use CCTP

yuli-ferna:
The condition follows these priority of the checks:

- Check whether the asset is registered on the dest chain
- Check whether the user has an associated token account created for that asset
- Check whether the chain pair supports CCTP (USDC only)

Result if the feature it's on:
image

Issue: https://github.com/wormhole-foundation/wormhole-connect/issues/1233


wormhole-connect/src/config/index.ts
```diff
@@ -49,6 +49,9 @@ export const ATTEST_URL =
     ? ''
     : 'https://wormhole-foundation.github.io/example-token-bridge-ui/#/register';
 
+export const PORTAL_USDC_BRIDGE_URL =
+  'https://www.portalbridge.com/usdc-bridge/';
```
sebastianscatularo: could we make it configurable? and if not, could we remove the `www.`

sebastianscatularo (changes requested):

wormhole-connect/src/views/Bridge/Inputs/TokenWarnings.tsx
```diff
@@ -205,10 +215,24 @@ function TokenWarnings() {
   ]);
 
   useEffect(() => {
+    // check if the token is USDC and the chains involved are not CCTP
+    const usdcAndNoCCTP =
```
sebastianscatularo, line 221: also, we may want this to be configurable to allow integrators to enable/disable at their convenience, not sure if all wants to get redirect to portal bridge

yuli-ferna: Done!!

wormhole-connect/src/config/index.ts
yuli-ferna: Done!!

wormhole-connect/src/views/Bridge/Inputs/TokenWarnings.tsx
```diff
@@ -7,7 +7,7 @@ import {
   setAssociatedTokenAddress,
   setForeignAsset,
 } from 'store/transferInput';
-import { ATTEST_URL, TOKENS } from 'config';
+import { ATTEST_URL, PORTAL_USDC_BRIDGE_URL, TOKENS } from 'config';
```
sebastianscatularo: maybe not PORTAL, just bridge :)

sebastianscatularo (dismissed):

wormhole-connect/src/views/Bridge/Inputs/TokenWarnings.tsx
yuli-ferna: done!

sebastianscatularo (dismissed):

PR #1277, opened 2023-11-28
[Bugfix] Only show supported wallets based on the networks config

yuli-ferna:
Issue: https://github.com/XLabs/portal-bridge-ui/issues/535

Result: 
Now we show the supported wallets based on the networks config

https://github.com/wormhole-foundation/wormhole-connect/assets/35125931/1d925521-23a7-4347-a183-e683f7b050b9


wormhole-connect/src/views/WalletModal.tsx
```diff
@@ -145,6 +145,14 @@ const getWalletOptions = async (
       [Context.SEI]: seiOptions,
       [Context.COSMOS]: wallets.cosmos,
     };
+    // filter allWallets that are not supported by network list config
+    const networkContext = CHAINS_ARR.map((chain) => chain.context);
+    const set = new Set(networkContext);
+    Object.keys(allWallets).forEach((context) => {
+      if (!set.has(context as Context)) {
+        delete allWallets[context as Context];
```
sebastianscatularo: Could we avoid any future side effects by deleting and, instead of that, create `allWallets` from the beginning with the set of wallets that we plan to support based on the configs?

sebastianscatularo (changes requested):

wormhole-connect/src/views/WalletModal.tsx
yuli-ferna: Ok I update the solution without the "delete" tell me what you think

sebastianscatularo (dismissed):

M-Picco (approved):

Issue #1289, opened 2023-12-01
[QA] [Testnet] Transaction from Solana fails


AigaD:
Preconditions:
Solflare wallet connected at sending From section
Metamask wallet connected at sending To section

Environment: MacBook Pro, macOS Ventura 13.0.1, Chrome Version 119.0.6045.159

Tested from: https://testnet.portalbridge.com/df5df83982/# / https://wormhole-connect.netlify.app/#

Steps:
1. Prepare to send 0.1 SOL from Solana to BSC
2. Click Approve and proceed with transaction
3. Observe

Actual result: Transaction fails, "Error with transfer, please try again", error in console: `WalletSendTransactionError: failed to send transaction: Transaction simulation failed: Error processing Instruction 4: custom program error: 0x7d0
    at Y8.solana.solflare.sendTransaction`

Expected result: Transaction is taken to next screen and is completed

Attachment:

https://github.com/wormhole-foundation/wormhole-connect/assets/91047061/2b5aeb04-103b-487f-9fdc-fcdd782f60b7

hsnice16:
Facing this issue with the Create account as well. 
`WalletSendTransactionError: failed to send transaction: Transaction simulation failed: Blockhash not found`

<img width="1389" alt="Screenshot 2023-12-06 at 7 42 30 PM" src="https://github.com/wormhole-foundation/wormhole-connect/assets/56081584/133efab6-b98d-42cb-97d4-32f74cd5288c">

AigaD:
Linking issue with creating associated account to this ticket: https://github.com/wormhole-foundation/wormhole-connect/issues/770

hsnice16:
Relevant Issue in `solana-labs` - https://github.com/solana-labs/wallet-adapter/issues/858

hsnice16:
@aadam-10 I would like to raise a PR for this issue. The issue is happening because of the incorrect Solana testnet RPC here - https://github.com/wormhole-foundation/wormhole-connect/blob/development/sdk/src/config/TESTNET.ts#L306

AigaD:
Tested OK, closing the ticket.
Tested from: https://wormhole-connect.netlify.app/#
Transaction from [Solana](https://explorer.solana.com/tx/1XTjNHawGdrBJMvZNqcfSxaMmPrdEVSBfJDCibqGjuUZU9MtCqqMKDCW8znweNLYG6zGsYHUmS3ajntYRQZRxxp?cluster=devnet) network was completed.
[...]

PR #1310, opened 2023-12-07
fix: solana testnet rpc endpoint

hsnice16:
This PR fixes issue #1289 
The issue was happening because of the incorrect Solana testnet RPC endpoint.


kev1n-peters:
Thank you for opening the PR and looking into the issue. However https://api.devnet.solana.com is the correct RPC URL that should be used when Connect is configured in testnet mode (even though it's called Solana "devnet"). The issue with the Solflare wallet is unrelated to this change.

hsnice16:
Okay. Any reason to use the devnet RPC endpoint for testnet environment?

kev1n-peters:
Yes, Solana devnet is the correct network to use when Wormhole Connect is configured in testnet mode. I agree that this is confusing and one would think Solana testnet should be used instead, but that is not the case as Wormhole uses Solana devnet as the environment for testing purposes (not testnet).

hsnice16:
Okay. Thanks for explaining.

PR #1319, opened 2023-12-11
Added ethBridge and wstETHBridge routes

### LINKED ISSUE #1390: Uniswap ETH Bridge support: ETH/wstETH route

kev1n-peters:
These routes utilize the Portico contracts to perform a swap between the native token and the canonical token (AKA the "highway token"), effectively allowing users to send and receive native ETH, WETH, and wstETH between the supported chains.


wormhole-connect/src/views/Bridge/RouteOptions.tsx
```diff
@@ -140,29 +145,59 @@ function Tag(props: TagProps) {
...
     async function load() {
-      const receiveAmt = await RouteOperator.computeReceiveAmount(
-        props.route.route,
-        Number.parseFloat(amount),
-        { toNativeToken, relayerFee },
-      );
-      setReceiveAmt(Number.parseFloat(toFixedDecimals(`${receiveAmt}`, 6)));
+      try {
+        const routeOptions = isPorticoRoute(props.route.route)
+          ? portico
```
artursapek, line 192: If we end up adding multiple other routes like this one (hashflow, and the other LL that was supposed to follow this) - won't this code end up getting pretty cluttered/complicated? are you worried about that? I might be way off here

kev1n-peters: I think this is a valid concern and we should try to figure something out to make this more maintainable. What we're doing here is selecting specific variables from the redux store based on the current route and creating the corresponding `routeOptions` object (each route has its own set of options). As more routes get added, we will have more cases to consider and the if statement will become more complex. I am not really sure what a better solution is yet or if it makes sense for a refactor to happen in this PR (I would prefer to reduce the surface area of potential regressions if possible). Maybe more complex routes could have their own `RouteOption` component, this way we know which variables to select from the store ahead of time. But this is just moving the conditional to wherever creates the route component

artursapek: cool, I agree it makes sense to deal with later to keep this change smaller.

sdk/src/contexts/eth/context.ts
```diff
@@ -769,6 +769,9 @@ export class EthContext<
         block: receipt.blockNumber,
         gasFee,
         payload: transferWithPayload.payload,
+        fromAddress: utils.hexlify(
```
M-Picco, line 772: Is it possible this field is missing from the return a bit further down below (line 786-810)

wormhole-connect/src/hooks/usePorticoRelayerFee.ts
```diff
@@ -0,0 +1,90 @@
...
+        '0',
+        fromChain,
+        toChain,
+      );
+      if (!isSupported) {
+        dispatch(resetRelayerFee());
+        return;
+      }
+      dispatch(setFetchingRelayerFee());
+      while (!cancelled) {
```
M-Picco, line 57: Should this forever even after successfully fetching the relayer fee?

wormhole-connect/src/store/helpers.ts
```diff
@@ -0,0 +1,42 @@
+export type DataWrapper<T> = {
+  data: T | null;
+  error: any | null;
+  isFetching: boolean;
```
M-Picco: Are either `isFetching` or `receivedAt` being used anywhere in the code?

wormhole-connect/src/store/porticoBridge.ts
M-Picco, line 1: For the future: maybe this slice could be generalized for anything that requires swaps?

M-Picco (dismissed):

wormhole-connect/src/hooks/usePorticoRelayerFee.ts
kev1n-peters: Yeah, the relayer fee could change so we fetch it every minute

wormhole-connect/src/store/helpers.ts
kev1n-peters: I don't believe so, will look at removing them

wormhole-connect/src/store/porticoBridge.ts
kev1n-peters: Possibly, we may have to revisit when we add another swap route

sdk/src/contexts/eth/context.ts
kev1n-peters: Yeah, I guess we should add it there too for completeness

M-Picco (approved):

PR #1320, opened 2023-12-11
[#1280] Conditionally render gas row if non-zero

### LINKED ISSUE #1280: [feat][enhancement] - hide gas drop off row for non gas drop off
As a user, I do not want to see gas drop-off details 
when the gas drop-off is zero.

image

artursapek:
Hides unnecessary zero value

Should close #1280


M-Picco (approved):
Tested on the preview site. LGTM!

PR #1341, opened 2023-12-14
Updated guardian RPCs and added round-robin support for fetching VAAs.

kev1n-peters:
Previously we would just fetch the VAA from the first guardian in the list. I didn't include wormholescan in the updated list of RPCs since there's a separate function for making calls to wormholescan: https://github.com/wormhole-foundation/wormhole-connect/blob/444c43176c3ae886b2febca4695ada0f8c8aedf8/wormhole-connect/src/utils/vaa.ts#L94

### DIFF
--- a/wormhole-connect/src/config/index.ts
+++ b/wormhole-connect/src/config/index.ts
@@ -69,15 +69,16 @@ export const USDC_BRIDGE_URL = config.cctpWarning?.href || '';
 export const WORMHOLE_RPC_HOSTS =
   ENV === 'MAINNET'
     ? [
-        'https://wormhole-v2-mainnet-api.certus.one',
-        'https://wormhole.inotel.ro',
         'https://wormhole-v2-mainnet-api.mcf.rocks',
         'https://wormhole-v2-mainnet-api.chainlayer.network',
         'https://wormhole-v2-mainnet-api.staking.fund',
-        'https://wormhole-v2-mainnet.01node.com',
       ]
     : ENV === 'TESTNET'
-    ? ['https://wormhole-v2-testnet-api.certus.one']
+    ? [
+        'https://guardian.testnet.xlabs.xyz',
+        'https://guardian-01.testnet.xlabs.xyz',
+        'https://guardian-02.testnet.xlabs.xyz',
+      ]
     : ['http://localhost:7071'];
 
 export const NETWORK_DATA =
--- a/wormhole-connect/src/utils/vaa.ts
+++ b/wormhole-connect/src/utils/vaa.ts
@@ -4,7 +4,7 @@ import { utils, providers, BigNumberish } from 'ethers';
 import axios from 'axios';
 import { ChainId, ChainName } from '@wormhole-foundation/wormhole-connect-sdk';
 
-import { CHAINS, ENV, WORMHOLE_API } from 'config';
+import { CHAINS, WORMHOLE_API, WORMHOLE_RPC_HOSTS } from 'config';
 import {
   ParsedMessage,
   ParsedRelayerMessage,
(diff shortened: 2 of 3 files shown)


wormhole-connect/src/utils/vaa.ts
```diff
@@ -184,12 +170,25 @@ export async function fetchVaaGuardian(
...
 
-  const { vaaBytes: vaa } = await getSignedVAA(
-    WORMHOLE_RPC_HOSTS[0],
-    emitterChain,
-    emitterAddress,
-    sequence,
-  );
+  // round-robin through the RPC hosts
+  let vaa: Uint8Array | undefined;
+  for (const host of WORMHOLE_RPC_HOSTS) {
```
artursapek, line 175: my understanding of round robin is it's stateful across requests: if you have hosts `[A, B, C]`, the first request will go to `A`, then the second will go to `B`. this code will always try `A` first and only use `B` if `A` failed.

kev1n-peters: Maybe we could simply add a module scoped variable that keeps track of the last index where the VAA was fetched successfully and start at that index on subsequent calls

M-Picco: I understand this call is a fallback for when the explorer api doesn't have the VAA yet. I'm not sure the guardians will receive much pressure from here so it doesn't matter if it always try for A first, specially since the poll interval is of around 30s. Still, implementing the round robin should be straightforward.

artursapek: ok. so this is more for failover rather than load balancing?

kev1n-peters: Yeah, we try to fetch the VAA from the wormholescan API first and if that doesn't have it, then we hit the guardians. If none of them have it then we wait 30 seconds and repeat this process.

artursapek:
Given that you aren't load balancing the requests, it seems the mcf.rocks Guardian RPC (https://wormhole-v2-mainnet-api.mcf.rocks) will suddenly start getting most of Connect's traffic. Are they aware of that, and good with it?

tonyjin:
> Given that you aren't load balancing the requests, it seems the mcf.rocks Guardian RPC (https://wormhole-v2-mainnet-api.mcf.rocks) will suddenly start getting most of Connect's traffic. Are they aware of that, and good with it?

The title of the PR suggests there's some round robin logic added so multiple RPC providers are used - is that not the case, or are you referring to the first request usually going to MCF?

Good shoutout though I can reach out to them.

Edit: ah I see the logic. @kev1n-peters it'd be best if there was some randomization of how that round trip logic works instead of rotating through in one order every time.

wormhole-connect/src/config/index.ts
```diff
@@ -69,15 +69,16 @@ export const USDC_BRIDGE_URL = config.cctpWarning?.href || '';
 export const WORMHOLE_RPC_HOSTS =
   ENV === 'MAINNET'
     ? [
-        'https://wormhole-v2-mainnet-api.certus.one',
-        'https://wormhole.inotel.ro',
         'https://wormhole-v2-mainnet-api.mcf.rocks',
         'https://wormhole-v2-mainnet-api.chainlayer.network',
         'https://wormhole-v2-mainnet-api.staking.fund',
-        'https://wormhole-v2-mainnet.01node.com',
```
tonyjin, line 77: Can this use:
```
const MAINNET_GUARDIAN_RPC: string[] =  [
    "https://api.wormholescan.io", // Explorer offers a guardian equivalent endpoint for fetching VAAs
    "https://wormhole-v2-mainnet-api.mcf.rocks",
    "https://wormhole-v2-mainnet-api.chainlayer.network",
    "https://wormhole-v2-mainnet-api.staking.fund",
]
```

(from our docs https://docs.wormhole.com/wormhole/reference/sdk-docs)

Relatedly, do we have to hit any Guardian RPC directly at all? On one hand depending on wormholescan is a bit centralized, but on the other, we already depend on wormholescan heavily for other status checks.

kev1n-peters: We try to fetch the VAA from the Wormholescan API first and if that doesn't have the VAA, then we hit the guardians. If none of them have it then we wait 30 seconds and repeat this process.

M-Picco: We do use the explorer api to fetch the VAA (see [here](https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/utils/vaa.ts#L105)). The guardians are queried only if the vaa was not found on the api.

M-Picco (approved):

artursapek:
lgtm (cant find the approve button)

PR #1347, opened 2023-12-14
add integrity check for js/css assets

sebastianscatularo:
simplest PoC to add integrity checks to Wormhole Connect Loader to avoid poisoned asset to be loaded

check: 
- https://www.w3schools.com/tags/att_script_integrity.asp
- https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#tools_for_generating_sri_hashes


artursapek:
I think this shouldn't be too hard to finish. we just need CI changes now right?

PR #1353, opened 2023-12-15
Load wallet code dynamically in separate chunks

### LINKED ISSUE #1343: [Perf] Wallet code should load dynamically using import()
## Proposed change

The Connect widget's "Choose Wallet" button will be disabled until the user selects a Network first

## Why?

- For the end user, seeing every one of the 33 supported wallets all at once isn't really that useful to begin with
- It will make it much easier to achieve a small bundle size via dynamic loading

## Dynamic loading

If we made this change, we'd be able to load wallet code only for the selected network, after the user selects it. This achieves two things:

1. Reduces initial bundle size by not including wallet code at all (which appears to be roughly 50% of total bundle size)
2. Reduces total code loaded by never loading wallet code for networks that the user doesn't select

Opening issue for discussion!

artursapek:
## Changes

- Disable the Connect Wallet button until the user has selected a chain/network
- Refactor `utils/wallet` so that all chain-specific wallet code lives in separate files, loaded dynamically on demand using `import()`.

Here's a build before these changes:

image

And after:

image

Our main entry point bundle has gone from 4.51MB to 2.4MB gzipped. There are now several other chunks, each corresponding to a different chain type. These will only be loaded when users actually select that chain and go to connect a wallet.

See it in action here. Upon first loading the page, we load the 15.4 MB (uncompressed) main bundle. Then, if we select Sei chain in the UI and click on Connect Wallet, the page loads the second, 3.5MB uncompressed Sei bundle!

https://github.com/wormhole-foundation/wormhole-connect/assets/897596/879346d9-0372-4df4-8f04-d2348667694f

## Further improvements

- The next low hanging fruit I see is `routes/operator`. Each route that `getRoute` returns could be loaded dynamically. I think this would make worthwhile reductions in main bundle size, because `cosmjs` continues to take up a lot of room in there. I imagine we can get under 2MB after doing that.

## Other thoughts

Besides bundle size/bandwidth savings, this has the added benefit of limiting our attack surface area in vendor code. If one of these wallet libraries gets compromised, we won't be automatically loading and running it for every Connect user.

## TODO

- [x] Add tooltip to disabled connect wallet button

## Follow-up:

https://github.com/wormhole-foundation/wormhole-connect/issues/1616


wormhole-connect/src/utils/wallet/index.ts
```diff
@@ -7,14 +7,15 @@ import {
...
-import { CHAIN_ID_SEI, postVaaSolanaWithRetry } from '@certusone/wormhole-sdk';
+import { postVaaSolanaWithRetry } from '@certusone/wormhole-sdk';
 import { ContractReceipt } from 'ethers';
 import { NotSupported, Wallet } from '@xlabs-libs/wallet-aggregator-core';
 import {
   EVMWallet,
   MetamaskWallet,
   WalletConnectWallet,
 } from '@xlabs-libs/wallet-aggregator-evm';
+import { SeiTransaction } from '@xlabs-libs/wallet-aggregator-sei';
```
artursapek, line 18: I still have to import this type from `wallet-aggregator-sei`, but it properly tree shakes and avoids bloating the main bundle size.

artursapek:
pushed error handling logic, which will tell the user to refresh if a chunk fails to load.

when Connect is properly deployed, this shouldn't happen. but shit happens. or maybe their wifi died.

https://github.com/wormhole-foundation/wormhole-connect/assets/897596/bf834d2a-8da4-4b50-b811-764cf597fb1e

wormhole-connect/src/utils/wallet.ts
```diff
@@ -1,409 +0,0 @@
-import {
```
artursapek, line 1: this file was moved to `utils/wallet/index.ts`, and then modified there.

kev1n-peters:
Tested an automatic transfer between Solana and Sui and saw the corresponding wallet code chunks loaded when selecting a wallet. The wallet code splitting is pretty cool.

kev1n-peters (dismissed):

kev1n-peters (approved):

PR #1360, opened 2023-12-18
[Enhancemenet] Show automatic redemption route even if insufficient funds are available

### LINKED ISSUE #1299: [Enhancemenet] Show automatic redemption route even if insufficient funds are available
- Show routes for both 1tx automatic redeems and 2tx manual redeems, with the 1tx automatic redeem grayed out with a tooltip or warning if the transferred amount is below the minimum amount. E.g. "Not enough funds on the source chain for automatic redemption" or "X more <token> is needed on the source chain for automatic redemption"
- Cosmos routes: do not show 2tx options as currently
- If a CCTP route is available between the source and the destination, do not show the wormhole token bridge route to avoid liquidity fragmentation.

yuli-ferna:
Issue: https://github.com/wormhole-foundation/wormhole-connect/issues/1299

Now we take into account a new field, that the route is supported, in that context it means that the amount is sufficient to cover the fee, even though the route is enabled.

Result:
image

[testt2.webm](https://github.com/wormhole-foundation/wormhole-connect/assets/35125931/b1ef64c3-bf47-4250-91fa-250344849c94)

### DIFF
--- a/wormhole-connect/src/routes/cctpRelay/cctpRelay.ts
+++ b/wormhole-connect/src/routes/cctpRelay/cctpRelay.ts
@@ -188,6 +188,16 @@ export class CCTPRelayRoute extends CCTPManualRoute implements RelayAbstract {
       CHAINS[destChainName]?.contracts.cctpContracts?.wormholeCircleRelayer;
     if (!bothHaveRelayer) return false;
 
+    return true;
+  }
+
+  async isRouteSupported(
+    sourceToken: string,
+    destToken: string,
+    amount: string,
+    sourceChain: ChainName | ChainId,
+    destChain: ChainName | ChainId,
+  ): Promise<boolean> {
     let relayerFee;
     try {
       relayerFee = await this.getRelayerFee(
--- a/wormhole-connect/src/routes/cctpRelay/cctpRelay.ts
+++ b/wormhole-connect/src/routes/cctpRelay/cctpRelay.ts
@@ -150,7 +150,7 @@ export class CCTPRelayRoute extends CCTPManualRoute implements RelayAbstract {
     });
   }
 
-  async isRouteAvailable(
+  async isRouteSupported(
     sourceToken: string,
     destToken: string,
     amount: string,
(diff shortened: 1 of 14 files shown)


wormhole-connect/src/routes/relay/relay.ts
```diff
@@ -83,7 +83,18 @@ export class RelayRoute extends BridgeRoute implements RelayAbstract {
     const accepted = await isAcceptedToken(tokenId);
 
     if (!accepted) return false;
+    return true;
```
M-Picco: You can remove the `accepted` variable and just do `return isAcceptedToken(tokenId);`

wormhole-connect/src/routes/cctpRelay/cctpRelay.ts
```diff
@@ -188,6 +188,16 @@ export class CCTPRelayRoute extends CCTPManualRoute implements RelayAbstract {
       CHAINS[destChainName]?.contracts.cctpContracts?.wormholeCircleRelayer;
     if (!bothHaveRelayer) return false;
 
+    return true;
```
M-Picco: You can just return `bothHaveRelayer`

wormhole-connect/src/utils/transferValidation.ts
```diff
@@ -269,7 +271,7 @@ export const validate = async (
     transferInput.destToken &&
     transferInput.amount &&
     Number.parseFloat(transferInput.amount) >= 0 &&
-    transferInput.availableRoutes !== undefined
+    transferInput.routeStates?.filter((rs) => rs.available) !== undefined
```
M-Picco: If you need to enable the validation only if there's at lease one available route you could use `transferInput.routeStates?.some(rs => rs.available)` and do away with the ternary operator

M-Picco (commented):
A naming suggestion: swap `supported` and `available`. The first communicates that the token can be sent through the route while the later that you can do it at this moment of time. Then, an `unsupported` route is for example a token that is not accepted by the relayers and an `unavailable` route is one you can send the token through but not now, for whatever reason (e.g. not enough funds to cover the fee).

What do you think?

M-Picco (approved):

M-Picco (approved):

sebastianscatularo (approved):

PR #1366, opened 2023-12-19
checkForeignAssetsConfig: skip asset in case of getting a 429 error when fetching decimals

M-Picco:


artursapek:
I feel like this will get forgotten about in like 2 seconds but the change LGTM lol

artursapek (approved):

PR #1384, opened 2023-12-21
connect: refine USDC warning

### LINKED ISSUE #1374: Don't show illiquid USDC banner warning if a token bridge routed transfer wouldn't result in a wrapped asset
Scenario:
1. Have some USDC(ethereum) on a non-cctp chain like Moonbeam
2. Try to bridge it to Ethereum mainnet

Experienced:
- The "This transaction will transfer wrapped USDC [...]" warning banner is shown

Desired:
- The warning isn't shown

evan-gray:
Suggesting a change to the USDC warning based on feedback in #1374. 

Essentially, show a warning if the result is wrapped USDC and the chains involved support CCTP

Rationale:
- transferring wrapped USDC back home (unwrapping) shouldn't be a warning
- CCTP would show as USDC on both ends, but the result would be native (the same check as above)
- if both chains don't support CCTP, it doesn't make sense to suggest using it, (at least not with this warning) as the user doesn't have a clear alternative and using the USDC (CCTP) only bridge wouldn't help

This new condition only shows the warning in a significantly more limited set of use cases, as I feel the warning should suggest to the user an actionable, better thing to do. In that light, I attempted to make the code match the spirit of the warning. Otherwise, it would suggest that users use the USDC (only) Bridge even when they couldn't.

> This transaction will transfer wrapped USDC (wUSDC) to the destination chain. If you want to transfer native USDC on chains supported by Circle's CCTP, use the USDC Bridge

This warning only shows if the `cctpWarning.href` is set in the config, so I tested by setting the html to:
```html
<div
  id="wormhole-connect"
  config='{"cctpWarning":{"href":"https://portalbridge.com/usdc-bridge/"}}'
></div>
```

Tested:
- Eth - USDC (Goerli) -> Avax - USDC (Fuji) = No warning (is using CCTP)
- Eth - USDC (Goerli) -> Avax - USDC (Goerli) = Shows warning (could be using CCTP but isn't, result is wrapped)
- Eth - USDC (Goerli) -> Moonbeam - USDC (Goerli) = No warning (there is no CCTP on Moonbeam)
- Avax - USDC (Goerli) -> Eth - USDC (Goerli) = No warning (this USDC is going home)
- Avax - USDC (Fuji) -> Eth - USDC (Goerli) = No warning (is using CCTP)
- Avax - USDC (Fuji) -> Eth - USDC (Fuji) = Shows warning (could be using CCTP but isn't, result is wrapped)
- Moonbeam - USDC (Goerli) -> Avax - USDC (Goerli) = No warning (there is no CCTP on Moonbeam)
- Moonbeam - USDC (Goerli) -> Goerli - USDC (Goerli) = No warning (there is no CCTP on Moonbeam)
- Moonbeam - USDC (Goerli) -> Fantom - USDC (Goerli) = No warning (neither of these have CCTP)

### DIFF
--- a/wormhole-connect/src/views/Bridge/Inputs/TokenWarnings.tsx
+++ b/wormhole-connect/src/views/Bridge/Inputs/TokenWarnings.tsx
@@ -217,13 +217,26 @@ function TokenWarnings() {
   useEffect(() => {
     // if the url it's empty that means the user doesn't want this feature
     const cctpWarningFlag = !!USDC_BRIDGE_URL;
-    // check if the token is USDC and the chains involved are not CCTP
-    const usdcAndNoCCTP =
-      cctpWarningFlag &&
+    // check if the tokens will be wrapped USDC
+    const isResultWrappedUSDC =
       tokenConfig?.symbol === 'USDC' &&
       destTokenConfig?.symbol === 'USDC' &&
-      (!(toChain && CCTP_CHAINS.includes(toChain)) ||
-        !(fromChain && CCTP_CHAINS.includes(fromChain)));
+      destTokenConfig?.nativeChain !== toChain;
+    // check if the chains support CCTP
+    const bothChainsSupportCCTP =
+      toChain &&
+      CCTP_CHAINS.includes(toChain) &&
+      fromChain &&
+      CCTP_CHAINS.includes(fromChain);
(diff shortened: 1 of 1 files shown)


M-Picco (approved):

artursapek (approved):
lgtm nice one-liner

PR #1393, opened 2023-12-22
Vite environment variables support

### LINKED ISSUE #1429: Move to Vite

kev1n-peters:
We can't access environment variables on the client from `process.env` in a Vite build. Changed the default env var prefix to be `REACT_APP_` to maintain backwards compatibility for now. See more here: https://vitejs.dev/guide/env-and-mode


kev1n-peters:
The loader is still built with CRA/Webpack, so I didn't change this reference: https://github.com/wormhole-foundation/wormhole-connect/blob/ad1acf876da22769ba53e03df462797a02a9f7fc/wormhole-connect-loader/src/index.tsx#L6

artursapek (dismissed):
thanks!

M-Picco (approved):

PR #1410, opened 2024-01-02
Add vite-plugin-environment and load REACT_APP_ prefixed env vars

M-Picco:
Vite only loads env variables prefixed with `VITE_` and exposes them through `import.meta.env` (see [docs](https://vitejs.dev/guide/env-and-mode#env-files)). I made the change but ran into some errors related to `import.meta` and ts targets. Since a simple `target` change didn't fix it I went for the [vite-plugin-environment](https://www.npmjs.com/package/vite-plugin-environment) which allows to load other prefixed variables and exposes them through `process.env`. Now it loads variables as they are, with the `REACT_APP_` prefix.

I'm also open to any other suggestions on how to beter handle this.


M-Picco:
Closed in favor of #1393

PR #1421, opened 2024-01-04
Reorganize dependencies and upgrade typescript to its latest versions

M-Picco:


sdk/src/contexts/solana/utils/wormhole/coder/instruction.ts
```diff
@@ -86,7 +86,10 @@ export class WormholeInstructionCoder implements InstructionCoder {
     encoding: 'hex' | 'base58' = 'hex',
   ): Instruction | null {
     if (typeof ix === 'string') {
-      ix = encoding === 'hex' ? Buffer.from(ix, 'hex') : bs58.decode(ix);
+      ix =
+        encoding === 'hex'
+          ? Buffer.from(ix, 'hex')
+          : Buffer.from(bs58.decode(ix));
```
artursapek: hah was this a bug before, which TypeScript wasn't catching? the bs58 case wasn't returning a buffer?

artursapek (dismissed):

sdk/src/contexts/solana/utils/wormhole/coder/instruction.ts
M-Picco: I think it wasn't needed. I reverted the change and it built just fine 🤷

artursapek (approved):

PR #1430, opened 2024-01-06
Fix glitchy error message, remove collapsing from route options

### LINKED ISSUE #952: [QA]Glitch with warning messages

### LINKED ISSUE #1394: [Enhancement] Update text for bridging UX
The below is too busy:
Screenshot 2023-12-22 at 3 47 32 PM

- Update text for "Manual Bridge" pillbox from `Approve a txn from your destination chain wallet` to `Approve transfer with destination wallet`
- Update disabled text for automatic route from `Not available` to `Transfer amount too low`
- Remove hover tooltip over disabled route
- Update text for manual transfer warning to `This transfer requires two transactions: one on the source chain and one on the destination chain. You will need gas on the destination chain.`

Screenshot 2024-01-04 at 3 19 51 PM

### LINKED ISSUE #1422: Always show both automatic and manual redeem routes with different treatment based on availability
THIS IS A SEPARATE TICKET FOR THE COMMENTS ON Issue 1394

1. We currently don't show anything besides manual redeem option for routes where relayer isn't live. This confuses users who are used to automatic relayer transfers.

2. We auto-hide the manual route, this confuses users who expect manual transfer and its not obvious to the user via the carrot that they can go back and select manual transfer.

Show the disabled automatic route on routes where we dont support it e.g. Eth -> AVAX, right now we only show this on routes where we support it but the amount is too low. Instead of the copy being "Amount too low" the copy should instead show "Automatic redeems coming soon". The goal here is to help users who've used the bridge in the past and simply clicked through to actively be aware of the different flow they are experiencing.

Show both routes even when they are both available. Right now we hide the "manual route" and auto-select the automatic route, we should show both, while still auto selecting automatic. Again, the purpose is to help users who've used the bridge in the past and simply clicked through to manual to be aware of the different flow they are experiencing.

artursapek:
*Fix to #952*

I can't reproduce "Scenario 1", perhaps that was already fixed.

For "Scenario 2" I removed the minimum amount validation. It appears that Connect always disables routes where the minimum is higher than the inputted amount. So there's no case where this error message is needed.

The only reason this bug happened is because the `amount` changes before the `route`, for a split second, because the `setRoutes` runs on a debounced `amount` value. Then `route` catches up when Connect forcibly changes it for the user in `setRoutes`. In that split second in between, we show an incorrect error message.

IMO it would be better to do all validation in the redux stores, and not in component effects. All of the debounces and deferred execution makes this kind of messy. But that's a larger refactor.

*I need someone more familiar with Connect to verify this is correct and my solution is valid.*

*Fix to #1394*

- Minor style and copy changes as per https://github.com/wormhole-foundation/wormhole-connect/issues/1394#issuecomment-1877707804

*Fix to #1422*

- Don't actually let the `Collapse` element collapse for the route options. It's a bit weird to continue using it at all, but it has styles consistent with the following sections. A larger refactor could allow us to get off the `Collapse` element here altogether but I'm trying to fight off scope creep here.

### DIFF
--- a/wormhole-connect/src/components/Options.tsx
+++ b/wormhole-connect/src/components/Options.tsx
@@ -23,6 +23,8 @@ const useStyles = makeStyles()((theme: any) => ({
   },
   optionContent: {
     padding: '16px',
+  },
+  optionContentEnabled: {
     '&:hover': {
       backgroundColor: theme.palette.options.hover,
     },
@@ -45,6 +47,7 @@ const useStyles = makeStyles()((theme: any) => ({
 export type Option = {
   key: any;
   child: any;
+  disabled?: boolean;
 };
 type Props = {
   children: Option[];
@@ -75,7 +78,10 @@ function Options(props: Props) {
             unmountOnExit
           >
             <div
-              className={classes.optionContent}
+              className={joinClass([
+                classes.optionContent,
+                !option.disabled && classes.optionContentEnabled,
+              ])}
               onClick={() => props.onSelect(option.key)}
               style={{
                 cursor: props.children.length > 0 ? 'pointer' : 'default',
--- a/wormhole-connect/src/utils/transferValidation.ts
+++ b/wormhole-connect/src/utils/transferValidation.ts
@@ -97,6 +97,7 @@ export const validateAmount = (
   balance: string | null,
   minAmt: number | undefined,
 ): ValidationErr => {
+  if (amount === '') return '';
   const numAmount = Number.parseFloat(amount);
   if (!numAmount) return 'Enter an amount';
   if (numAmount <= 0) return 'Amount must be greater than 0';
--- a/wormhole-connect/src/utils/transferValidation.ts
+++ b/wormhole-connect/src/utils/transferValidation.ts
@@ -95,17 +95,15 @@ export const validateDestToken = (
 export const validateAmount = (
   amount: string,
   balance: string | null,
-  minAmt: number | undefined,
 ): ValidationErr => {
   if (amount === '') return '';
+
   const numAmount = Number.parseFloat(amount);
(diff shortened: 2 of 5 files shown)


artursapek:
@aakashthumaty Do we still want to kill this tooltip? It wasn't clear in #1394. It does seem redundant now, and it's kinda ugly.

image

M-Picco (approved):
LGTM

aakashthumaty:
yes lets kill the tooltip!

PR #1503, opened 2024-01-18
Remove wrapped options for "Frankenstein" ETH

### LINKED ISSUE #1487: Remove Wrapped Options For Frankenstein ETH
Screenshot 2024-01-12 at 2 22 21 PM
Having the wrapped ETH always show up as a destination asset option has become especially confusing in a world with ETH bridge.

We should depreate wrapped ETH from non Ethereum entirely. We should remove non primary eth options in the “destination asset” view so if you’re bridging a non Ethereum ETH eg Arbitrum ETH to eg Sui normally it would show you a “wormhole wrapped Arbitrum ETH” option to land on Sui with this Frankenstein asset - this should no longer be presented to the user. 

The logic would be a simple “turn off token bridge route/destination asset if original asset == ETH && ETH.origin != Ethereum”

Some sample situations:

ETH(Ethereum) on Ethereum -> Solana: ALLOWED and the destination asset list displays wormhole wrapped ETH as an option 
ETH(Optimism) on Optimism -> Solana: should be disallowed
ETH(Avalanche) on Avalanche -> Solana: should be disallowed
ETH(Ethereum) on Sui -> Solana: ALLOWED and the destination asset list displays wormhole wrapped ETH as an option 
ETH(Ethereum) on Solana -> Optimism: ALLOWED and the destination asset list displays wormhole wrapped ETH as an option 
ETH(Ethereum) on Solana -> Ethereum: ALLOWED and the destination asset list displays wormhole wrapped ETH as an option

kev1n-peters:
Addresses #1487

### DIFF
--- a/wormhole-connect/src/routes/bridge/baseRoute.ts
+++ b/wormhole-connect/src/routes/bridge/baseRoute.ts
@@ -25,9 +25,10 @@ import { hexlify } from 'ethers/lib/utils.js';
 
 export abstract class BaseRoute extends RouteAbstract {
   async isSupportedSourceToken(
-    token: TokenConfig | undefined,
-    destToken: TokenConfig | undefined,
+    token?: TokenConfig,
+    destToken?: TokenConfig,
     sourceChain?: ChainName | ChainId,
+    destChain?: ChainName | ChainId,
   ): Promise<boolean> {
     if (!token) return false;
     // if (destToken) {
@@ -44,8 +45,8 @@ export abstract class BaseRoute extends RouteAbstract {
   }
 
   async isSupportedDestToken(
-    token: TokenConfig | undefined,
-    sourceToken: TokenConfig | undefined,
+    token?: TokenConfig,
+    sourceToken?: TokenConfig,
     sourceChain?: ChainName | ChainId,
     destChain?: ChainName | ChainId,
   ): Promise<boolean> {
--- a/wormhole-connect/src/routes/bridge/baseRoute.ts
+++ b/wormhole-connect/src/routes/bridge/baseRoute.ts
@@ -18,7 +18,7 @@ import {
   isSignedWormholeMessage,
   TransferDestInfo,
 } from 'routes/types';
-import { formatGasFee } from 'routes/utils';
+import { formatGasFee, isIlliquidDestToken } from 'routes/utils';
 import { toDecimals } from 'utils/balance';
 import { NO_INPUT } from 'utils/style';
 import { hexlify } from 'ethers/lib/utils.js';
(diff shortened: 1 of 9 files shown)


wormhole-connect/src/routes/operator.ts
```diff
@@ -238,7 +239,8 @@ export class Operator {
             sourceChain,
             destChain,
           );
-          if (isSupported) {
+          const isIlliquid = this.isIlliquidDestToken(token, destChain);
```
artursapek: IMO it's not necessary to add this here as a special case. instead I would put this check inside the `isSupported` method for the route(s) we're concerned with (token bridge & ETH bridge?). this kind of specific logic doesn't belong in the top level routes operator.

`isSupported` already checks that the tokens are supported - this could just be another level on that check.

in the [new Route API](https://github.com/wormhole-foundation/connect-sdk/pull/216/files#diff-49618d9f838bc1dfa71b211e57e3556cab6da7a720637c898293fdcf2e7bef19R22) that @barnjamin and I are working on, we expect all logic to be contained within `isSupported` (and `isAvailable` for routes using a relayer)

artursapek (changes requested):

wormhole-connect/src/routes/operator.ts
kev1n-peters: Changed. It applies to the Bridge, Relay, and CosmosGateway routes

artursapek (dismissed):
nice, looks much cleaner now.

I don't understand why `sourceChain` and `destChain` have `?` as possibly undefined, aren't those required for any transfer? could we make those mandatory?

I also notice we use both `?` as well as ` | undefined` in the same function signature:

```
  async isSupportedDestToken(
    token: TokenConfig | undefined,
    sourceToken: TokenConfig | undefined,
    sourceChain?: ChainName | ChainId,
    destChain?: ChainName | ChainId,
  ): Promise<boolean> {
  ```

at any rate, the change looks OK to me.

**Comment by kev1n-peters** [comment]
> nice, looks much cleaner now.
> 
> I don't understand why `sourceChain` and `destChain` have `?` as possibly undefined, aren't those required for any transfer? could we make those mandatory?
> 
> I also notice we use both `?` as well as ` | undefined` in the same function signature:
> 
> ```
>   async isSupportedDestToken(
>     token: TokenConfig | undefined,
>     sourceToken: TokenConfig | undefined,
>     sourceChain?: ChainName | ChainId,
>     destChain?: ChainName | ChainId,
>   ): Promise<boolean> {
> ```
> 
> at any rate, the change looks OK to me.

I suspect that those parameters were originally made optional in `RouteAbstract` so the method can be called before the user selects a source or dest chain. I'm not sure if it still makes sense for them to be optional without digging deeper into the code. The inconsistency with `| undefined` and `?` is not ideal and should probably be refactored which I'll look into doing in this PR.

File: wormhole-connect/src/routes/bridge/baseRoute.ts [hunk 2]
```diff
@@ -25,9 +25,10 @@ import { hexlify } from 'ethers/lib/utils.js';
...
 export abstract class BaseRoute extends RouteAbstract {
   async isSupportedSourceToken(
-    token: TokenConfig | undefined,
-    destToken: TokenConfig | undefined,
+    token?: TokenConfig,
+    destToken?: TokenConfig,
     sourceChain?: ChainName | ChainId,
+    destChain?: ChainName | ChainId,
   ): Promise<boolean> {
     if (!token) return false;
```
kev1n-peters, line 33: This function probably shouldn't even be called if the `token` is `undefined` and token should just be `TokenConfig`, but I don't want to stray too far off into a refactoring adventure...

artursapek (approved):
ship it

PR #1512, opened 2024-01-19
Add time estimate to route selection cells

### LINKED ISSUE #1497: Add time estimate to route selection cells
Right now we dont display an expected time on any route, all currently available routes are expected to take as long as network finality given they are all reliant on token bridge. 

We should show this network finality estimate under the fee on right hand side in the route selection cells. 

Under "after fees" we should show an estimate with the copy "x minutes/seconds" based on the destination chain. 

Screenshot 2024-01-18 at 12 09 59 AM

Disclaimer: I know this this isn't a scalable solution given the hard-coded values we should think about the interface for taking in a "completion time" value for each new "route" integrated.

yuli-ferna:
Issue: https://github.com/wormhole-foundation/wormhole-connect/issues/1497
**And edit vite config to only evaluate the src file**

We have three cases:

image

- The transaction is instantly completed

image

- In the case of some chains, we don't have the data to calculate the time, so we would leave it as it was originally.

image

The list of the chain we can know the estimated time:

- Ethereum
- Solana
- Polygon
- Bsc
- Fantom
- Celo
- Moonbeam
- Avalanche
- Sui
- Algorand
- Aptos
- Sei
- Near
- Terra
- Terra2
- Xpla
- Injective
- Karura
- AcalaEthereum


barnjamin:
[This](https://github.com/wormhole-foundation/connect-sdk/pull/269) pr changes the polygon finality level, if this gets merged a fast follow should be to update to `0.3.2` with this change

sebastianscatularo (approved):

PR #1530, opened 2024-01-23
Don't show wrapped tBTC on chains deployed with canonical tBTC

### LINKED ISSUE #1529: Canonical tBTC deployments on different chains shown as wormhole wrapped
1. Select Ethereum as the from network
2. Open the token list
3. Filter the token list to tBTC
4. Check the results

- ✅ The tBTC that's native on Ethereum is marked as native 
- ❌ The list displays tBTC on other chains marked as wormhole wrapped, while the contract addresses point to the canincal tbtc deployment on those chains

While point 2 is technically true, it's misleading as tBTC has always been a burn and mint style token, with wormhole-wrapped flavors not being officially supported. We should find a way to filter them from the list.

kev1n-peters:
Addresses #1529

### DIFF
--- a/wormhole-connect/src/routes/bridge/baseRoute.ts
+++ b/wormhole-connect/src/routes/bridge/baseRoute.ts
@@ -22,6 +22,7 @@ import { formatGasFee, isIlliquidDestToken } from 'routes/utils';
 import { toDecimals } from 'utils/balance';
 import { NO_INPUT } from 'utils/style';
 import { hexlify } from 'ethers/lib/utils.js';
+import { isTBTCToken } from 'routes/tbtc/utils';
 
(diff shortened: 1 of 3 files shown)


wormhole-connect/src/routes/bridge/baseRoute.ts
```diff
@@ -41,6 +42,9 @@ export abstract class BaseRoute extends RouteAbstract {
     if (!token.tokenId && token.nativeChain !== chainName) {
       return false;
     }
+    if (isTBTCToken(token) && token.nativeChain !== chainName) {
```
kev1n-peters, line 45: We don't want any routes besides the `tbtc` route to support tbtc tokens

artursapek (approved):

PR #1568, opened 2024-01-29
Portico: Prevent transfers when slippage is too high

### LINKED ISSUE #1569: Prevent transfers when slippage is too high
## Description and context

Slippage tolerance is currently based on the number from calling `quoteExactInputSingle`. There should be an additional check to prevent the user from signing a transaction that would result in slippage higher than a certain amount (50bps) **from the original input value**.

## Definition of done

User is prevented to sign a transaction that results in overall slippage higher than a hardcoded value.

kev1n-peters:


kev1n-peters:
Prevent users from transferring when slippage is > 50bps.

yuli-ferna (approved):

PR #1582, opened 2024-01-31
feat: initial injective support

bangjelkoski:


wormhole-connect/src/utils/wallet/index.ts
```diff
@@ -224,18 +225,28 @@ export const getWalletOptions = async (
...
-        wallets: { cosmos },
+        wallets: { cosmosEvm },
       } = await import('utils/wallet/cosmos');
-      return Object.values(mapWallets(cosmos, Context.COSMOS));
-    } else {
+
+      return Object.values(
+        mapWallets(cosmosEvm, Context.COSMOS, ['OKX Wallet']),
+      );
+    } else if (config.id === CHAIN_ID_INJECTIVE) {
```
artursapek, line 236: you could combine this with the case above, right?

`if (config.id === CHAIN_ID_EVMOS || config.id === CHAIN_ID_INJECTIVE) {`

artursapek (approved):
changes look fine. I left one comment on your wallet code changes. other than that, this lgtm.

wormhole-connect/src/utils/wallet/index.ts
bangjelkoski: We might wanna use different wallets in the future which is why I've left this `else if` in case we need it. I can remove it if you feel like that's the correct approach

Issue #1287, opened 2023-11-30
[Enhancement] Integrator-configurable arbitrary token bridging


aadam-10:
- Enable integrators to add tokens that they want to support in Connect. 
- In this iteration these tokens should use the 2-tx route.
- There needs to be validation to disallow duplicate tokens (among other things)

rhlsthrm:
Subscribing to follow. Please reach out if you need beta testers, we are willing to pilot this feature.

tradefortendies:
Subscribing to follow.

tsadovska:
Results of BONK transactions on https://preview.portalbridge.com/7246284e5d/ [Mainnet]: 

![image](https://github.com/wormhole-foundation/wormhole-connect/assets/10746654/2d67111d-02c5-4f52-8d64-d956f0ffc1f0)

PR #1601, opened 2024-02-04
Add BSKT to wormhole connect token selection. 

crypt0xg:
BSKT is looking to use wormhole connect for an easy to use interface to travese cross chain. 

PR submitted for details to get added. 

PR also submitted to SDK repo to add details there also.

Please let us know if more submissions are needed.

### DIFF
--- a/builder/src/consts.ts
+++ b/builder/src/consts.ts
@@ -179,6 +179,7 @@ export const MAINNET_TOKEN_KEYS: string[] = [
   "EVMOS",
   "KUJI",
   "PYTH",
+  "BSKT",
   "WETHpolygon",
   "WETHbsc",
   "wstETHarbitrum",
--- a/wormhole-connect/src/config/mainnet/tokens.ts
+++ b/wormhole-connect/src/config/mainnet/tokens.ts
@@ -2664,4 +2664,73 @@ export const MAINNET_TOKENS: TokensConfig = {
       },
     },
   },
+  BSKT: {
+    key: 'BSKT',
+    symbol: 'BSKT',
+    nativeChain: 'solana',
+    tokenId: {
+      chain: 'solana',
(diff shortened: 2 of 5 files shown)


artursapek:
hi - we not adding new tokens at this time. however we do hope to provide a way for developers to add arbitrary tokens via a config when deploying Connect on their website. you can track progress on this in https://github.com/wormhole-foundation/wormhole-connect/issues/1287.

crypt0xg:
@artursapek Thanks for the response. A little frustrating to be honest. Is there a particular reason we can't get added? Kind of goes against the whole decentralization aspect and "open source" nature if we're told that token additions are behind a gate like this.

We're trying to make the experience for users, understanding why we're being blocked, when there's no similar requests would be helpful to be able to explain to the community.

artursapek:
@crypt0xg you're welcome to fork the repo and add it to your fork, and deploy that! as you said it's completely open source. we just want to avoid the pull requests end up being an overwhelming firehose of listing requests.

crypt0xg:
@artursapek Appreciate that. With us being the first request and likely to have the most cross-chain connects this week than any other token, I was hoping there would be a little more latitude. Will fork and deploy.

aadam-10:
Hi @crypt0xg, to give a little more context, the process to add tokens in the main repo in this manner isn't sustainable and we simply can't take on the responsibility of handling these requests (which are numerous). The main repo itself must have an owner too, so even automating the acceptance of such requests would still mean that acceptance depends on the maintainers in some way, so it wouldn't be much help on the decentralization front.

We will be mitigating this problem in stages: first to enable integrators to easily configure supported tokens, then to allow Connect to handle arbitrary tokens without configuration, and later to automate collecting metadata about any arbitrary token so that the UI looks appropriate even when dealing with an unknown token. While I don't have exact dates right now, these improvements aren't far off.

PR #1674, opened 2024-02-16
Feature: Native Token Transfers (NTT) integration

### LINKED ISSUE #1585: Integrate NTT route for Wormhole Connect

### LINKED ISSUE #1857: Fantom USDC bridge support

kev1n-peters:
Addresses #1585

Go through the the NTT [README](https://github.com/wormhole-foundation/example-native-token-transfers/blob/main/README.md) to gain a high-level understanding. NTT is integrated into Connect via two routes: `nttManual` and `nttRelay`. These routes currently support EVM chains and Solana. This implementation supports managers configured with a single wormhole transceiver. The `nttManual` route requires users to pay a fee on the source and destination chain, while the `nttRelay` route requires users to pay a relayer fee on the source chain only (in native gas units). The `nttRelay` route utilizes standard relayers for transfers between EVM chains and a specialized relayer for transfers to and from Solana. The route code resides in `wormhole-connect/src/routes/ntt`.


wormhole-connect/src/routes/ntt/nttBase.ts
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  async isRouteSupported(
+    sourceToken: string,
+    destToken: string,
+    amount: string,
+    sourceChain: ChainName | ChainId,
+    destChain: ChainName | ChainId,
+  ): Promise<boolean> {
+    return Promise.all([
```
sebastianscatularo: maybe this could be more clear if we use await here

```suggestion
    try {
    
    const results = await Promise.all([
      this.isSupportedChain(wh.toChainName(sourceChain)),
      this.isSupportedChain(wh.toChainName(destChain)),
      this.isSupportedSourceToken(
        TOKENS[sourceToken],
        TOKENS[destToken],
        sourceChain,
        destChain,
      ),
      this.isSupportedDestToken(
        TOKENS[destToken],
        TOKENS[sourceToken],
        sourceChain,
        destChain,
      )]);
    return result.every(result => result);
    } catch (e: any) {
    // for debug porpouse
    console.trace(e);
    return false;
    }
```

wormhole-connect/src/routes/ntt/nttRelay.ts
```diff
@@ -0,0 +1,203 @@
...
+    destToken: string,
+    amount: string,
+    sourceChain: ChainName | ChainId,
+    destChain: ChainName | ChainId,
+  ): Promise<boolean> {
+    const managerAddress = TOKENS[sourceToken]?.nttManagerAddress;
+    if (!managerAddress) {
+      return false;
+    }
+    return Promise.all([
```
sebastianscatularo: same as above

wormhole-connect/src/routes/ntt/platforms/index.ts
```diff
@@ -0,0 +1,17 @@
+import { isEvmChain, wh } from 'utils/sdk';
+import { ChainName, ChainId } from '@wormhole-foundation/wormhole-connect-sdk';
+import { NTTEvm } from './evm';
+import { NTTSolana } from './solana';
+
+export const getPlatform = (
+  chain: ChainName | ChainId,
+  managerAddress: string,
+): NTTEvm | NTTSolana => {
+  if (isEvmChain(chain)) {
```
sebastianscatularo: all evm will be supported initially by NTT?

wormhole-connect/src/views/Bridge/NTTCapacityWarning.tsx
```diff
@@ -0,0 +1,101 @@
...
+    fetchCapacity();
+    return () => {
+      active = false;
+    };
+  }, [chain, token, managerAddress, fromChain]);
+
+  useEffect(() => {
+    if (!chain || !managerAddress) return;
+    let active = true;
+    const fetchRateLimitDuration = async () => {
```
sebastianscatularo: looks like the async is unnecessary here

wormhole-connect/src/routes/ntt/platforms/index.ts
kev1n-peters: The NTT contracts can be deployed to any EVM chain so this route supports all EVM chains

wormhole-connect/src/routes/ntt/nttBase.ts
kev1n-peters: agreed

wormhole-connect/src/routes/ntt/nttRelay.ts
kev1n-peters: agreed

wormhole-connect/src/routes/ntt/platforms/evm/getMessage.ts
```diff
@@ -0,0 +1,108 @@
...
+  if (!token) {
+    throw new Error(`Token ${tokenId} not found`);
+  }
+  const wormholeLog = await getWormholeLogEvm(fromChain, receipt);
+  const parsedWormholeLog =
+    Implementation__factory.createInterface().parseLog(wormholeLog);
+  const relayingInfoEvent = receipt.logs.find(
+    (log) =>
+      log.topics[0] ===
+      '0x375a56c053c4d19a2e3445e97b7a28bf4e908617ce6d766e1e03a9d3f5276271',
```
artursapek: should this go into Connect's internal SDK as a const?

wormhole-connect/src/routes/types.ts
```diff
@@ -41,14 +66,18 @@ export const isSignedCCTPMessage = (
   typeof message === 'object' &&
   'message' in message &&
   'attestation' in message;
+export const isSignedNttMessage = (
+  message: SignedMessage,
+): message is SignedNttMessage =>
+  isSignedWormholeMessage(message) && isUnsignedNttMessage(message);
 
 export interface TransferInfoBaseParams {
-  txData: ParsedMessage | ParsedRelayerMessage;
```
artursapek, line 46: is there a way to avoid adding this type for a single route (NTT) to this type which is used by all routes?

wormhole-connect/src/store/redeem.ts
```diff
@@ -36,7 +39,11 @@ export const redeemSlice = createSlice({
   reducers: {
     setTxDetails: (
       state: RedeemState,
-      { payload }: PayloadAction<ParsedMessage | ParsedRelayerMessage>,
+      {
+        payload,
+      }: PayloadAction<
+        ParsedMessage | ParsedRelayerMessage | UnsignedNttMessage
```
artursapek, line 45: same question here. It seems like we're adding route-specific code to a generic function

wormhole-connect/src/routes/ntt/platforms/evm/getMessage.ts
kev1n-peters: Yeah, we could move it to a consts file and name it something descriptive like `RELAYING_INFO_EVENT_TOPIC`

wormhole-connect/src/utils/transferValidation.ts
```diff
@@ -311,7 +317,7 @@ export const validateAll = async (
     receiveAmount: '',
   };
 
-  if (isAutomatic) {
```
artursapek, line 314: same comment here - this is generic code but now has NTT-specific logic

wormhole-connect/src/routes/types.ts
kev1n-peters: Given that only the `relayerFee` property is used in the `UnsignedNttMessage` and it's already present on the `ParsedRelayerMessage`, I think this type can be removed from the union. However `ParsedRelayerMessage` is currently specific to the `relay` route, so it's still kind of a code smell IMO. I feel like this should be refactored so we don't have to rely on these union types.

wormhole-connect/src/store/redeem.ts
kev1n-peters: The properties on `UnsignedNttMessage` are needed throughout the codebase for the ntt routes. I'm not really sure what the right solution is here given that `ParsedRelayerMessage` is also route-specific. Maybe we could remove `UnsignedNttMessage` and add an optional `ntt` property to `ParsedMessage` instead, but I feel like that's not really solving the issue at its core

sdk/src/config/TESTNET.ts
```diff
@@ -55,12 +55,6 @@ const TESTNET: { [chain in TestnetChainName]: ChainConfig } = {
     contracts: {
       ...CONTRACTS.TESTNET.ethereum,
       relayer: '0x9563a59c15842a6f322b10f69d1dd88b41f2e97b',
-      cctpContracts: {
```
kev1n-peters, line 58: Goerli is deprecated, enabled Sepolia CCTP instead

wormhole-connect/package.json
```diff
@@ -30,7 +31,7 @@
     "@reduxjs/toolkit": "^1.9.1",
     "@solana/wallet-adapter-wallets": "^0.19.25",
     "@solana/web3.js": "^1.73.0",
-    "@wormhole-foundation/sdk-definitions": "^0.5.0",
+    "@wormhole-foundation/sdk-definitions": "^0.5.2-beta.0",
```
kev1n-peters, line 34: this is currently the only version with ntt layouts

wormhole-connect/src/AppRouter.tsx
```diff
@@ -71,6 +73,8 @@ function AppRouter({ config }: Props) {
     if (prevRoute === redeemRoute && route !== redeemRoute) {
       dispatch(clearRedeem());
       dispatch(clearWallets());
+      dispatch(clearNtt());
```
artursapek, line 76: I'm ok with this for now, but stuff like this is going to make it harder to do the SDKv2 migration. in theory `AppRouter` should never have route-specific logic in it.

wormhole-connect/src/config/testnet/tokens.ts
```diff
@@ -2232,4 +2262,84 @@ export const TESTNET_TOKENS: TokensConfig = {
...
+      address: '0x0e15979a7a1eFAEf20312CA45A59eb141bF7E340',
+    },
+    coinGeckoId: '',
+    color: '#2774CA',
+    decimals: {
+      Ethereum: 18,
+      default: 8,
+    },
+  },
+  TEST_NTTsolana: {
```
artursapek: is there a way to restructure our config so we don't have this be so repetitive? could all of these `TEST_NTT...` configs live under a single `TEST_NTT` entry?

kev1n-peters: These are actual tokens (they just happen to be prefixed with `TEST_NTT`, they can be bridged via other routes, too). I think extending how token configs are defined is outside of the scope of this PR

kev1n-peters: For example, `USDCsepolia` is configued in the `TEST_USDC` ntt group

wormhole-connect/src/store/transferInput.ts
```diff
@@ -182,7 +184,14 @@ const performModificationsIfFromChainChanged = (state: TransferInputState) => {
         state.amount = '';
       }
     }
-    if (
+    if (isNttRoute(route) && destToken) {
+      const groupKey = getNttGroupKey(tokenConfig, config.tokens[destToken]);
+      if (groupKey) {
+        state.token = getNttTokenByGroupKey(groupKey, fromChain!)?.key || '';
```
kev1n-peters: fromChain can be undefined

wormhole-connect/src/store/transferInput.ts
```diff
@@ -204,14 +213,24 @@ const performModificationsIfFromChainChanged = (state: TransferInputState) => {
...
   if (destToken) {
     const tokenConfig = config.tokens[destToken];
     if (!toChain) {
       state.destToken = '';
     }
-    if (tokenConfig.symbol === 'USDC' && tokenConfig.nativeChain !== toChain) {
+    if (isNttRoute(route) && token) {
+      const groupKey = getNttGroupKey(tokenConfig, config.tokens[token]);
+      if (groupKey) {
+        state.destToken = getNttTokenByGroupKey(groupKey, toChain!)?.key || '';
```
kev1n-peters: toChain can be undefined

wormhole-connect/src/utils/transferValidation.ts
kev1n-peters: Agree, routes should be able to validate themselves

artursapek:
gonna track test cases I do on https://deploy-preview-1674--wormhole-connect-mainnet.netlify.app/

- [x] CCTP, mainnet: USDC from Solana to Ethereum ✅
- [x] Token bridge, testnet: WMATIC from Polygon to Fantom ✅

wormhole-connect/src/routes/ntt/nttBase.ts
```diff
@@ -0,0 +1,465 @@
...
+    token: TokenId | 'native',
+    amount: string,
+    sendingChain: ChainName | ChainId,
+    senderAddress: string,
+    recipientChain: ChainName | ChainId,
+    recipientAddress: string,
+    routeOptions?: any,
+  ): Promise<BigNumber> {
+    if (isEvmChain(sendingChain)) {
+      const gasLimit = this.TYPE === Route.NttManual ? 200_000 : 250_000;
```
kev1n-peters: Make sure these are right

artursapek (approved):

PR #1676, opened 2024-02-18
fix: modal-theme-update-on-state-change

ChefBingbong:
Bug fix for the issue from widget v0.2.0 and up where the widgets state wont update when the config changes on theme change or env change. 

I believe the migration to vite caused the bug as ever since the migration, the trick of passing a key to the react widget as the strinified config stopped working and caused the widget to not mount on change in the config

solution is just to manage some the worm-hole apps config as state variables instad of just inline exports such as most of the config is.

the widgets config really would only change either when the theme is changed or the env is changed. all other params either depend on these, or woill likely not change. so for now i extended the theme config and the env/netowrk config items to a context at the top of the app

updating the widget externally seems to never work by conventional means in terms of trying to cause internal rerenders when the config changes. im not sure why exactly this is spent a while trying many ways. i think maybe its because the config is initialised in-line and non stateful and el the selector that gets the wormhole-connect is too. so i wonder does the config just nevr change because the worm-hole div selector never changes. it initialized once in current implementation and never updates as config changes.

so thinking this i add the config initialuzation and this selector init to a useeffect and create a custom event trigger to update the  config when triggered. every is fixed now the widget now updates on config change


artursapek:
thanks for the PR! I'm addressing this in #1698 by allowing the theme to be updated via a normal React prop. going to close this one.

PR #1682, opened 2024-02-19
Add warning for transfer larger than CCTP max transfer limit (1M) 

### LINKED ISSUE #1504: Add warning for transfer larger than CCTP max transfer limit (1M)
A user hit a limit of 100k on the CCTP bridge in Connect, we know Circle imposes limits of 1M on CCTP transfers (dependent on the chain) but are unclear on where this 100k limit is being imposed. We need to investigate where its propagating from (doesn't seem to be relayers). My hypothesis is that the CCTP limits are on an address basis (we hit CCTP from the WH CCTP contracts) and over some small window of time, amounting perhaps to 1M in the period this user tried it.

1. find the source of the limit and remedy it or decide to keep it if its unsolvable or legitimate 
2. implement elegant error handling that matches the UI of "amount must be greater than 0" below:  
Screenshot 2024-01-19 at 4 53 10 AM

yuli-ferna:
Issue: https://github.com/wormhole-foundation/wormhole-connect/issues/1504

I decided to keep a const value because query in [TokenMinter contracts](https://developers.circle.com/stablecoins/docs/evm-smart-contracts#tokenminter-mainnet) the result in burnLimitsPerMessage method with the USDC token is the following:

image
The result with decimals is: 1,000,000.00

Result:
image

### DIFF
--- a/wormhole-connect/src/utils/transferValidation.ts
+++ b/wormhole-connect/src/utils/transferValidation.ts
@@ -112,7 +112,9 @@ export const validateAmount = (
     if (numAmount > b) return 'Amount cannot exceed balance';
   }
   if (isCctp && numAmount >= CCTP_MAX_TRANSFER_LIMIT)
-    return 'Your transaction exceeds the maximum transfer limit of 1,000,000 USDC. Please use a different amount.';
+    return `Your transaction exceeds the maximum transfer limit of ${Intl.NumberFormat(
+      'en-EN',
+    ).format(CCTP_MAX_TRANSFER_LIMIT)} USDC. Please use a different amount.`;
   if (numAmount > maxAmount) {
     return `At the moment, amount cannot exceed ${maxAmount}`;
   }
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/utils/transferValidation.ts
```diff
@@ -108,6 +111,8 @@ export const validateAmount = (
     const b = Number.parseFloat(balance);
     if (numAmount > b) return 'Amount cannot exceed balance';
   }
+  if (isCctp && numAmount >= CCTP_MAX_TRANSFER_LIMIT)
+    return 'Your transaction exceeds the maximum transfer limit of 1,000,000 USDC. Please use a different amount.';
```
sebastianscatularo: would be good to get the amount from the constant and format, so, if we need to change the amount, we just need to change the const value

sebastianscatularo (changes requested):

wormhole-connect/src/utils/transferValidation.ts
yuli-ferna: Done!

sebastianscatularo (approved):

PR #1692, opened 2024-02-21
Remove dead file index.tsx

artursapek:
@chefbingbong pointed out that `index.tsx` is no longer in use since the Vite change. I accidentally left it checked in. this change reverts back to it and removes the duplication.


kev1n-peters (approved):

sebastianscatularo (approved):
LGTM 🚀

PR #1698, opened 2024-02-22
Direct import & config overhaul

### LINKED ISSUE #1715: Library mode (direct import) and config overhaul
## Tl;dr

- Integrators can change the color scheme dynamically using React props
- The `wormhole-connect-loader` is going bye-bye
- Connect will be importable by third party apps as a regular React component without relying on unpkg.com
- Using unpkg.com will still be an option for people who prefer to just copy & paste HTML tags

## Description and context
This ticket represents a large change to the code base. There are two main goals behind it:

1. Make it possible to change the Connect config dynamically at runtime (for example, swap between two custom color schemes)
2. Make it possible to import the Connect widget into a React app, bundle it, and host it directly (not using unpkg.com)
3. Get rid of unpkg.com as a central point of failure for Connect

Neither of these is possible today due to how Connect is published and how it reads its config from the DOM.

## Implementation Details
The first goal, dynamic config, requires a change to how Connect gets its custom config.. The original Connect code expected the config to be a stringified JSON object in the `config` attribute on the `wormhole-connect` element. This gets parsed when the `config.ts` file is *first evaluated* by the browser, and is static from then on.

```
<div
  id="wormhole-connect"
  config="{"walletConnectProjectId":"e5ef8b58d14ead228f38db4eb5c7148b","env":"mainnet","rpcs":{"wormchain":"...`
></div>
```

This makes it impossible to use Connect as a component in a React app, because it expects a specific DOM node to exist before React even starts running.

We are moving to the config & theme being passed as props to the top level component:

```
<WormholeConnect
  theme={myCustomTheme}
  config={myCustomConfig}
/>
```

The theme is no longer a part of the config, it's a separate prop altogether. The component no longer expects these to be hard-coded into the DOM.

For now, we will still disallow updates to the config during runtime because this can cause lots of weird, broken behavior. For example, switching from `testnet` to `mainnet` mid-transfer causes Connect to get very confused.

This first change enables the second change: allowing apps to import Connect itself as a component from NPM. Importing Connect this way will no longer load the "loader" which merely injects unpkg.com tags into your HTML. It will actually import the Connect component itself.

`wormhole-connect-loader`, which exists for the sole purpose of exporting a React component which simply injects two HTML tags loading Connect from unpkg.com, is being taken out back.

We will still support the HTML copy-pasta approach of integrating Connect for people not using React.

### Config imports

The main disruptive aspect of this change is how we use `config`. Before this change, the config was exported by `src/config/index.ts` as dozens of all-caps consts like:

```
export const ROUTES = config && config.routes ? config.routes : Object.values(Route);
```

Instead, we will export a single object called `config` and the code base will import that, and refer to `config.routes` as opposed to just `ROUTES`.

### Duplicate React

Today, any website which uses React and also uses Connect is making their users load React twice (since it's in the Connect bundle). That sucks! This ticket will eliminate that problem.

artursapek:
Addresses https://github.com/wormhole-foundation/wormhole-connect/issues/1715 (click to learn more)

image


package-lock.json
```diff
@@ -40,7 +39,7 @@
         "@types/node": "^16.18.25",
         "@types/react": "^18.2.0",
         "@types/react-dom": "^18.2.1",
-        "@wormhole-foundation/wormhole-connect": "^0.1.4",
+        "@wormhole-foundation/wormhole-connect": "file:../wormhole-connect/wormhole-foundation-wormhole-connect-0.0.1-beta.0.tgz",
```
kev1n-peters: Should we change this before merging or is it needed until a new version of wormhole-connect with this PR's changes is published?

wormhole-connect/src/components/Background/BackgroundImage.tsx
```diff
@@ -118,21 +117,8 @@ type Props = {
...
-        <div className={classes.gradientRight}></div>
-        <div className={classes.gradientRight2}></div>
-        <div className={classes.gradientLeft}></div>
-        <div className={classes.gradientLeft2}></div>
-      </div>
-    </div>
-  ) : (
-    <div className={classes.children}>{children}</div>
-  );
+  // TODO remove this component
```
kev1n-peters: Should we remove the `Background` component?

package-lock.json
artursapek: didnt mean to include this

wormhole-connect/src/components/Background/BackgroundImage.tsx
artursapek: yeah thanks :) done

CONTRIBUTING.md
```diff
@@ -1,3 +1,31 @@
...
+
+Run `npm i` at the root of the repo
+
+2) Build
+
+Run `npm run build` at the root fo the repo
+
+3) Start
+
+Start wormhole-connect UI and view in browser at localhost:3000
```
kev1n-peters, line 17: nit: `localhost:5173/` since the vite change

kev1n-peters (approved):

CONTRIBUTING.md
artursapek: will fix separately. not a new issue

PR #1755, opened 2024-03-06
#1697: add priority fee to Solana transactions

### LINKED ISSUE #1697: Add integrator way to configure Solana priority fees
## Description and context
Solana transactions require setting a priority fee to land transactions during periods of high congestion.

See:
- https://docs.solanalabs.com/proposals/fee_transaction_priority
- https://docs.helius.dev/solana-rpc-nodes/alpha-priority-fee-api

I propose adding a config that allows users to pass in a method for calculating appropriate Solana priority fees to set per transaction. Passing in a method allows integrators to customize how they calculate this, including:
- using Solana RPC `getRecentPriorityFees` to estimate based on past priority fees
- using Helius `getPriorityFeeEstimate` to lean on an external service to estimate appropriate priority fees

artursapek:
addresses https://github.com/wormhole-foundation/wormhole-connect/issues/1697

adds a compute unit price of 75th percentile of recent fees paid

### DIFF
--- a/sdk/src/contexts/solana/context.ts
+++ b/sdk/src/contexts/solana/context.ts
@@ -1219,12 +1219,12 @@ export class SolanaContext<
   ): Promise<TransactionInstruction[]> {
     let fee = 100_000; // Set fee to 100,000 microlamport by default
 
-    const recentFeesResponse =
-      await this.connection?.getRecentPrioritizationFees({
-        lockedWritableAccounts,
-      });
-
     try {
+      const recentFeesResponse =
+        await this.connection?.getRecentPrioritizationFees({
+          lockedWritableAccounts,
+        });
+
       if (recentFeesResponse) {
         // Get 75th percentile fee paid in recent slots
         const recentFees = recentFeesResponse
--- a/sdk/src/contexts/solana/context.ts
+++ b/sdk/src/contexts/solana/context.ts
@@ -274,7 +274,7 @@ export class SolanaContext<
       payerPublicKey,
       tokenPublicKey,
     );
-    const transaction = new Transaction().add(createAccountInst);
+    const transaction = this.newTransaction().add(createAccountInst);
     const { blockhash } = await this.connection.getLatestBlockhash(commitment);
     transaction.recentBlockhash = blockhash;
     transaction.feePayer = payerPublicKey;
@@ -387,13 +387,11 @@ export class SolanaContext<
     );
 
     const { blockhash } = await this.connection.getLatestBlockhash(commitment);
-    const transaction = new Transaction();
-    const priorityIx = await this.getPriorityFeeIx(contracts.token_bridge);
+    const transaction = this.newTransaction();
 
     transaction.recentBlockhash = blockhash;
     transaction.feePayer = payerPublicKey;
     transaction.add(
-      priorityIx,
       createAncillaryAccountIx,
       initialBalanceTransferIx,
       initAccountIx,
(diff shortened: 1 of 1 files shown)


sdk/src/contexts/solana/context.ts
```diff
@@ -1176,4 +1187,25 @@ export class SolanaContext<
       chain: 'solana',
     };
   }
+
+  async getPriorityFeeIx(address: string) {
+    let unitPrice = 10_000; // use minimum of 10,000 microlmaports
```
tonyjin: Misspelled

sdk/src/contexts/solana/context.ts
```diff
@@ -1176,4 +1187,25 @@ export class SolanaContext<
...
+
+    const recentFees = await this.connection?.getRecentPrioritizationFees({
+      lockedWritableAccounts: [new PublicKey(address)],
+    });
+
+    if (recentFees) {
+      const maxFee = Math.max(
+        ...recentFees.map(({ prioritizationFee }) => prioritizationFee),
+      );
+      unitPrice = Math.max(unitPrice, maxFee * 10);
```
tonyjin: Would be good to document why 10x chosen here. Also what happens if there are multiple Connect transactions submitted within a short period of time, could the later ones detect earlier ones resulting in 10*10*10...

sdk/src/contexts/solana/context.ts
artursapek: lmao-ports. thanks

sdk/src/contexts/solana/context.ts
artursapek: yeah I had this same worry. I was just trying this to see if it helped with timeouts, but it wasn't. I don't want to keep this `* 10` bit. I'm not even sure we want to use this RPC endpoint at all because it's limited to 60 seconds of history.

artursapek:
we could make this configurable by letting the integrator pass in a percentile. i think shipping at 75th and seeing how it goes is fine for now though.

sdk/src/contexts/solana/context.ts
```diff
@@ -77,6 +78,9 @@ import {
 const SOLANA_SEQ_LOG = 'Program log: Sequence: ';
 const SOLANA_CHAIN_NAME = MAINNET_CONFIG.chains.solana!.key;
 
+// Add priority fee according to 75th percentile of recent fees paid
+const SOLANA_FEE_PERCENTILE = 0.75;
```
barnjamin, line 82: this smth that should go in an overridable config?

sdk/src/contexts/solana/context.ts
```diff
@@ -1073,37 +1098,49 @@ export class SolanaContext<
...
-        formattedRecipient,
-        recipientChainId,
-        nonce,
-        wrapToken,
+
+      transaction.add(
+        ...(await this.determineComputeBudget(writableAddresses)),
+      );
+      transaction.add(
+        await createTransferNativeTokensWithRelayInstruction(
```
barnjamin, line 1106: could do this and the above concurrently?

barnjamin (approved):

kev1n-peters (approved):

sebastianscatularo (approved):

sdk/src/contexts/solana/context.ts
artursapek: definitely could, but I wanna get this change in first. can always do that later

sdk/src/contexts/solana/context.ts
artursapek: valid suggestion but IMO not worth the code complexity

PR #1771, opened 2024-03-11
fix: bridge not show after router change in SPA

alexayan:
Modules loaded with the script tag will only be executed once.
When the Bridge component is unmount from the page and remount, it will not be render again.


artursapek:
https://github.com/wormhole-foundation/wormhole-connect/issues/1715 obsoletes this change. you can try it right now: [0.3.0-beta.5-development](https://www.npmjs.com/package/@wormhole-foundation/wormhole-connect/v/0.3.0-beta.5-development)

Issue #1769, opened 2024-03-10
[Bug] Wrong provider returned after rejecting a TX and swapping chains


kev1n-peters:
Steps to reproduce:

Approve a TX for any route and then reject the TX from Metamask. Now swap the chains and approve the TX again. Notice that `wh.mustGetProvider(sourceChain)` and `wh.mustGetProvider(targetChain)` are the same provider (have the same EVM chain ID).

kev1n-peters:
I suspect that `registerSigner` is setting the source chain's provider as a reference to `signer.provider` which is problematic when switching chains (because different chains reference the same provider). https://github.com/wormhole-foundation/wormhole-connect/blob/f52e719da48d81c8e5bb2630445aa0ef94bf3203/wormhole-connect/src/utils/wallet/index.ts#L65. Calling `wh.registerProviders()` before switching chains fixes this, but it seems like a potential bug in the @nomad-xyz/multi-provider package.

kev1n-peters:
The issue stems from the @nomad-xyz/multi-provider package. When calling `registerSigner`, if the call to `signer.connect` fails, then the `provider` for the `domain` will be set to `signer.provider`. https://github.com/nomad-xyz/monorepo/blob/9b6f68278cf687f9f19065d52b0fac6a128346b1/packages/multi-provider/src/provider.ts#L242-L256 This is an issue, because `signer.provider` may be connected to a chain different than the `domain`. This is the exception thrown when `signer.connect` is called: https://github.com/ethers-io/ethers.js/issues/1088#issuecomment-704347898

PR #1777, opened 2024-03-11
Fix EVM provider reference issue

kev1n-peters:
Addresses #1769

This PR addresses an issue found in the `MultiProvider` class from the `@nomad-xyz/multichain-provider` package, which is no longer maintained. The relevant source code from the package has been copied into a new file, multi-provider.ts. The original `MultiProvider` class from the package contained a problematic line (https://github.com/nomad-xyz/monorepo/blob/9886b9f1ac5c7bc27c2a135c8ebcd51ff99b70df/packages/multi-provider/src/provider.ts#L255), which caused the registration of a signer to also register its provider. This PR comments out that line, preventing the provider from being registered when a signer is registered (see the change in the `MultiProvider` class here: https://github.com/wormhole-foundation/wormhole-connect/pull/1777/files#diff-81a5051cf616340ceb8ae37f59516d72dfc3534be65722ca56958a1162fe779eR280). This issue arose when multiple chains had signers registered, as each chain would reference the same provider (`signer.provider`), which can only be connected to one network. This posed a problem when attempting to use providers for multiple chains after multiple signers had been registered.

Note: Upon further inspection, it appears that `MultiProvider` is not designed to be used with a single instance of a `signer` registered for multiple chains/domains. For example, `mustGetConnection` is going to return the `provider` if the domain has a `signer` registered, but there's no guarantee that the `provider` is connected to the network corresponding to the chain/domain.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36938,7 +36938,6 @@
         "@cosmjs/stargate": "^0.31.3",
         "@cosmjs/tendermint-rpc": "^0.31.3",
         "@mysten/sui.js": "^0.32.2",
-        "@nomad-xyz/multi-provider": "^1.1.0",
         "@solana/spl-token": "^0.3.7",
         "@solana/web3.js": "^1.73.0",
         "aptos": "1.5.0",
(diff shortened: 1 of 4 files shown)


artursapek (approved):
ok so it looks like we are just adding providers via `registerRpcProvider`. makes sense. looks good.

PR #1786, opened 2024-03-12
chore: add arb token

shane-moore:
## Changes

- added mainnet ARB token metadata so the token can be exchanged via wormhole connect


aadam-10:
Apologies but currently we're not accepting token additions in the `wormhole-connect` repo via this method. Wormhole Connect can be configured by the integrator to support any token and we encourage them to do so.

You can see how here: https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/config/examples.md

PR #1795, opened 2024-03-14
solana: add priority fee to redeem transactions

### LINKED ISSUE #1808: Adjust compute budgets in Solana to address congestion
## Description and context

Experiment with adjustments to the compute budget (default is 200k per instruction which may be too much) as a means to improve txn error rates on Solana

## Definition of done

200k compute unit set to a lower value such as 80-100k. Observe improved txn rates, if possible.
Verify no regression to manual + automatic txns for Solana

kev1n-peters:


sdk/src/contexts/solana/context.ts
```diff
@@ -1239,9 +1250,6 @@ export class SolanaContext<
     console.info(`Setting Solana compute unit price to ${fee} microLamports`);
 
     return [
-      ComputeBudgetProgram.setComputeUnitLimit({
```
kev1n-peters, line 1242: I think setting the compute unit limit is unnecessary since, "If a SetComputeUnitLimit instruction isn’t provided, the limit is calculated as the product of the number of instructions in the transaction and the [default compute unit limit](https://github.com/solana-labs/solana/blob/4293f11cf13fc1e83f1baa2ca3bb2f8ea8f9a000/program-runtime/src/compute_budget.rs#L13)." - it could become an issue if the units we set aren't enough for a transaction with multiple instructions (I think we should let the Solana runtime determine this value dynamically)

wormhole-connect/src/utils/wallet/index.ts
```diff
@@ -160,6 +157,8 @@ export const postVaa = async (
   if (!wallet) throw new Error('not connected');
   const pk = (wallet as any).adapter.publicKey;
   const MAX_VAA_UPLOAD_RETRIES_SOLANA = 5;
+  const computeBudgetInstructions =
+    await solanaContext().determineComputeBudget([]);
```
kev1n-peters: It's difficult to know which accounts to pass here since we have to know the route from which this VAA is for. Even then it seems like it would be complex and a lot of code to figure that out. I think it's sufficient for the fee to be at least 100_000 micro-lamports here.

sdk/src/contexts/solana/context.ts
artursapek: makes sense 👍

wormhole-connect/src/utils/wallet/index.ts
```diff
@@ -4,20 +4,17 @@ import {
   Context,
   SendResult,
   ChainConfig,
-} from '@wormhole-foundation/wormhole-connect-sdk';
-import {
   postVaaSolanaWithRetry,
```
kev1n-peters, line 7: We were previously using `postVaaSolanaWithRetry` from `@certusone/wormhole-sdk`, but now we are using a modified version that sets the priority fee from `@wormhole-foundation/wormhole-connect-sdk`.

sdk/src/contexts/solana/context.ts
```diff
@@ -1230,7 +1238,10 @@ export class SolanaContext<
         const recentFees = recentFeesResponse
           .map((dp) => dp.prioritizationFee)
           .sort((a, b) => a - b);
-        fee = recentFees[Math.floor(recentFees.length * SOLANA_FEE_PERCENTILE)];
+        fee = Math.max(
+          recentFees[Math.floor(recentFees.length * SOLANA_FEE_PERCENTILE)],
+          fee,
+        );
```
artursapek: 👍

artursapek (dismissed):

artursapek (dismissed):

artursapek (dismissed):

sdk/src/contexts/solana/context.ts
```diff
@@ -953,8 +950,10 @@ export class SolanaContext<
           parsed.tokenChain,
           parsed.tokenAddress,
         );
-    const transaction = new Transaction();
-    transaction.add(...(await this.determineComputeBudget([mint])));
+
+    const transaction = new Transaction(
+      await this.connection?.getLatestBlockhash('finalized'),
```
barnjamin, line 955: why finalized? confirmed might give you a couple more seconds but generally we should probably not set these on a per call basis

sdk/src/contexts/solana/context.ts
```diff
@@ -1065,7 +1063,10 @@ export class SolanaContext<
     );
     const recipientChainId = this.context.toChainId(recipientChain);
     const nonce = createNonce().readUint32LE();
-    const transaction = new Transaction();
+    const transaction = new Transaction(
+      await this.connection?.getLatestBlockhash('finalized'),
```
barnjamin, line 1067: same

sdk/src/contexts/solana/utils/sendAndConfirmPostVaa.ts
```diff
@@ -40,6 +41,9 @@ export async function postVaaWithRetry(
       vaa,
       commitment,
     );
+  for (const unsignedTransaction of unsignedTransactions) {
+    addComputeBudget(connection, unsignedTransaction, []);
```
barnjamin: await?

sdk/src/contexts/solana/context.ts
artursapek: I just maintained what was there. check the diff.

image

sdk/src/contexts/solana/context.ts
artursapek: same answer

sdk/src/contexts/solana/utils/sendAndConfirmPostVaa.ts
artursapek: good catch thx

sdk/src/contexts/solana/context.ts
barnjamin: sure, but since we're here trying to make it work gooder, we should consider this up for debate

I'm positive @nonergodic has experience with varying consistency levels

barnjamin (dismissed):

sdk/src/contexts/solana/context.ts
barnjamin: approving regardless but we should make an issue to rant

nonergodic: Yeah, the general rule (though ofc there are exceptions) is that you don't want to override the commitment level of the connection (unless you really know what you are doing). It's really quite the footgun in the Solana SDK due to how juicy and tempting it is to just muck with it and suddenly all hell breaks loose, especially when someone changes the default commitment level on the connection.

This case seems to be a potential example of that. If you call `getLatestBlockhash` with finalized commitment, then I'd expect that blockhash to be comparatively old (32 slots) which can then lead to it timing out faster. I'm not sure if it is really a problem here in particular or whether this is the right approach (I'm mostly working on the on-chain side of things and so my knowledge of the solana web3 sdk and interacting with the chain via a node/validator are somewhat limited) but I have seen various related instances where somebody was like "I know, I'll just fetch the latest blockhash using `processed` commitment level because that will give me the newest one!" and then they tried to submit that transaction with a more robust commitment level and the Solana node said "I don't know this blockhash." (when it should have said "I don't know this blockhash at the more robust commitment level you have selected.").

nonergodic: an old example of these sorts of issues that came to my mind:
https://github.com/XLabs/portal-bridge-ui/pull/154#issuecomment-1558071589

artursapek (approved):

PR #1842, opened 2024-03-26
solana fix: setting compute budget when redeeming WSOL causes signatu…

kev1n-peters:
…re failure

### DIFF
--- a/sdk/src/contexts/solana/context.ts
+++ b/sdk/src/contexts/solana/context.ts
@@ -912,8 +912,8 @@ export class SolanaContext<
           signedVAA,
         );
 
-    transaction.feePayer = new PublicKey(payerAddr);
-    await addComputeBudget(this.connection!, transaction);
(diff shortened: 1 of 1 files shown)


sdk/src/contexts/solana/context.ts
```diff
@@ -912,8 +912,8 @@ export class SolanaContext<
           signedVAA,
         );
 
-    transaction.feePayer = new PublicKey(payerAddr);
-    await addComputeBudget(this.connection!, transaction);
+    // TODO: adding a compute budget causes the redeem to fail for WSOL, unsure of why
+    // await addComputeBudget(this.connection!, transaction);
```
barnjamin, line 916: intentionally also taking out fee payer?

kev1n-peters: It's already set on the transaction by the time we get here so it's redundant

artursapek (approved):

PR #1858, opened 2024-04-01
Update Builder to use 0.3

artursapek:
This makes several major changes to connect-in-style.wormhole.com:
- Moves the build system off `create-react-app` (which is unmaintained and SLOW) to `vite` with `swc`
- Updates the version of Connect it serves from 0.1 to 0.3
- Uses the new Connect component library, rather than serving the hosted unpkg.com version
  - This lets us update color scheme and network settings dynamically using React props
  - This also lets us render an actual mainnet version of Connect, with the "submit" button missing (via the new `previewMode` config)
- Removes a ton of duplicated configs and consts, now importing them from the `wormhole-connect` library

This also fixes a mistake in the build that causes the stylesheet to not load in the hosted unpkg.com build (this bug has been present since 0.3.0)


wormhole-connect/src/views/Bridge/Inputs/From.tsx
```diff
@@ -187,14 +187,16 @@ function FromInputs() {
         balance={balance}
         tokenPrice={getTokenPrice(prices, config.tokens[token])}
       />
-      <TokensModal
```
artursapek, line 190: rendering this modal when it's not even showing was sometimes causing errors in builder when switching between testnet and mainnet due to a race condition.

i think it's faster and simpler to just not render it (and run all of its code) in the default, zero-state view of Connect.

kev1n-peters (approved):

PR #1869, opened 2024-04-02
solana relayer: Added WSOL min send amount

kev1n-peters:
An attempt to relay WSOL less than the minimum rent exempt amount to a new account on Solana will result in a transaction failure. This PR sets the minimum transfer amount of WSOL to the min rent exempt amount when relaying to Solana. The amount is small enough that I don't think it justifies an extra network call / complexity to check if the account already exists / has a balance.

### DIFF
--- a/wormhole-connect/src/routes/abstracts/routeAbstract.ts
+++ b/wormhole-connect/src/routes/abstracts/routeAbstract.ts
@@ -129,7 +129,11 @@ export abstract class RouteAbstract {
   /**
    * These operations have to be implemented in subclasses.
    */
-  public abstract getMinSendAmount(routeOptions: any): number;
+  public abstract getMinSendAmount(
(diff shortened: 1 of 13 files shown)


kev1n-peters:
Addresses https://github.com/wormhole-foundation/wormhole-connect/issues/1875

artursapek (approved):
changes look fine, my only complaint is the use of the word `recipient`. we already use `toChain`, `destChain`, and `destinationChain` in the connect code. can we not add another one? :D

kev1n-peters:
> changes look fine, my only complaint is the use of the word `recipient`. we already use `toChain`, `destChain`, and `destinationChain` in the connect code. can we not add another one? :D

Agree, I was just being consistent with the other functions in the AbstractRoute class. I like the use of `destChain` and `destToken`

PR #1893, opened 2024-04-05
ntt: make automatic route appear first

kev1n-peters:


kev1n-peters:
Need to think through this more holistically rather than adding a special case for NTT

PR #1913, opened 2024-04-10
Automatically connect to previously used wallet for a given chain context

### LINKED ISSUE #1800: UX: Connect should auto-connect to the wallet you last used for a chain
There's no reason we can't remember this if you've connected once already

artursapek:
Addresses #1800

This change moves all the wallet connection logic out of `WalletModal` (☺️) as well as using a lil' localStorage magic to automatically do this when we select a chain, if the user has done it before. 

https://github.com/wormhole-foundation/wormhole-connect/assets/897596/9cedf880-cf99-464a-9773-4a6776d9714b

TODO

- [x] cover the case where there is a default chain in `bridgeDefaults`

### DIFF
--- a/wormhole-connect/src/utils/wallet/index.ts
+++ b/wormhole-connect/src/utils/wallet/index.ts
@@ -13,11 +13,17 @@ import {
   Wallet,
   WalletState,
 } from '@xlabs-libs/wallet-aggregator-core';
+import {
+  connectReceivingWallet,
+  connectWallet as connectSourceWallet,
+  clearWallet,
+} from 'store/wallet';
 
 import config from 'config';
 import { getChainByChainId } from 'utils';
 
 import { AssetInfo } from './evm';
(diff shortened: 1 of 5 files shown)


yuli-ferna (dismissed):

wormhole-connect/src/utils/wallet/index.ts
```diff
@@ -44,6 +53,92 @@ export const setWalletConnection = (type: TransferWallet, wallet: Wallet) => {
...
+  walletInfo: WalletData,
+  dispatch: Dispatch<any>,
+) => {
+  const { wallet, name } = walletInfo;
+
+  setWalletConnection(type, wallet);
+
+  const chainConfig = config.chains[chain]!;
+  if (!chainConfig) {
+    throw new Error('');
```
kev1n-peters: Should we provide an error message here?

kev1n-peters (approved):

PR #1927, opened 2024-04-12
TokensModal only show native token when possible

kev1n-peters:
Prevent users from receiving wrapped tokens when they could receive the  native version.

### DIFF
--- a/wormhole-connect/src/views/Bridge/Bridge.tsx
+++ b/wormhole-connect/src/views/Bridge/Bridge.tsx
@@ -158,11 +158,17 @@ function Bridge() {
 
   useEffect(() => {
     const computeDestTokens = async () => {
-      const supported = await RouteOperator.allSupportedDestTokens(
+      let supported = await RouteOperator.allSupportedDestTokens(
         config.tokens[token],
         fromChain,
         toChain,
       );
+      // If any of the tokens are native to the chain, only select those.
+      // This is to avoid users inadvertently receiving wrapped versions of the token.
+      const nativeTokens = supported.filter((t) => t.nativeChain === toChain);
+      if (nativeTokens.length > 0) {
+        supported = nativeTokens;
+      }
       dispatch(setSupportedDestTokens(supported));
       const allSupported = await RouteOperator.allSupportedDestTokens(
         undefined,


sebastianscatularo (dismissed):

wormhole-connect/src/components/TokensModal.tsx
```diff
@@ -511,6 +511,16 @@ function TokensModal(props: Props) {
       const isNonzeroBalance = b !== null && b !== '0';
       return isNonzeroBalance;
     });
+
+    // If a destination token is native, display only that token.
+    // This is to avoid users inadvertently receiving wrapped versions of the token.
+    if (type === 'dest') {
+      const nativeToken = supportedTokens.find((t) => t.nativeChain === chain);
+      if (nativeToken) {
+        filtered = [nativeToken];
```
artursapek: is it possible there's multiple native tokens eventually? like if Connect can perform swaps?

this logic is basically "find the first native token and if it exists, only offer that." we could also have it be "find all the native tokens, if any exist, offer them all."

kev1n-peters: I think we can change it to be "future proof"

kev1n-peters: changed

wormhole-connect/src/components/TokensModal.tsx
```diff
@@ -511,6 +511,16 @@ function TokensModal(props: Props) {
       const isNonzeroBalance = b !== null && b !== '0';
       return isNonzeroBalance;
     });
+
+    // If any of the tokens are native to the chain, only show those.
+    // This is to avoid users inadvertently receiving wrapped versions of the token.
+    if (type === 'dest') {
+      const nativeTokens = filtered.filter((t) => t.nativeChain === chain);
+      if (nativeTokens) {
```
artursapek:
```suggestion
      if (nativeTokens.length > 0) {
```

an empty array is truthy

artursapek (changes requested):

wormhole-connect/src/components/TokensModal.tsx
kev1n-peters: fixed

artursapek (approved):

PR #1953, opened 2024-04-17
SDKV2 Integration - first baby steps

artursapek:
## Changes

- Install SDKv2 as a dependency
- Switch to `vitest` for unit tests (jest compatible interface, but uses the Vite config for import path resolution)
- Implement beginnings of a `converter` between the two SDKs, which can convert between networks, chains, and tokens in both directions.
- Write unit tests for `converter`


wormhole-connect/package.json
```diff
@@ -109,7 +110,7 @@
     "vite-bundle-visualizer": "^0.11.0",
     "vite-plugin-checker": "^0.6.2",
     "vite-plugin-dts": "^3.7.3",
-    "vite-plugin-node-polyfills": "^0.17.0"
+    "@artursapek/vite-plugin-node-polyfills": "0.21.1"
```
artursapek: this is a fork which contains this fix https://github.com/davidmyersdev/vite-plugin-node-polyfills/pull/89/files

wormhole-connect/vite.config.ts
```diff
@@ -37,6 +40,8 @@ const resolve = {
     hooks: path.resolve(__dirname, './src/hooks'),
     consts: path.resolve(__dirname, './src/consts'),
     public: path.resolve(__dirname, './public'),
+    'process/': 'process',
+    'buffer/': 'buffer',
```
artursapek, line 44: this is required to un-break one of our downstream dependencies that imports these with a trailing slash...

wormhole-connect/src/config/converter.ts
```diff
@@ -0,0 +1,106 @@
...
+    const isNative = tokenId.address === 'native';
+    const chain = this.toChainNameV1(tokenId.chain);
+
+    for (const key in this.tokens) {
+      const token = this.tokens[key];
+      if (token.nativeChain === chain) {
+        if (isNative && token.tokenId === undefined) {
+          // Connect's TokenConfig lacks a tokenId field when it's the native gas token
+          return token;
+        } else if (token.tokenId?.address === tokenId.address.toString()) {
```
kev1n-peters: Should we do a case insensitive comparison here? EVM token addresses should checksummed (may not be the case for custom tokens), not sure about other platforms (e.g., Sui, Aptos, etc.)

artursapek: I guess it wouldn't hurt to do that here. I prefer to be strict wherever possible but I guess this accepts arbitrary input from integrators so it could be improperly capitalized. will make this change

kev1n-peters (approved):

wormhole-connect/package.json
```diff
@@ -31,6 +31,7 @@
     "@reduxjs/toolkit": "^1.9.1",
     "@solana/wallet-adapter-wallets": "^0.19.25",
     "@solana/web3.js": "^1.73.0",
+    "@wormhole-foundation/sdk": "^0.5.3-beta.6",
     "@wormhole-foundation/sdk-definitions": "^0.5.2-beta.0",
```
barnjamin, line 35: probably should just rm `sdk-definitions` if installing sdk, also means no need to update versions in multiple deps

artursapek: WILL DO IN FOLLOWUP PR

Issue #1950, opened 2024-04-17
[task] - add repair logic VAA to repair it when ever it is possible


sebastianscatularo:
## Description and context

Due the guardian set upgrade, there where created VAAs with guardian set 3 and 4, and the VAAs with old guardian are not working, adding the logic to repair it, would solved this automatically, this feature already exist on Advanced Tools and the old SDK

https://github.com/XLabs/portal-bridge-ui/blob/advanced-tools/src/components/Recovery.tsx#L203
https://github.com/wormhole-foundation/wormhole/blob/8ed75c9bfe1bcecc8d07c80ba38a502a39bcbb94/sdk/js/src/utils/repairVaa.ts#L44

## Definition of done

when a tx with and old guardian set, but enough valid signatures should proceed without issues

## Sample Txs

https://wormholescan.io/#/tx/0x20f85f700d0866962162c8d24ceb73947bd78b6a2d1d6b38bdc10f6a5f1903a5?view=overview

PR #1954, opened 2024-04-17
#1950 - try to repair vaa for vaas with old guardian index set

sebastianscatularo:
Sample Tx without enough signatures to gets the vaa repair 

https://wormholescan.io/#/tx/0x20f85f700d0866962162c8d24ceb73947bd78b6a2d1d6b38bdc10f6a5f1903a5?view=overview

image

### DIFF
--- a/wormhole-connect/src/utils/vaa.ts
+++ b/wormhole-connect/src/utils/vaa.ts
@@ -3,6 +3,7 @@ import { Implementation__factory } from '@certusone/wormhole-sdk/lib/esm/ethers-
 import { utils, providers, BigNumberish } from 'ethers';
 import axios from 'axios';
 import { ChainId, ChainName } from '@wormhole-foundation/wormhole-connect-sdk';
+import { repairVaa as sdkRepairVaa } from '@certusone/wormhole-sdk/lib/esm/utils/repairVaa';
 
 import config from 'config';
 import {
@@ -13,6 +14,7 @@ import {
 } from './sdk';
 
 export type ParsedVaa = {
+  guardianSetIndex: number;
   bytes: string;
   hash: string;
   amount: string;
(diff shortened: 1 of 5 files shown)


wormhole-connect/src/utils/vaa.ts
```diff
@@ -37,6 +39,140 @@ export type MessageIdentifier = {
...
+
+// Create an executor that uses homogeneous function types and handles arbitrary arguments
+const createCustomExecutor =
+  <T, Args extends any[]>(
+    ...funcs: Array<HomogeneousFunction<T, Args>>
+  ): ((...args: Args) => Promise<T | undefined>) =>
+  async (...args: Args): Promise<T | undefined> => {
+    for (let i = 0; i < funcs.length; i++) {
+      const result = await funcs[i](...args);
+      if (result !== undefined) {
```
artursapek, line 95: the only failure case you handle here is an `undefined` return value. is it possible that one of the given functions throws an error? should you try/catch in here so that you continue to the next function in that case?

yuli-ferna (approved):

wormhole-connect/src/views/Redeem/Redeem.tsx
```diff
@@ -90,7 +94,14 @@ function Redeem({
       while (signed === undefined && !cancelled) {
         try {
           signed = await RouteOperator.getSignedMessage(route, txData);
-        } catch {
+        } catch (e: any) {
+          console.trace(e);
```
artursapek, line 98:
```suggestion
```

wormhole-connect/src/views/Redeem/Confirmations.tsx
```diff
@@ -89,7 +91,21 @@ function Confirmations(props: Props) {
             {confirmations} / {chainConfig.finalityThreshold} Confirmations
           </>
         ) : route ? (
-          RoutesConfig[route].pendingMessage
+          isInvalidVaa ? (
+            <>
+              There are not enough valid signatures to repair the VAA <br />{' '}
+              Please reach out to the Wormhole support on{' '}
```
kev1n-peters, line 97: Nit: Can we say, Please reach out to Wormhole support on

wormhole-connect/src/views/Redeem/Redeem.tsx
```diff
@@ -90,7 +94,14 @@ function Redeem({
...
+          if (
+            e?.message === 'There are not enough valid signatures to repair.'
```
artursapek, line 100: minor comment but this could break if someone changes the error message slightly where it gets emitted. maybe make it a const?

wormhole-connect/src/views/Redeem/Confirmations.tsx
kev1n-peters: I would advocate for saying something less technical, but this should be an extremely rare occurrence moving forward, so it's probably okay

wormhole-connect/src/utils/vaa.ts
```diff
@@ -37,6 +39,140 @@ export type MessageIdentifier = {
...
+  }
+}
+
+type VaaParseInput = {
+  vaa: Uint8Array;
+  txData: ParsedMessage | ParsedRelayerMessage;
+};
+
+function parseVaa({ vaa, txData }: VaaParseInput) {
+  const parsed = parseTokenTransferVaa(vaa);
```
kev1n-peters, line 152: This might not be a token transfer vaa in the case of the portico, ntt, and tbtc route. Can we just parse the vaa header which includes the guardian set index? There should be a function for that

wormhole-connect/src/utils/vaa.ts
artursapek: do we need this generics code in here? it seems convoluted and overkill tbh.

artursapek (changes requested):
I would like to see if we can make this change without adding that generic code

wormhole-connect/src/utils/vaa.ts
sebastianscatularo: I didn't this logic, I've just moved, it was strage to me as well

sebastianscatularo:
> I would like to see if we can make this change without adding that generic code

we can, I'm just alway trying to improve the base code, if not, we ends with a lot of duplicated and boilerplate code that start getting expesive to maintain, I'll create a minimal version of it with what is just needed, closing the PR for now

wormhole-connect/src/utils/vaa.ts
sebastianscatularo: it is just an high order function, the argument function should be in charge of ensure the return value is undefined, and it is scoped to this file, so, I do not expect this function to be used elsewhere

wormhole-connect/src/views/Redeem/Confirmations.tsx
sebastianscatularo: Sure, I've just copied from an existing message at the repo

PR #1960, opened 2024-04-18
netlify demo app lets you use custom config

artursapek:
this changes the local demo app (and netlify deploy). it adds a textarea where we can provide arbitrary configs. this will be useful for testing NTT tokens and other features.

https://github.com/wormhole-foundation/wormhole-connect/assets/897596/1c0e8205-75ab-4cfd-a031-486d5e4233d7


wormhole-connect/src/demo.tsx
```diff
@@ -1,13 +1,48 @@
...
+    setShowConnect(false);
+    await sleep(10);
+    console.log(config);
+    setCustomConfig(config);
+    setCustomConfigNonce(customConfigNonce + 1);
+    await sleep(10);
+    setShowConnect(true);
+  };
+
+  console.log(customConfig);
```
sebastianscatularo: is it necessary?

wormhole-connect/src/components/DemoAppHeader.tsx
```diff
@@ -0,0 +1,67 @@
...
+    localStorage.getItem(LOCAL_STORAGE_KEY) ?? '',
+  );
+
+  const updateCustomConfig = (e: any) => {
+    const input = e.target.value;
+    setCustomConfigInput(input);
+  };
+
+  const emitCustomConfig = () => {
+    localStorage.setItem(LOCAL_STORAGE_KEY, customConfigInput);
```
sebastianscatularo: would be nice if also we can load custom config form a query param

sebastianscatularo (dismissed):

wormhole-connect/src/components/DemoAppHeader.tsx
artursapek: oh cool idea. let me see if I can do that

artursapek: done. great idea man

wormhole-connect/package.json
```diff
@@ -53,7 +53,7 @@
     "web-vitals": "^2.1.4"
   },
   "scripts": {
-    "start": "REACT_APP_CONNECT_ENV=testnet vite",
+    "start": "NODE_OPTIONS=--max-http-header-size=100000 REACT_APP_CONNECT_ENV=testnet vite",
```
artursapek: this just lets us have really long URLs in development so we can permalink large configs (I think I'll have to do the same on Netlify)

this is a security thing, but for local dev server & internal Netlify it should be OK
https://vitejs.dev/guide/troubleshooting.html#_431-request-header-fields-too-large

wormhole-connect/src/config/utils.ts
```diff
@@ -144,7 +143,11 @@ export const mergeNttGroups = (
   return builtin;
 };
 
-export const validateDefaults = (defaults: BridgeDefaults | undefined) => {
+export const validateDefaults = (
+  defaults: BridgeDefaults,
+  chains: ChainsConfig,
+  tokens: TokensConfig,
```
artursapek, line 149: this solves a bug: when switching networks `config.chains` is still the chains from the previous network.

wormhole-connect/src/components/DemoAppHeader/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const input = e.target.value;
+    setCustomConfigInput(input);
+  };
+
+  const emitCustomConfig = () => {
+    localStorage.setItem(LOCAL_STORAGE_KEY, customConfigInput);
+    setUrlQueryParam(customConfigInput);
+
+    try {
+      const parsed = eval(`(function() {
```
artursapek: I'm parsing the config as Javascript code, not merely JSON, so we can put in callback functions and stuff. I figure it's safe to do this for local dev server & internal Netlify builds. but it would technically be safer to only accept JSON and use `JSON.parse` here instead.

artursapek: nvm I got scared and changed this to vanilla JSON.parse

kev1n-peters:
I set a bad config and the error boundary got hit which just renders "unexpected error occurred". I had to remove the bad config from the store before I could get Connect working again. Is this something we should care about?

artursapek:
> I set a bad config and the error boundary got hit which just renders "unexpected error occurred". I had to remove the bad config from the store before I could get Connect working again. Is this something we should care about?

i don't think so, as long as deleting the bad config from the textarea and blurring correctly fixed the Connect widget. all it's doing is showing you what output you get for a given input.

wormhole-connect/src/components/DemoApp/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { WormholeConnectConfig } from 'config/types';
+import { compressToBase64, decompressFromBase64 } from 'lz-string';
+
+const MAX_URL_SIZE = 30_000; // 30kb (HTTP header limit is set to 32kb)
+
+const parseConfig = (config: string): WormholeConnectConfig => {
+  if (config) {
+    try {
+      return eval(
+        `(function() { return ${config} })()`,
```
kev1n-peters, line 14: Are we calling `eval` here to allow javascript in the config (e.g. a `WormholeConnectEventHandler` function?)

kev1n-peters (approved):

wormhole-connect/src/components/DemoApp/index.tsx
artursapek: that, and also just to allow us to copy & paste WIP source code in here, which is usually not valid JSON with string keys. eg `{ nttGroups: {...} }` vs `{ "nttGroups": {...} }`

artursapek: JSON also screams about trailing commas, etc. it's just annoyingly strict.

PR #1961, opened 2024-04-18
#1950 - try to repair vaa for vaas with old guardian index set

sebastianscatularo:
tested with 

- https://wormholescan.io/#/tx/0x7de3aa3afcf212553d29396d1ce91caac064230bde88eb1927cbdd270e782397?view=advanced
- https://wormholescan.io/#/tx/0x9e056c31848f998219528a63cfbf9a7e199c1d8ae9a316d3d41007e7ba321cc9?view=advanced

image


wormhole-connect/src/utils/vaa.ts
```diff
@@ -11,6 +15,61 @@ import {
...
 } from './sdk';
+import { repairVaa } from './repairVaa';
+
+type GuardianSetData = {
+  index: number;
+  keys: string[];
+  expiry: number;
+};
+
+const GUARDIAN_SETS: { [key: string]: GuardianSetData } = {
```
artursapek: `key` should be `Network`, right? there's a type for that.

also, we tend to keep network-specific constants like this here:

https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/config/mainnet/

for example, built-in tokens, RPCs, etc all live in there. can you move this const to there?

artursapek (changes requested):

wormhole-connect/src/utils/repairVaa.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Takes in a hexstring representation of a signed vaa and a guardian set.
+ * Attempts to remove invalid guardian signatures, update total remaining
+ * valid signatures, and update the guardian set index
+ * @throws if not enough valid signatures remain
+ * @see https://github.com/wormhole-foundation/wormhole/blob/main/sdk/js/src/utils/repairVaa.ts#L44
+ *
+ * NOTE: copied since original function does not normalize the public keys
+ **/
+export function repairVaa(
+  vaaHex: string,
```
kev1n-peters: Can this take in a `Uint8Array` instead of a hex string? We already call `parseVaa` before calling this function (to check the guardian set index), so calling `parseVaa` again in this function is extra work that can be avoided.

wormhole-connect/src/consts/guardianSet.ts
```diff
@@ -0,0 +1,44 @@
...
+    expiry: 0,
+  },
+  testnet: {
+    index: 0,
+    keys: ['0x13947Bd48b18E53fdAeEe77F3473391aC727C638'],
+    expiry: 0,
+  },
+};
+
+export const GUARDIAN_SET = GUARDIAN_SETS[config.network]!;
```
artursapek: actually this isn't good. `config.network` can change at runtime now, and in fact does once when Connect first binds to the DOM. this will not work right.

https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/config/index.ts#L187

just export the entire object and let the consumer decide which network to use when it's needed at runtime.

artursapek (changes requested):

artursapek (approved):

PR #1982, opened 2024-04-22
Token balance display fix

kev1n-peters:
Previously the token balances would only be shown after the TokensModal was opened to fetch and set the wallet's balances. The useGetTokenBalance hook is now used to fetch the balances on demand (if not already cached).


kev1n-peters:
Do we want to update `balances` in the store after a cache miss and successful balance fetch in the hook?

wormhole-connect/src/views/Bridge/Inputs/From.tsx
```diff
@@ -44,15 +44,13 @@ function FromInputs() {
...
     toChain,
-    balances,
     token,
     amount,
     isTransactionInProgress,
     destToken,
   } = useSelector((state: RootState) => state.transferInput);
   const tokenConfig = token && config.tokens[token];
-  const balance =
-    accessBalance(balances, wallet.address, fromChain, token) || undefined;
```
kev1n-peters, line 55: For context: `balances` is set in `TokensModal`, so previously `balance` would always be `undefined` until the `TokensModal` was opened. This can be problematic when bridge defaults are set or when the destination token is automatically set

artursapek (commented):
thanks for fixing this. my only reservation is we now have somewhat duplicated logic between the [TokenModal](https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/components/TokensModal.tsx#L383-L450) and now this new thing. is there a way to consolidate that?

what if this new `useGetTokenBalance` took a list of token symbols strings as an argument, and return an object that maps the symbols to formatted balances? would that satisfy TokenModal's needs and let us remove that older code?

artursapek (approved):
👏

PR #2015, opened 2024-04-29
Fixed destructure property error when resuming transaction

kev1n-peters:
`transferInput` will not have any values set when going through the resume transfer flow, so we just use the properties from `txData` instead


artursapek (dismissed):

artursapek:
ty

artursapek (approved):
beauty

PR #2036, opened 2024-05-03
Added validateTransfer handler to config

kev1n-peters:
Addresses #2037


wormhole-connect/src/config/types.ts
```diff
@@ -75,6 +75,20 @@ export interface BridgeDefaults {
   requiredNetwork?: ChainName;
 }
 
+export interface ExtendedTransferDetails extends TransferDetails {
+  fromWallet: string;
```
artursapek: we use "wallet" to refer to the extension used (phantom, etc) in a different telemetry event. so can we use `walletAddress` here?

artursapek (approved):

PR #2038, opened 2024-05-06
Fix: Resume transaction always trigger 'transfer.success' event

yuli-ferna:
Due every time a transaction is resumed, I decided to add a new variable to the store to avoid this behavior.

### DIFF
--- a/wormhole-connect/src/store/redeem.ts
+++ b/wormhole-connect/src/store/redeem.ts
@@ -20,6 +20,7 @@ export interface RedeemState {
   route: Route | undefined;
   transferDestInfo: TransferDestInfo | undefined;
   deliveryStatus: DeliveryStatus | undefined;
+  isResumeTx: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


artursapek:
do we need to set it back to `false` after it's done, in case they go on to do a fresh transfer from the beginning?

yuli-ferna:
> do we need to set it back to `false` after it's done, in case they go on to do a fresh transfer from the beginning?

@artursapek  In this case I do not think it is necessary, because when I when I leave the resume tx section, the function `clearRedeem()`, that set in `false` the var `isResumeTx`, is executed.

```
  useEffect(() => {
    const redeemRoute = 'redeem';
    const bridgeRoute = 'bridge';
    // reset redeem state on leave
      dispatch(clearRedeem());
      ...
    }
    ...
  }, [route, prevRoute, dispatch]);
```

artursapek (approved):

PR #2061, opened 2024-05-08
bump polyfills package to remove buffer patch

kev1n-peters:
includes https://github.com/kev1n-peters/vite-plugin-node-polyfills/commit/be6f6c8c60b9edde427eac38b2941f8532b9858c

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4599,6 +4599,24 @@
         "node": ">= 6"
       }
     },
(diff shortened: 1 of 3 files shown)


kev1n-peters:
Use a forked version of the vite-plugin-node-polyfills package that removes the patch to buffer@0.6.3. The patch was causing dev builds to throw a runtime error when referencing the `globalThis` variable before its defined.

artursapek (approved):

PR #2063, opened 2024-05-08
Disable ETH (Ethereum) > Scroll transfers (frankenstein ETH)

kev1n-peters:


artursapek (approved):
I know we already established this pattern but I think ideally this logic should be implemented in the `Route` implementation instead, eventually. `isRouteSupported` should just return `false` for all these cases.

PR #2085, opened 2024-05-15
Added a message to show the user the wallet address has been copied

CodeQueenTosin:
Added a message to show the user the wallet address has been copied.


aadam-10:
@CodeQueenTosin hello, thanks for submitting but we're working on an UI overhaul and would not like to extend the current one. the current behavior of the dropdown immediately disappearing after clicking any button also makes this new text behavior strange, as it flashes for a moment and disappears.

PR #2090, opened 2024-05-15
SDKv2 migration: Route adapter

artursapek:
This change introduces a class `SDKv2Route` which implements `RouteAbstract` by wrapping a [`routes.Route`](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/connect/src/routes/route.ts) from SDKv2.

The first milestone I'd like to merge implements a subset of the required methods:

- `isSupportedChain`
- `isSupportedSourceToken`
- `isSupportedDestToken`
- `supportedSourceTokens`
- `supportedDestTokens`
- `computeReceiveAmount`
- `computeReceiveAmountWithFees` (I think this needs tweaking)

this allows us to actually view quotes in Connect using SDKv2. to test this one needs to set a flag in `localStorage` and refresh the page:

```
localStorage.setItem('CONNECT_SDKV2', 1);
```


package.json
```diff
@@ -29,5 +29,27 @@
...
+      "@ledgerhq/errors": "6.10.2",
+      "@ledgerhq/logs": "6.12.0",
+      "@ledgerhq/devices": "6.27.1"
+    },
+    "@wormhole-foundation/sdk-definitions-ntt": {
+      "@wormhole-foundation/sdk-definitions": "^0.7.0-beta.1",
+      "@wormhole-foundation/sdk-base": "^0.7.0-beta.1"
+    },
+    "@wormhole-foundation/wormhole-connect": {
+      "aptos": "1.5.0"
```
kev1n-peters, line 52: Do we need this override since `wormhole-connect/package.json` already has this version?

package.json
```diff
@@ -29,5 +29,27 @@
...
+      "@ledgerhq/errors": "6.10.2",
+      "@ledgerhq/logs": "6.12.0",
+      "@ledgerhq/devices": "6.27.1"
+    },
+    "@ledgerhq/hw-transport-webhhid@6.27.1": {
+      "@ledgerhq/errors": "6.10.2",
+      "@ledgerhq/logs": "6.12.0",
+      "@ledgerhq/devices": "6.27.1"
+    },
+    "@wormhole-foundation/sdk-definitions-ntt": {
```
kev1n-peters, line 47: Should probably have QA do a few NTT transfers just to make sure that nothing broke (I wouldn't think so, but I guess to make sure)

wormhole-connect/src/config/converter.ts
```diff
@@ -106,4 +111,28 @@ export class SDKConverter {
       }
     }
   }
+
+  getTokenIdV2ForSymbol<C extends v2.Chain>(
```
kev1n-peters: Minor nit: Should we rename to `getTokenIdV2ForKey` or something since the `key` is what uniquely identifies a token (`symbol` is a different property)

wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -0,0 +1,514 @@
...
+  ): Promise<any> {
+    throw new Error('Method not implemented.');
+  }
+
+  public redeem(
+    destChain: ChainName | ChainId,
+    messageInfo: SignedMessage,
+    recipient: string,
+  ): Promise<string> {
+    throw new Error('Method not implemented.');
```
kev1n-peters, line 441: Will these methods be implemented in a future PR?

kev1n-peters: Ah, that's mentioned in the PR description.

wormhole-connect/src/routes/operator.ts
```diff
@@ -284,19 +259,44 @@ export class Operator {
...
-          const isSupported = await this.isSupportedDestToken(
-            route as Route,
-            config.tokens[key],
-            sourceToken,
-            sourceChain,
-            destChain,
-          );
-          if (isSupported) {
-            supported[key] = config.tokens[key];
+      if (!config.routes.includes(route)) {
```
kev1n-peters: I think we can remove the `if` statement here

package.json
artursapek: yes because [this version](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/platforms/aptos/package.json#L50) in SDKv2 clashes and NPM is not good at figuring out what to do unless I explicitly define it here

wormhole-connect/src/routes/operator.ts
artursapek: why?

wormhole-connect/src/config/converter.ts
artursapek: done

wormhole-connect/src/routes/operator.ts
kev1n-peters: The `for...of` loop already iterates over each `route` in `config.routes` unless I'm missing something

artursapek: ah right. thanks.

kev1n-peters (approved):

Issue #2112, opened 2024-05-22
Allow configuration of priority fee settings


barnjamin:
Allow an integrator to configure the priority fee settings, currently hardcoded

https://github.com/wormhole-foundation/wormhole-connect/blob/8063cc2f159459b22346f0dff01e9d1b4cce12bb/sdk/src/contexts/solana/utils/computeBudget/index.ts#L39

PR #2128, opened 2024-05-23
#2112 - add option to configure priority fees

sebastianscatularo:
- expose option to set priority fees percentile
- expose function to compute priority fees with an alternative mechanism

### DIFF
--- a/sdk/src/contexts/solana/utils/computeBudget/index.ts
+++ b/sdk/src/contexts/solana/utils/computeBudget/index.ts
@@ -7,7 +7,7 @@ import {
 } from '@solana/web3.js';
 
 // Add priority fee according to 50th percentile of recent fees paid
-const DEFAULT_FEE_PERCENTILE = 0.5;
+export const DEFAULT_FEE_PERCENTILE = 0.5;
 
 export async function addComputeBudget(
   connection: Connection,
@@ -15,6 +15,9 @@ export async function addComputeBudget(
   lockedWritableAccounts: PublicKey[] = [],
   feePercentile: number = DEFAULT_FEE_PERCENTILE,
   minPriorityFee: number = 0,
+  customDeterminePriorityFee?: (
+    lockedWritableAccounts: PublicKey[],
+  ) => Promise<number>,
 ): Promise<void> {
   if (lockedWritableAccounts.length === 0) {
     lockedWritableAccounts = transaction.instructions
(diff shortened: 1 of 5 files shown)


sdk/src/contexts/solana/utils/computeBudget/index.ts
```diff
@@ -89,14 +97,20 @@ export async function determineComputeBudget(
   ];
 }
 
+const DEFAULT_PRIORITY_FEE = (_: PublicKey[]) => Promise.resolve(1); // Set fee to 1 microlamport by default
+
 async function determinePriorityFee(
   connection: Connection,
   lockedWritableAccounts: PublicKey[] = [],
   percentile: number,
+  customDeterminePriorityFee: (
```
artursapek, line 106: I don't see how it would be useful to an integrator because we're not providing the `connection` to this custom function, only the addresses. they can't do much with only addresses unless they're providing their own connection.

also, the way you're using it here is you're just setting the default value using it, rather than the final value. it will still get overridden by our own logic, which follows it in this function starting on line 115. so this doesn't seem like it will even work.

It seems to me we want to either:
- let integrators completely override the `determinePriorityFee` function
- or let them just provide a custom percentile and keep our logic the same
 
this code change is sort of a weird combination of the two.

I would keep things simple and go with option 2 for now. can we just remove this custom callback function, and only add the percentile `number` as a new config option?

if there's need for deeper customizability in the future we can revisit overriding `determineComputeBudget` later.

artursapek (changes requested):

sdk/src/contexts/solana/utils/computeBudget/index.ts
germor197: So this solution wont' work? I asked @barnjamin about this functionality but i'm having problems trying to clone this commit and installing it to my react app so i couldn't test it, and as i see it didn't merged to development brach yet right?

sebastianscatularo: good call. I’ll add it as a parameter. The idea is to allow integrators to use a different API, like what Helius provides for estimating priority fees, or any custom solution that someone might want to try, the limit is the skies 😅

artursapek: I think it would be good to offer both: custom percentile, or custom `determinePriorityFee` function altogether.

PR #2131, opened 2024-05-24
Skip custom tokens that define foreign assets

kev1n-peters:
Incorrect foreign assets can result in issues such as an incorrect ATA being generated on Solana.


emreboga (dismissed):
LGTM!

emreboga (approved):

PR #2137, opened 2024-05-28
Removing unnecessary SDK calls during initial load

emreboga:
During initial load of the Connect bridge, we are making unnecessary calls to fetch supported source and destination tokens, even before user selects those networks. In this PR I'm skipping those calls if the networks not selected yet.

Please see the loom for more details:
https://www.loom.com/share/231cc074f3e742b38b06ce13dd75de67?sid=17ea8b52-a2d9-4e68-8dc2-6f253a80d657

### DIFF
--- a/wormhole-connect/src/routes/sdkv2/route.ts
+++ b/wormhole-connect/src/routes/sdkv2/route.ts
@@ -101,10 +101,10 @@ export class SDKv2Route<N extends Network> extends RouteAbstract {
 
     return routes.RouteTransferRequest.create(
       wh,
+      /* @ts-ignore */
       {
         source: srcTokenV2,
         destination: dstTokenV2,
-        /* @ts-ignore */
       },
       srcChain,
       dstChain,
--- a/wormhole-connect/src/views/Bridge/Bridge.tsx
+++ b/wormhole-connect/src/views/Bridge/Bridge.tsx
@@ -115,8 +115,12 @@ function Bridge() {
 
   // check destination native balance
   useEffect(() => {
-    if (!fromChain || !toChain || !receiving.address) return;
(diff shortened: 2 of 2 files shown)


wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -101,10 +101,10 @@ export class SDKv2Route<N extends Network> extends RouteAbstract {
 
     return routes.RouteTransferRequest.create(
       wh,
+      /* @ts-ignore */
```
emreboga, line 104: Moved it up here for the ignore directive to work properly.

wormhole-connect/src/views/Bridge/Bridge.tsx
```diff
@@ -231,21 +246,29 @@ function Bridge() {
...
     computeDestTokens();
+
     return () => {
       canceled = true;
     };
     // IMPORTANT: do not include destToken in dependency array
   }, [route, token, fromChain, toChain, dispatch]);
 
   useEffect(() => {
+    if (!route || !amount || !token || !destToken || !fromChain || !toChain) {
```
emreboga, line 259: @artursapek I assume if one of these are missing, we don't need to calculate the receive amount, right?

wormhole-connect/src/views/Bridge/Bridge.tsx
```diff
@@ -231,21 +246,29 @@ function Bridge() {
...
+      return;
+    }
+
     const recomputeReceive = async () => {
       if (!route) return;
```
artursapek: I think you could remove this now.

artursapek (approved):

PR #2148, opened 2024-05-29
Fix Route resolution issue

artursapek:
This fixes route resolution by adding some missing parameters to old implementation of `supportedSourceTokens` and `supportedDestTokens` in `baseRoute.ts`.


sebastianscatularo (approved):
LGTM 🚀

PR #2149, opened 2024-05-30
isRouteSupported unit test suite

artursapek:
This adds unit tests which cover the very basic `isRouteSupported` check for inputs we expect to be supported for each route. This will catch any changes which cause an entire route to stop being recognized by Connect.

### DIFF
--- a/wormhole-connect/tests/ci/routes.test.ts
+++ b/wormhole-connect/tests/ci/routes.test.ts
@@ -28,6 +28,8 @@ describe('supported routes', () => {
     // ETH bridge
     ['ETH', 'WETHbsc', 'ethereum', 'bsc', [Route.ETHBridge]],
     ['ETH', 'WETHpolygon', 'ethereum', 'polygon', [Route.ETHBridge]],
+    ['WETHarbitrum', 'WETHpolygon', 'arbitrum', 'polygon', [Route.ETHBridge]],
+    ['WETHpolygon', 'WETHarbitrum', 'polygon', 'arbitrum', [Route.ETHBridge]],
     // wstETHBridge
     ['wstETH', 'wstETHpolygon', 'ethereum', 'polygon', [Route.wstETHBridge]],
     ['wstETH', 'wstETHarbitrum', 'ethereum', 'arbitrum', [Route.wstETHBridge]],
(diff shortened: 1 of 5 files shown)


wormhole-connect/tests/ci/routes.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        const isSupported = await r.isRouteSupported(
+          sourceToken,
+          destToken,
+          '1.0',
+          sourceChain,
+          destChain,
+        );
+
+        expect(isSupported).toBeTruthy();
```
emreboga, line 91: @artursapek Should we check any fail scenarios as well?

wormhole-connect/tests/ci/routes.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  for (let [
+    sourceToken,
+    destToken,
+    sourceChain,
+    destChain,
+    routes,
+  ] of testCases) {
+    for (let route of routes) {
+      test(`${route} ${sourceChain}:${sourceToken} -> ${destChain}:${destToken}`, async () => {
+        const r = getRouteImpls(route).v1;
```
emreboga, line 81: Should we v2s as well wherever available? By the way I'm not 100% sure of the difference here, but still wanted to point it out.

sebastianscatularo (approved):

wormhole-connect/tests/ci/routes.test.ts
artursapek: I actually have a different test file I'm working on that compares v1 to v2, for the purposes of the migration:

https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/tests/ci/sdkv2.test.ts

however, the migration is incomplete so those tests are also incomplete. as I add support for stuff, I am trying to add coverage for it.

wormhole-connect/tests/ci/routes.test.ts
artursapek: I'm trying to keep this test file focused on ensuring we don't break route resolution again for the legacy code. this was broken earlier this week by me in the SDKv2 migration.

I had started writing a more comprehensive test suite that tests for both `true` and `false` results by comparing v1 to v2 routes, but this runs so many tests that it gets rate limited by even paid RPC nodes: https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/tests/manual/sdkv2.comparisons.ts#L28

so I agree we should test for the negative as well. but in this test file, I'd like to just keep it simple and avoid regressions in the legacy routes.

emreboga (approved):
LGTM!

PR #2153, opened 2024-05-30
solana: Optimized TXs / client-side retry logic

kev1n-peters:
Only works for posting the VAA currently

### DIFF
--- a/sdk/src/contexts/solana/utils/utils/transaction.ts
+++ b/sdk/src/contexts/solana/utils/utils/transaction.ts
@@ -4,11 +4,11 @@ import {
   Connection,
   PublicKeyInitData,
   PublicKey,
-  ConfirmOptions,
   RpcResponseAndContext,
   SignatureResult,
   TransactionSignature,
   Signer,
+  Commitment,
 } from '@solana/web3.js';
 
 /**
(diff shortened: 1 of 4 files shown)


sdk/src/contexts/solana/utils/utils/transaction.ts
```diff
@@ -206,3 +142,71 @@ export async function sendAndConfirmTransactionsWithRetry(
...
+  confirmTransactionPromise = connection.confirmTransaction(
+    {
+      signature,
+      blockhash,
+      lastValidBlockHeight,
+    },
+    commitment,
+  );
+  // Keep checking for confirmation until the promise resolves or the timeout is reached
+  while (!confirmedTx) {
```
artursapek, line 190: this will retry indefinitely. do you think we should still give up after some amount of time, to avoid an infinite loading spinner? for example what if there's a network outage? 

also, what if the transaction throws an error? wouldn't we be spending the user's gas on infinite retries?

sdk/src/contexts/solana/utils/utils/transaction.ts
```diff
@@ -206,3 +142,71 @@ export async function sendAndConfirmTransactionsWithRetry(
...
+    if (confirmedTx) {
+      break;
+    }
+    console.log(
+      `Tx not confirmed after ${
+        txRetryInterval * txSendAttempts++
+      }ms, resending`,
+    );
+    // Resend the transaction if it hasn't been confirmed
+    await connection.sendRawTransaction(serializedTx, sendOptions);
```
artursapek: don't you need to get a fresh blockhash before doing this?

sdk/src/contexts/solana/utils/utils/transaction.ts
kev1n-peters: This will retry until the call to `confirmTransaction` throws, e.g. if the tx isn't confirmed after the block height expires (`TransactionExpiredBlockheightExceededError`)

Transaction errors should be caught during simulation. The TX isn't resigned, we just broadcast the same signed TX repeatedly until it's confirmed or not

sdk/src/contexts/solana/utils/utils/transaction.ts
kev1n-peters: I don't think it matters since we definitely don't want to change the `lastValidBlockHeight` - we want to rebroadcast the same signed TX until it's confirmed or not. Triton One also doesn't do this in the example code https://github.com/rpcpool/optimized-txs-examples/blob/main/selfSolTransfer.mjs

sdk/src/contexts/solana/utils/utils/transaction.ts
tonyjin: I see this follows Triton's example here: https://github.com/rpcpool/optimized-txs-examples/blob/main/jupiterSwap.mjs#L147

The `while`/`break` behavior + Promises created in a loop make me a bit nervous, would be good to double check this can't result in an infinite loop or unsettled promises

kev1n-peters: The `while` loop won't result in an infinite loop because it breaks when the tx is confirmed. If the tx is not confirmed because the block height is exceeded, then `confirmTransactionPromise` will throw an exception and break the loop.

artursapek (approved):

PR #2155, opened 2024-05-30
Removing multiple setConfig calls

emreboga:
In AppRouter we are calling `setConfig` multiple times, which creates WH Context multiple times and causes unnecessary RPC calls to register providers. The root cause of this is the condition `if (props.config)` which evals to `true` for empty objects. I've added lodash lib and used isEmpty method for the condition.
Using lodash for a single case seems like an overkill, I'm guessing there will be other instances lodash lib can become very handy. Please LMK if you think otherwise.

Number of fetch calls in local dev (including fail and retries):
Before: 135 requests 
After: 69 requests


artursapek:
fair enough, but can we please not use lodash for this? I've been trying to keep reducing the bundle size since I got here and this would add a big dependency for a check that could just be inlined without adding a new dependency.

image

artursapek (approved):

PR #2190, opened 2024-06-13
Added Mantle mainnet and testnet networks

kev1n-peters:


artursapek:
are we also adding these new chains to SDKv2 at the same time? @barnjamin

kev1n-peters:
> are we also adding these new chains to SDKv2 at the same time? @barnjamin

Yeah, testnet was already there, added mainnet things here: https://github.com/wormhole-foundation/wormhole-sdk-ts/pull/586

PR #2197, opened 2024-06-14
Add mantle

kev1n-peters:


artursapek (approved):

wormhole-connect/tests/ci/sdkv2.test.ts
```diff
@@ -213,35 +213,35 @@ describe('token', () => {
...
-          expect(isSupportedV1).toEqual(isSupportedV2);
-        });
-      }
-    }
-  };
-
-  compareChains('mainnet');
-  //compareChains('testnet');
-});
+//describe('compare isSupportedChain between v1 and v2 routes', () => {
```
emreboga: @kev1n-peters are going to need this test in the future? Can we remove it or revisit to re-enable it later on?

kev1n-peters: I think we could leave the test but add a case that skips the 'mantle' chain for now. The current version of sdkv2 installed doesn't support mantle and bumping the version to latest (that supports mantle) causes all sorts of dependency and build issues that aren't worth solving in this PR, especially since the ongoing migration to use sdkv2 is going to be changing these dependencies.

emreboga (approved):
LGTM!

PR #2198, opened 2024-06-14
NTT display name changes

kev1n-peters:


wormhole-connect/src/views/Bridge/RouteOptions.tsx
```diff
@@ -281,6 +283,11 @@ function RouteOption(props: { route: RouteData; disabled: boolean }) {
     [route, toChain],
   );
 
+  const displayName = isNttRoute(routeName)
+    ? getNttDisplayName(fromTokenConfig, toTokenConfig) || props.route.name
```
artursapek, line 287: if this wasnt one-off hack only on the release branch I would cringe at putting it here but we good

artursapek (approved):

PR #2229, opened 2024-07-01
remove deprecated goerli testnets & respective tokens

evgeniko:
removed deprecated goerli, arbitrum_goerli, optimism_goerli & base_goerli chains & tokens. Changed testnet token address of tokens that could be clearly identified using etherscan & respective testnet chain explorers


kev1n-peters:
Thanks for the contribution, however this targets the deprecated / legacy version of Connect which will not be maintained (especially for testnet) moving forward.

PR #2230, opened 2024-07-05
Adding working dir change since workspaces were removed

emreboga:
We need to move the current directory to `./wormhole-connect` for the `npx` commands to work. This is probably due to we no longer having workplaces in our repo.

### DIFF
--- a/wormhole-connect/package.json
+++ b/wormhole-connect/package.json
@@ -129,7 +129,7 @@
     "react-dom": "^18.2.0"
   },
   "lint-staged": {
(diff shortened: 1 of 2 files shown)


wormhole-connect/package.json
```diff
@@ -129,7 +129,7 @@
     "react-dom": "^18.2.0"
   },
   "lint-staged": {
-    "**/*": "prettier --workspaces --if-present --write --ignore-unknown"
```
emreboga, line 132: we also do not need these two params, workspaces is no longer and if-present is deprecated.

artursapek (approved):

PR #2243, opened 2024-07-10
Foreign Assets cache, custom tokens support

artursapek:


wormhole-connect/src/utils/sdkv2.ts
```diff
@@ -1,54 +1,41 @@
...
 
-  // First, get token key from origin chain
-  let tokenKey: TokenKey | null = null;
-  for (const tc of Object.values(originChainContext.config.tokenMap || {})) {
-    if (tc.address === token.address.toString()) {
-      tokenKey = tc.key;
-    }
-  }
+  const tokenId = config.sdkConverter.toTokenIdV2(token);
+  const wrapped = await tb.getWrappedAsset(tokenId);
```
kev1n-peters: Should we try catch here in case the wrapped asset is not found and throws?

wormhole-connect/src/config/types.ts
```diff
@@ -363,3 +367,54 @@ export type NttGroup = {
...
+        this.set(key, chainV2, addr);
+      }
+
+      // Cache it on its native chain too
+      if (token.tokenId) {
+        const nativeChainV2 = converter.toChainV2(token.nativeChain);
+        this.set(
+          key,
+          nativeChainV2,
+          WormholeV2.parseAddress(nativeChainV2, token.tokenId!.address),
```
emreboga: Nitpick: Do you need `token.tokenId!` as you already checked `token.tokenId` in the if condition?

wormhole-connect/src/config/types.ts
```diff
@@ -363,3 +367,54 @@ export type NttGroup = {
...
+  caches: Partial<Record<ChainV2, ForeignAssets<ChainV2>>>;
+
+  constructor(tokens: TokensConfig, converter: SDKConverter) {
+    this.caches = {};
+
+    // Pre-populate cache with values from built-in config
+    for (const key in tokens) {
+      const token = tokens[key];
+
+      for (const chain in token.foreignAssets) {
```
emreboga: Can `foreignAssets` be null or undefined?

wormhole-connect/src/config/types.ts
```diff
@@ -363,3 +367,54 @@ export type NttGroup = {
...
+        const foreignAsset = tokens[key].foreignAssets![chain];
+        const chainV2 = converter.toChainV2(chain as ChainName);
+        const addr = WormholeV2.parseAddress(chainV2, foreignAsset.address);
+        this.set(key, chainV2, addr);
+      }
+
+      // Cache it on its native chain too
+      if (token.tokenId) {
+        const nativeChainV2 = converter.toChainV2(token.nativeChain);
```
emreboga: Can there be a potential bug where `nativeChainV2` is returned null or undefined from the converter? You might wanna add a check below before caching it.

wormhole-connect/src/routes/operator.ts
```diff
@@ -369,34 +369,15 @@ export class Operator {
...
       const r = this.getRoute(route as Route);
 
-      try {
-        const sourceTokens = await r.supportedSourceTokens(
-          config.tokensArr,
-          destToken,
-          sourceChain,
-          destChain,
-        );
+      const sourceTokens = await r.supportedSourceTokens(
```
emreboga, line 372: Does removing the try/catch mean we won't get any exceptions from any of the routes for supported source tokens? If not, maybe leave it just logging purposes?

wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -277,13 +206,33 @@ export class SDKv2Route {
...
+    // we can't match the result of supportedDestinationTokens back up to a TokenConfig
+    // to get its name and logo.
+    //
+    // Since token bridge only outputs the same token as you put in, we're just
+    // returning sourceToken here as a hack so that we can maintain the name and logo.
+    //
+    // A longer term solution to this might be to add methods to SDKv2 for fetching token
+    // metadata like name/logo and not relying on configuration for this at all. At that
+    // point all that would be required would be an address.
+    if (this.TYPE === 'bridge' || this.TYPE === 'relay') {
```
kev1n-peters: We should also add cosmosGateway as it uses wrapped assets

wormhole-connect/src/config/types.ts
artursapek: yeah and ts doesnt complain about this. I guess javascript is perfectly ok with it:

image

wormhole-connect/src/config/types.ts
artursapek: it will throw if it gets invalid inputs. if that happens in this case it means the core config has bad data. I can catch here in case that happens.

wormhole-connect/src/routes/operator.ts
artursapek: no. this try/catch was only here for a short time in an attempt to gracefully transition to SDKv2, which returns all supported destination tokens in one method call. the catch case was for legacy interface which wouldn't handle that, and instead need to check each token individually. we got rid of the legacy interface completely so this try/catch isn't required anymore.

wormhole-connect/src/config/types.ts
artursapek: youre right

wormhole-connect/src/routes/sdkv2/route.ts
artursapek: to be honest, I want to get rid of `TYPE` completely. we have to do that, to support plugin routes, which Mayan is supposed to be. but I can do this temporarily.

emreboga (dismissed):
LGTM after latest changes.

wormhole-connect/src/config/converter.ts
```diff
@@ -152,12 +153,14 @@ export class SDKConverter {
...
-        );
-      }
+      // For token bridge route, we might be trying to fetch a token's address on its
+      // non-native chain.
+      const foreignAddress = await getTokenBridgeWrappedTokenAddress(
+        tokenConfigs[key],
+        this.toChainV2(chain),
+      );
+      if (!foreignAddress) return undefined;
+      return this.tokenIdV2(chainName, foreignAddress?.toString());
```
kev1n-peters: Nit: remove the `?`

wormhole-connect/src/config/types.ts
kev1n-peters: I wonder if this will throw a `Uncaught TypeError: undefined is not iterable` error if a token doesn't have `foreignAssets` defined. Maybe we could just do `if (!token.foreignAssets) continue` to play it safe?

artursapek: done

kev1n-peters (approved):

PR #2246, opened 2024-07-11
Remove ethers5 as direct dependency

artursapek:


wormhole-connect/src/hooks/useIsTransferLimited.ts
```diff
@@ -80,7 +80,7 @@ const useIsTransferLimited = (): IsTransferLimitedResult => {
         const tokenConfig = config.tokens[token];
         const tokenId = getWrappedTokenId(tokenConfig);
         const tokenChain = config.wh.toChainId(tokenId.chain);
-        const formatted = hexlify(
+        const formatted = encoding.hex.encode(
           await formatAssetAddress(tokenChain, tokenId.address),
         );
```
barnjamin: does hexlify prepend a `0x`? if so, and we want to keep it identical, add a second argument `true` to make sure its prefixed

barnjamin (dismissed):

wormhole-connect/src/hooks/useIsTransferLimited.ts
artursapek: good question!

artursapek: I solved this by just removing this file completely lol

barnjamin (approved):

PR #2249, opened 2024-07-16
Type: Export WormholeConnectEvent from index

jovinlidan:
Currently, to import WormholeConnectEvent, it is imported from telemetry. Now, it is exported from the index file instead.


wormhole-connect/src/config/types.ts
```diff
@@ -76,6 +76,8 @@ export type SupportedRoutes = keyof typeof Route;
 
 export type Network = 'mainnet' | 'testnet' | 'devnet';
 
+export { WormholeConnectEvent } from 'telemetry/types';
```
artursapek: seems weird to export it from `config/types`, and have index import it from here. why can't index just import it directly from `telemetry/types`?

jovinlidan: oh yes, you right, i just refactor the export code

artursapek (dismissed):

PR #2259, opened 2024-07-17
Unable to send native token via token bridge route

kev1n-peters:
supportedDestTokens should return the wrapped version of the token otherwise the token bridge routes don't support it


artursapek (approved):
lgtm

PR #2264, opened 2024-07-18
[Redesign] Routes list and Review transaction

emreboga:
This PR implements the new UI for the routes list and review transaction step in Bridge view.

Please see the recording here:
https://www.loom.com/share/9b2407b4b8dc411686646a49aab9d504?sid=12dc083e-8fc1-4bff-9768-9482d58120fe


wormhole-connect/src/hooks/useAvailableRoutes.ts
emreboga, line 1: Existing functionality moved into a separate file (from `src/views/Bridge/RouteOptions.tsx`) for reusability and easier to maintain.

wormhole-connect/src/hooks/useComputeDestinationTokensV2.ts
emreboga, line 1: Existing functionality from `useComputeDestinationTokens` but instead of setting the Redux state, it returns the computed values. The reason is to maintain some of these values locally in separate components, instead of dumping them all in the Redux state.

wormhole-connect/src/hooks/useComputeFees.ts
emreboga, line 1: Existing functionality from `src/views/Bridge/RouteOptions.tsx` but instead of setting the Redux state, it returns the computed values. The reason is to get and re-compute these values locally to easily maintain and reuse.

wormhole-connect/src/hooks/useComputeSourceTokensV2.ts
emreboga, line 1: Existing functionality from `useComputeSourceTokens` but instead of setting the Redux state, it returns the computed values. The reason is to maintain some of these values locally in separate components, instead of dumping them all in the Redux state.

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/GasSlider.tsx
```diff
@@ -0,0 +1,206 @@
...
+            <Typography color={theme.palette.text.secondary} fontSize={14}>
+              Estimated amount needed for this transaction
+            </Typography>
+            <div>
+              <StyledSlider
+                slots={{ thumb: ThumbWithTokenIcon }}
+                aria-label="Native gas conversion amount"
+                defaultValue={0}
+                value={percentage}
+                baseColor="#C1BBF6"
```
emreboga, line 188: These color codes will eventually end up in theme overrides when doing the styling work.

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/GasSlider.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+  const {
+    usdPrices: { data },
+  } = useSelector((state: RootState) => state.tokenPrices);
+
+  const prices = data || {};
+
+  const destChainConfig = config.chains[destChain!];
+  const sourceTokenConfig = config.tokens[sourceToken];
+  const nativeGasToken = config.tokens[destChainConfig!.gasToken];
```
emreboga, line 103: I don't prefer using the `!` operator to force TS compiler for prop's existence. Ideally the code should be able to handle the existence of `destChainConfig` at runtime.
I'll revisit this line.

wormhole-connect/src/views/Bridge/RouteOptions.tsx
```diff
@@ -364,17 +364,12 @@ function RouteOption(props: { route: RouteData; disabled: boolean }) {
...
-    route,
-    routeStates,
-    token,
-    destToken,
-    fromChain,
-    toChain,
-    amount,
-  } = useSelector((state: RootState) => state.transferInput);
-  const { toNativeToken } = useSelector((state: RootState) => state.relay);
+  const { isTransactionInProgress, route, routeStates } = useSelector(
```
emreboga, line 367: Self note: Revisit this

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+    (state: RootState) => state.wallet,
+  );
+
+  const { toNativeToken } = useSelector((state: RootState) => state.relay);
+
+  const { disabled: isGasSliderDisabled, showGasSlider } = useGasSlider({
+    destChain,
+    destToken,
+    route,
+    valid: true,
```
emreboga, line 95: Self note: Revisit.

wormhole-connect/src/hooks/useAvailableRoutes.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    const getAvailable = async () => {
+      const routes: RouteState[] = [];
+      for (const value of config.routes) {
+        const r = value as Route;
+
+        let available = false;
+        let supported = false;
+
+        try {
```
kev1n-peters: I think we should call `isRouteSupported` first and only call `isRouteAvailable` if it's supported to save on RPC calls. Also if a route is not supported, then it will not be available.

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/index.tsx
```diff
@@ -0,0 +1,341 @@
...
+      <SingleRoute
+        config={RoutesConfig[route]}
+        available={true}
+        isSelected={false}
+        showDestinationGasFee
+        title="You will receive"
+      />
+      <Collapse in={showGasSlider}>
+        <GasSlider
+          destinationGasFee={receiveNativeAmt}
```
artursapek: I don't like calling this a fee. This isn't a fee you are paying, it's an amount you are swapping of the destination token you are bridging, to the destination chain's native gas token. A more appropriate name would be `destinationGasReceiveAmount`.

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    return (
+      <Stack direction="row" justifyContent="space-between">
+        <Typography
+          color={theme.palette.text.secondary}
+          fontSize={14}
+          paddingLeft="8px"
+        >
+          Destination gas fee
```
artursapek: Same comment here. This isn't a fee you are paying, it's an amount you are receiving.

wormhole-connect/src/views/v2/Bridge/Routes/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const supportedRoutes = useMemo(() => {
+    if (!routeStates) {
+      return [];
+    }
+
+    return routeStates.filter((rs) => rs.supported);
+  }, [routeStates]);
+
+  const walletsConnected = useMemo(
+    () => Boolean(sendingWallet.address) && Boolean(receivingWallet.address),
```
artursapek: is this `Boolean()` the same as just doing `!!`?

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
```diff
@@ -0,0 +1,318 @@
...
+    if (!props.config.route) {
+      return false;
+    }
+
+    const routeConfig = RouteOperator.getRoute(props.config.route);
+
+    return !(
+      routeConfig.AUTOMATIC_DEPOSIT ||
+      (destChain && isGatewayChain(destChain)) ||
+      destChain === 'sei' ||
```
artursapek: I want to stop having route- or chain-specific logic in our view code. Connect UI should be "dumb" and SDK's abstractions should be enough to satisfy UI needs. what would it take to get rid of these?

wormhole-connect/src/views/v2/Bridge/Routes/index.tsx
emreboga: Yes, they are both converting the value to truthy under the hood. It was more of a styling choice, and I can use `!!` for consistency.

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
emreboga: This was a copy/paste from the old code to show/hide the warning "This transfer requires two transactions ..."
@artursapek is there a utility function or other means to check without a specific chain? Or would `!routeConfig.AUTOMATIC_DEPOSIT` be sufficient?

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/index.tsx
emreboga: Yup, we have already discussed this below and I have it in my follow-up changes:
https://wormholelabsxyz.slack.com/archives/C06T9FN059C/p1721406582348799

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
emreboga: Same response as above.

wormhole-connect/src/hooks/useFetchTokenPricesV2.ts
```diff
@@ -0,0 +1,69 @@
+import config from 'config';
+import { useEffect, useState } from 'react';
+
+const COINGECKO_URL = 'https://api.coingecko.com/';
+const COINGECKO_URL_PRO = 'https://pro-api.coingecko.com/';
+
+const useFetchTokenPricesV2 = (): {
+  prices: any;
```
kev1n-peters: Can we (ideally) specify a different type or add a ts-ignore to suppress the linter error?

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
kev1n-peters: I think for now we can just change this to `!routeConfig.AUTOMATIC_DEPOSIT`. We are not including sei or cosmos gateway initially and whether those are automatic or manual should be provided by the SDK

kev1n-peters (approved):

PR #2271, opened 2024-07-21
Native Sui token transfers fix/hack

kev1n-peters:
This allows the Sui gas token to be transferred.

### DIFF
--- a/wormhole-connect/src/routes/sdkv2/route.ts
+++ b/wormhole-connect/src/routes/sdkv2/route.ts
@@ -19,7 +19,7 @@ import {
   TransferInfoBaseParams,
 } from 'routes/types';
 import { TokenPrices } from 'store/tokenPrices';
-import { ParsedMessage } from 'utils/sdk';
+import { ParsedMessage, toChainName } from 'utils/sdk';
 
 import { SDKv2Signer } from './signer';
 
@@ -37,6 +37,8 @@ export class SDKv2Route {
   TYPE: Route;
   NATIVE_GAS_DROPOFF_SUPPORTED = false;
   AUTOMATIC_DEPOSIT = false;
+  // TODO: remove this
+  IS_TOKEN_BRIDGE_ROUTE = false;
 
   constructor(readonly rc: routes.RouteConstructor, routeType: Route) {
     this.TYPE = routeType;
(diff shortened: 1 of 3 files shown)


wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -95,13 +100,29 @@ export class SDKv2Route {
...
+      // which stores them in a map. Ideally, the SDK should provide a generic method for this conversion
+      // applicable to all chains.
+      // For routes involving the token bridge, there's only one supported destination token so
+      // we don't need to check the address.
+      if (
+        this.IS_TOKEN_BRIDGE_ROUTE &&
+        supportedDestinationTokens.length === 1 &&
+        toChain.chain === 'Sui' &&
+        toTokenIdV2.chain === 'Sui'
+      ) {
```
artursapek, line 124: I'm pretty sure this would return `true` for any given `destToken` if it's on Sui, which is incorrect

wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -45,6 +46,10 @@ export class SDKv2Route {
       this.NATIVE_GAS_DROPOFF_SUPPORTED = true;
       this.AUTOMATIC_DEPOSIT = true;
     }
+    this.IS_TOKEN_BRIDGE_ROUTE =
```
artursapek, line 50: please add a `TODO` above this to call out we should remove it some day 😭

artursapek (commented):
not sure I understand this change tbh. also, does this mean that other SDK users, like people building wallet UIs, would also have this issue with Sui?

wormhole-connect/src/routes/sdkv2/route.ts
kev1n-peters: This is an issue for any native Sui token

kev1n-peters: Changing to check `SUI` token only

wormhole-connect/src/routes/sdkv2/route.ts
kev1n-peters: done

artursapek (approved):

PR #2274, opened 2024-07-22
Resurrect NTT routes

kev1n-peters:


wormhole-connect/src/views/Redeem/DestinationQueued.tsx
```diff
@@ -0,0 +1,173 @@
+import InputContainer from 'components/InputContainer';
```
artursapek, line 1: we should make sure @emreboga is aware of this view being added to the legacy UI. I don't know if the redesign covered any of these edge cases with queues/governor (probably not)

wormhole-connect/src/utils/ntt.ts
```diff
@@ -122,3 +123,31 @@ export const getNttTokenByGroupKey = (
...
+      })
+      .filter((token) => token !== null);
+
+    if (tokens.length > 0) {
+      result.tokens[name] = tokens;
+    }
+  });
+
+  return result;
+}
```
artursapek, line 153: what do you think about converging on SDKv2's NTT config types in Connect's config? since this is going to be a major version change to Connect anyway, I think this is our opportunity to make some breaking changes to the `config` interface. we could just adopt the `NttRoute.Config` type in the `WormholeConnectConfig` interface.

it would be nice to be consistent across Connect & SDK for NTT for those config types and it would remove the need to add this code at all.

kev1n-peters: I think that's a good idea, I can look into changing it

artursapek (approved):

wormhole-connect/src/utils/ntt.ts
kev1n-peters: Will do this in a follow up PR

PR #2278, opened 2024-07-23
Fix ability to transfer CELO token

kev1n-peters:
The SDK doesn't work when we try to transfer "native" from Celo (since the token bridge doesn't have WETH set).

### DIFF
--- a/wormhole-connect/src/config/converter.ts
+++ b/wormhole-connect/src/config/converter.ts
@@ -84,10 +84,7 @@ export class SDKConverter {
         }
       } else {
         // Getting native address
-        const address =
-          getGasToken(token.nativeChain).key === token.key
-            ? 'native'
-            : token.tokenId?.address;
+        const address = this.getNativeTokenAddressV2(token);
         if (!address) throw new Error('no address');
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/config/converter.ts
```diff
@@ -167,4 +160,15 @@ export class SDKConverter {
       return this.tokenIdV2(chainName, foreignAddress.toString());
     }
   }
+
+  getNativeTokenAddressV2(token: TokenConfigV1): string | undefined {
+    const address =
```
artursapek, line 165: this is minor but I think we overuse ternaries in our code. I'd write it like this:

```ts
if (getGasToken(token.nativeChain).key === token.key && token.key !== 'CELO') {
  return 'native';
}

return token.tokenId?.address
```

idk maybe 5 years of golang rotted my brain but to me this makes it easier to read, and clearer that there's a special case.

artursapek (approved):

wormhole-connect/src/config/converter.ts
kev1n-peters: Will refactor in another PR since I don't want your approval to be dismissed

Issue #2239, opened 2024-07-09
[Bug] UI error after a successful BSC Safe -> Eth EOA transfer "Error: missing revert data in call exception"


tonyjin:
## Description and context
After completing a successful NTT transfer of wstETH from a BSC Safe to an Eth EOA, the UI hangs on step 2 in the transaction screen "Send to".

## Steps to reproduce
1. Connect a BSC Safe multisig via WalletConnect.
2. Initiate a transfer of wstETH from the BSC Safe to an Eth EOA. You may need to disconnect/reconnect wallets as the destination wallet is initially auto-populated with the same wallet as the source.
3. Approve the Safe txn for approving token allowance.
4. Approve the Safe txn for approving the transfer.
5. See that the transaction screen is stuck on step 2 with the error:
> Error: missing revert data in call exception; Transaction reverted without a reason string [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (data="0x", transaction={"to":"0xb948a93827d68a82F6513Ad178964Da487fe2BD9","data":"0x396c16b7a7cb292c468bbf3f59949749e3d334ba7bafe8c8f69876c8b1f98cb7ebfce56f","accessList":null}, error={"code":-32000,"message":"execution reverted"}, code=CALL_EXCEPTION, version=providers/5.7.2)

![Screenshot 2024-07-09 at 4 03 18 PM](https://github.com/wormhole-foundation/wormhole-connect/assets/689351/6f64f8a8-9383-40a4-b526-4500235b9753)

Successful txns:
- send: https://bscscan.com/tx/0x970a70eb6775306b49bb5b5bf115b0ca0fd3c70aee3de759730c2278a432d170
- receive: https://etherscan.io/tx/0x38dadfc7d791767a80cea417f61b0f0959c62eab997168416708c597056aed47

## Expected behavior
The transaction screen should successfully progress from step 2 -> step 3 "Transaction complete".

sebastianscatularo:
I've tried this flow successfully @tonyjin 

https://github.com/user-attachments/assets/9853cf06-507b-487a-9283-157b3d49caf9

tsadovska:
Re-checked the issue with the scenario mentioned in the Description.
Sent 0.000001 wstETH from BSC Safe (https://app.safe.global/) to Ethereum (Rabby). 
Transaction on Ethereum was loading around 20 minutes and finished successfully. "Transaction complete" step was shown on https://portalbridge.com/. 
[...]

PR #2314, opened 2024-08-01
#2239 - override properly default rpcs with provided rpcs

sebastianscatularo:
No error appears when default config is present

https://github.com/user-attachments/assets/bd288dd8-a536-4220-a3fc-c23fc0faa578


yuli-ferna (approved):

tsadovska:
Checked on https://preview.portalbridge.com/859350d189/.
User can successfully connect MetaMask wallet via extension, MetaMask wallet via WalletConnect, Rabby wallet via extension. Token balances are successfully loaded.

PR #2315, opened 2024-08-01
solana transaction retries

kev1n-peters:


kev1n-peters:
This is mostly taken from legacy Connect with and modified to throw an exception when simulation fails. We don't want to land transactions that will fail on-chain.

wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -69,52 +75,129 @@ export function fetchOptions() {
...
   options?: ConfirmOptions,
 ) {
-  if (!wallet || !wallet.signAndSendTransaction) {
-    throw new Error('wallet.signAndSendTransaction is undefined');
+  if (!wallet) throw new Error('Wallet not found');
+  if (!config.rpcs.solana) throw new Error('Solana RPC not found');
+
+  const commitment = options?.commitment ?? 'finalized';
+
+  const unsignedTx = request.transaction.transaction;
```
emreboga, line 92: I assume this property chain has checked before undefined?

wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -69,52 +75,129 @@ export function fetchOptions() {
...
+  const connection = new Connection(config.rpcs.solana);
+  const { blockhash, lastValidBlockHeight } =
+    await connection.getLatestBlockhash(commitment);
```
emreboga, line 95: In case of any connection errors to get the latest block hash, is this going to throw or just return undefined? If it's the later, do you think we should check the value afterward?

wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -69,52 +75,129 @@ export function fetchOptions() {
...
+
+  if (isVersionedTransaction(unsignedTx)) {
+    const message = TransactionMessage.decompile(unsignedTx.message);
+    message.recentBlockhash = blockhash;
+    message.instructions.push(
+      ...(await createPriorityFeeInstructions(connection, unsignedTx)),
```
emreboga, line 101: Does `createPriorityFeeInstructions` end up being an external call? If so, do you thin we can check the returned value before deconstruct and push?

wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -69,52 +75,129 @@ export function fetchOptions() {
...
+    );
+    unsignedTx.message = message.compileToV0Message();
+    unsignedTx.sign(request.transaction.signers ?? []);
+  } else {
+    unsignedTx.recentBlockhash = blockhash;
+    unsignedTx.lastValidBlockHeight = lastValidBlockHeight;
+    unsignedTx.add(
+      ...(await createPriorityFeeInstructions(connection, unsignedTx)),
```
emreboga, line 109: Same as above.

wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -69,52 +75,129 @@ export function fetchOptions() {
...
+  transaction: Transaction | VersionedTransaction,
+  commitment?: Commitment,
+) {
+  const response = await (isVersionedTransaction(transaction)
+    ? connection.simulateTransaction(transaction, {
+        commitment,
+        replaceRecentBlockhash: true,
+      })
+    : connection.simulateTransaction(transaction));
+  // const response = await connection.simulateTransaction(transaction);
```
emreboga: Safe to remove this line?

wormhole-connect/src/utils/wallet/solana.ts
kev1n-peters: This property will exist on the `SolanaUnsignedTransaction` - passing that is the caller's responsibility

wormhole-connect/src/utils/wallet/solana.ts
kev1n-peters: This is going to throw if we can't get the blockhash which is what we want

wormhole-connect/src/utils/wallet/solana.ts
kev1n-peters: `createPriorityFeeInstructions` is in this same file and it will throw if we can't simulate the transaction. It will always return a non-empty array currently. Even if it doesn't the push would be a no-op

wormhole-connect/src/utils/wallet/solana.ts
kev1n-peters: Good catch

emreboga (approved):

PR #2317, opened 2024-08-02
xlabs/portal-bridge-ui#970 - disable switch network on manual target address

sebastianscatularo:
https://github.com/user-attachments/assets/d036b4f4-7543-48ca-adac-b7b8f82982dd


yuli-ferna (approved):
Thanks!

PR #2340, opened 2024-08-07
Added useFetchRedeemTx hook

kev1n-peters:
The SDK currently doesn't set the redeem TX in all cases, so we use this hook as a stop-gap until it does.

### DIFF
--- a/wormhole-connect/src/hooks/useFetchRedeemTx.ts
+++ b/wormhole-connect/src/hooks/useFetchRedeemTx.ts
@@ -20,7 +20,7 @@ const useFetchRedeemTx = (): void => {
   useEffect(() => {
     if (
       !receipt ||
-      receipt.state < TransferState.Attested ||
+      receipt.state < TransferState.DestinationInitiated ||
       'originTxs' in receipt === false ||
       receipt.originTxs.length === 0
     ) {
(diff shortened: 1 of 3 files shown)


wormhole-connect/src/hooks/useFetchRedeemTx.ts
```diff
@@ -0,0 +1,67 @@
...
+              if (isActive) {
+                dispatch(setRedeemTx(redeemTx));
+              }
+              break;
+            }
+          }
+        } catch (e) {
+          console.warn(e);
+        }
+        await sleep(10_000);
```
artursapek, line 58: do we want this thing to try forever? probably better to give up eventually no?

also, I think you could run this only when we see a state >= `TransferState.DestinationInitiated`. no reason to expect a destination tx when we're at `TransferState.Attested`.

wormhole-connect/src/hooks/useFetchRedeemTx.ts
```diff
@@ -0,0 +1,67 @@
...
+      'originTxs' in receipt === false ||
+      receipt.originTxs.length === 0
+    ) {
+      return;
+    }
+
+    let isActive = true;
+
+    const fetchRedeemTx = async () => {
+      const wormholeApi = config.wormholeApi.replace(/\/$/, '');
```
artursapek, line 34: TODO, make `api` package of new SDK handle this internally :P

kev1n-peters: good idea: https://github.com/wormhole-foundation/wormhole-sdk-ts/pull/661

wormhole-connect/src/hooks/useFetchRedeemTx.ts
kev1n-peters: Maybe we could give up after a minute? If the state is DestinationInitiated and after a minute we still haven't fetched it then we can give up

artursapek: ok this goes for 1 min 40 sec but lgtm

artursapek (approved):

PR #2343, opened 2024-08-07
Mayan route

artursapek:


wormhole-connect/src/config/routes.ts
```diff
@@ -107,4 +107,12 @@ export const RoutesConfig: {
     icon: XLabsIcon,
     pendingMessage: 'Waiting for Wormhole network consensus . . .',
   },
+  [Route.Mayan]: {
+    route: Route.Mayan,
+    name: 'Mayan Swap',
+    providedBy: 'Mayan Finance',
+    link: 'https://mayan.finance/',
+    icon: XLabsIcon,
```
kev1n-peters, line 115: Should we change this to `WormholeIcon` since xlabs doesn't relay for mayan?

wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -47,16 +47,10 @@ export class SDKv2Route {
...
-      this.AUTOMATIC_DEPOSIT = true;
-    } else if (routeType === Route.NttRelay) {
-      this.AUTOMATIC_DEPOSIT = true;
-    } else if (routeType === Route.CCTPRelay) {
-      this.NATIVE_GAS_DROPOFF_SUPPORTED = true;
-      this.AUTOMATIC_DEPOSIT = true;
-    }
+
+    this.AUTOMATIC_DEPOSIT = rc.IS_AUTOMATIC;
+    this.NATIVE_GAS_DROPOFF_SUPPORTED = rc.NATIVE_GAS_DROPOFF_SUPPORTED;
```
kev1n-peters, line 52: I realized that this will enable the gas slider for NTT and the NTT automatic route sets `NATIVE_GAS_DROPOFF_SUPPORTED = true`, but it's not *actually* supported by the relayer contracts at the moment. I will change this in the NTT SDK to be `false`.

wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -178,25 +202,61 @@ async function createPriorityFeeInstructions(
...
-        commitment,
-        replaceRecentBlockhash: true,
-      })
-    : connection.simulateTransaction(transaction));
-  if (response.value.err) {
-    throw new Error(`Simulation failed: ${response.value.err}`);
+  if (isVersionedTransaction(transaction)) {
+    // This is required for versioned transactions - simulateTransaction throws
+    // if recentBlockhash is an empty string.
+    const { blockhash } = await connection.getLatestBlockhash(commitment);
```
kev1n-peters, line 208: Can we set this using the `blockhash` in `signAndSendTransaction` instead to reduce RPC calls?

wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -178,25 +202,61 @@ async function createPriorityFeeInstructions(
...
+    transaction.message.recentBlockhash = blockhash;
```
kev1n-peters, line 212: Can we set this using the `blockhash` in `signAndSendTransaction` instead to reduce RPC calls?

wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -178,25 +202,61 @@ async function createPriorityFeeInstructions(
...
+      : connection.simulateTransaction(transaction));
+
+    if (response.value.err) {
+      // In some cases which aren't deterministic, like a slippage error, we can retry the
+      // simulation a few times to get a successful response.
+      if (response.value.logs) {
+        for (const line of response.value.logs) {
+          if (line.includes('SlippageToleranceExceeded')) {
+            console.info('Slippage failure during simulation. Trying again.');
+            continue simulationLoop;
```
kev1n-peters, line 235: Should there be a small sleep in here?

wormhole-connect/src/config/routes.ts
artursapek: this was a TODO that I never did because the new UI doesn't even use these icons

wormhole-connect/src/utils/wallet/solana.ts
artursapek: I don't think so because the simulation fails if it's not set, and we have to do the simulation to create the priority fee ix before we call `signAndSend`

wormhole-connect/src/utils/wallet/solana.ts
artursapek: yeah probably a good idea. will add

kev1n-peters (dismissed):

kev1n-peters (approved):

PR #2345, opened 2024-08-08
disable cosmos chains until cosmos gateway route is supported

kev1n-peters:


artursapek (dismissed):
`TODO SDKV2` comments above would be appreciated but lgtm

artursapek (approved):

PR #2357, opened 2024-08-10
prevent screen to crash due re render issue on network switch

sebastianscatularo:
https://github.com/user-attachments/assets/1113882f-a70f-485a-b8d1-c13caacf2ddf


yuli-ferna (approved):
:O

PR #2367, opened 2024-08-12
[Redesign] Redeem view

emreboga:
This PR adds the Redeem view for the Connect redesign.

Demo for the automatic route:
https://www.loom.com/share/1a723691d68b42a39a6705b625e8941f?sid=87a74ac9-8f83-40de-8a34-1fbf8e1567ee

Demo for the manual route:
https://www.loom.com/share/a3e1470c93b8494bba7840a6f73bf33b?sid=a257e56e-4f96-47b1-8bb2-0aa8ad2016ea

### DIFF
--- a/wormhole-connect/src/hooks/useComputeQuote.ts
+++ b/wormhole-connect/src/hooks/useComputeQuote.ts
@@ -6,7 +6,7 @@ import { setReceiveAmount, setReceiveAmountError } from 'store/transferInput';
 import type { Route } from 'config/types';
 import type { ChainName } from 'sdklegacy';
 import { getRoute } from 'routes/mappings';
-import { setReceiveNativeAmt, setRelayerFee } from 'store/relay';
+import { setReceiveNativeAmt, setRelayerFee, setEta } from 'store/relay';
 import { amount as sdkAmount } from '@wormhole-foundation/sdk';
 import { toDecimals } from 'utils/balance';
 import config from 'config';
@@ -80,6 +80,7 @@ const useComputeQuote = (props: Props): returnProps => {
             dispatch(setReceiveAmountError(quote.error.message));
             dispatch(setReceiveNativeAmt(0));
             dispatch(setRelayerFee(undefined));
+            dispatch(setEta(0));
           }
           setIsFetching(false);
           return;
@@ -119,6 +120,10 @@ const useComputeQuote = (props: Props): returnProps => {
           } else {
             dispatch(setRelayerFee(undefined));
           }
+
+          if (quote.eta) {
+            dispatch(setEta(quote.eta));
+          }
         }
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
       } catch (e: any) {
@@ -126,6 +131,7 @@ const useComputeQuote = (props: Props): returnProps => {
           dispatch(setReceiveAmountError(e.message));
           dispatch(setReceiveNativeAmt(0));
           dispatch(setRelayerFee(undefined));
+          dispatch(setEta(0));
         }
       } finally {
         setIsFetching(false);
--- a/wormhole-connect/src/hooks/useComputeQuote.ts
+++ b/wormhole-connect/src/hooks/useComputeQuote.ts
@@ -1,11 +1,7 @@
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 import { useDispatch } from 'react-redux';
 
-import {
-  setReceiveAmount,
-  setFetchingReceiveAmount,
-  setReceiveAmountError,
-} from 'store/transferInput';
+import { setReceiveAmount, setReceiveAmountError } from 'store/transferInput';
 
 import type { Route } from 'config/types';
 import type { ChainName } from 'sdklegacy';
@@ -25,7 +21,11 @@ type Props = {
   toNativeToken: number;
 };
 
-export const useComputeQuote = (props: Props): void => {
+type returnProps = {
+  isFetching: boolean;
+};
+
+const useComputeQuote = (props: Props): returnProps => {
   const {
     sourceChain,
     destChain,
@@ -37,6 +37,8 @@ export const useComputeQuote = (props: Props): void => {
   } = props;
 
   const dispatch = useDispatch();
+
+  const [isFetching, setIsFetching] = useState(false);
 
   useEffect(() => {
     if (
@@ -53,7 +55,7 @@ export const useComputeQuote = (props: Props): void => {
     let isActive = true;
     const computeQuote = async () => {
       try {
-        dispatch(setFetchingReceiveAmount());
+        setIsFetching(true);
 
         const parsedAmount = Number.parseFloat(amount);
         if (Number.isNaN(parsedAmount)) {
@@ -79,6 +81,7 @@ export const useComputeQuote = (props: Props): void => {
             dispatch(setReceiveNativeAmt(0));
             dispatch(setRelayerFee(undefined));
           }
+          setIsFetching(false);
           return;
         }
 
@@ -124,6 +127,8 @@ export const useComputeQuote = (props: Props): void => {
           dispatch(setReceiveNativeAmt(0));
           dispatch(setRelayerFee(undefined));
         }
+      } finally {
+        setIsFetching(false);
       }
     };
 
@@ -142,4 +147,10 @@ export const useComputeQuote = (props: Props): void => {
     sourceChain,
     dispatch,
   ]);
+
+  return {
+    isFetching,
+  };
 };
+
+export default useComputeQuote;
--- a/wormhole-connect/src/hooks/useComputeQuoteV2.ts
+++ b/wormhole-connect/src/hooks/useComputeQuoteV2.ts
@@ -22,7 +22,7 @@ type Props = {
 };
 
 type returnProps = {
-  eta: Number;
+  eta: number;
   isFetching: boolean;
   receiveAmount: string;
   receiveAmountError: string;
--- a/wormhole-connect/src/hooks/useComputeQuoteV2.ts
+++ b/wormhole-connect/src/hooks/useComputeQuoteV2.ts
@@ -22,9 +22,10 @@ type Props = {
 };
 
 type returnProps = {
+  eta: Number;
+  isFetching: boolean;
   receiveAmount: string;
   receiveAmountError: string;
-  isFetching: boolean;
   receiveNativeAmt: number;
   relayerFee: number;
 };
@@ -40,9 +41,10 @@ const useComputeQuoteV2 = (props: Props): returnProps => {
     toNativeToken,
   } = props;
 
+  const [eta, setEta] = useState(0);
+  const [isFetching, setIsFetching] = useState(false);
   const [receiveAmount, setReceiveAmount] = useState('');
   const [receiveAmountError, setReceiveAmountError] = useState('');
-  const [isFetching, setIsFetching] = useState(false);
   const [receiveNativeAmt, setReceiveNativeAmt] = useState(0);
   const [relayerFee, setRelayerFee] = useState(0);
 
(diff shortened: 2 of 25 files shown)


wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
```diff
@@ -149,20 +136,23 @@ const SingleRoute = (props: Props) => {
   }, [destChain, isFetchingQuote, props.destinationGasDrop]);
 
   const timeToDestination = useMemo(() => {
+    const etaMins = Math.floor((estimatedTime as number) / 60000);
```
emreboga: TO-DO: Will replace these with the existing `millisToMinutesAndSeconds`

wormhole-connect/src/views/v2/Redeem/TransactionDetails/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+        </Link>
+      </Stack>
+    );
+  }, [sendTx]);
+
+  const timeToDestination = useMemo(() => {
+    let etaDisplay: string | ReactNode = <CircularProgress size={14} />;
+
+    if (eta) {
+      const etaMins = Math.floor(eta / (1000 * 60));
```
emreboga: TO-DO: Will replace these with the existing `millisToMinutesAndSeconds`.

wormhole-connect/src/hooks/useComputeQuoteV2.ts
emreboga, line 1: `useComputeQuoteV2` is used in SingleRoute component where we need to render multiple routes and each one of them has a separate quote.

wormhole-connect/src/hooks/useComputeQuote.ts
emreboga, line 1: `useComputeQuote` is used for a single route during "Review transaction" or Redeem view.

wormhole-connect/src/hooks/useTrackTransfer.ts
emreboga, line 1: A little bit of code-styling here, nothing else.

wormhole-connect/src/utils/sdkv2.ts
```diff
@@ -49,6 +49,9 @@ export interface TransferInfo {
   // Amount of native gas being received, in destination gas token units
   // For example 1.0 is 1.0 ETH, not 1 wei
   receiveNativeAmount?: number;
+
+  // ETA for the route this transfer was initiated on
+  eta?: number;
```
emreboga, line 54: Adding eta prop to `TransferInfo` type

wormhole-connect/src/utils/sdkv2.ts
```diff
@@ -192,6 +195,15 @@ const parseTokenBridgeReceipt = async (
     txData.tokenKey = tokenV1.key;
     txData.receivedTokenKey = tokenV1.key;
     txData.receiveAmount = txData.amount;
+    txData.receiveNativeAmount = Number(
```
emreboga, line 198: Here I added parsing for `receiveNativeAmount` and `relayerFee` to `parseTokenBridgeReceipt`. This is needed when resuming a transaction that takes user to the Redeem view (see `views/TxSearch.tsx`).
I believe we need to add a similar parsing to `parseCCTPReceipt` and `parseNttReceipt` as well. @artursapek if that's the case, can you point me to sample transaction configs for those two routes so I can debug, rev-eng and add those parsing logic as well?

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/index.tsx
```diff
@@ -210,6 +211,9 @@ const ReviewTransaction = (props: Props) => {
         throw new Error("Can't find txid in receipt");
       }
 
+      // Set the start time of the transaction
+      dispatch(setTimestamp(Date.now()));
```
emreboga, line 215: Note: This way we will have a start timestamp for a new transaction, but still need a solution to get the start time for "Resume transaction".
Cc: @artursapek

wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
emreboga, line 1: Changes in this file are related to couple of bug fixes rather than the Redeem view. I didn't want to exclude them as the changes were pretty straightforward.

wormhole-connect/src/hooks/useTrackTransfer.ts
```diff
@@ -18,37 +18,46 @@ const useTrackTransfer = (): void => {
...
       const { route, receipt } = routeContext;
+
       if (!route || !receipt) {
         return;
       }
+
       let stateChanged = false;
+
       while (isActive && !isCompleted(receipt) && !stateChanged) {
+        console.log(`Current state: ${receipt.state}`);
```
kev1n-peters: Do we want to keep these console logs?

emreboga: I moved them outside of try/catch, but I don't think we need to keep them in the first place. We already log in catch block.
@artursapek ?

emreboga: Removed 👍

wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
```diff
@@ -106,6 +111,26 @@ const AmountInput = (props: Props) => {
     );
   }, [isInputDisabled, balances, tokenBalance, sendingWallet.address]);
 
+  useEffect(() => {
+    // Validation for the amount value
+    const amountValidation = validateAmount(
+      tokenAmount,
+      tokenBalance,
+      getMaxAmt(route),
+      isCctp(sourceToken, destToken, sourceChain, destChain),
```
artursapek: we shouldnt have any code like this anymore, referring to specific routes in shared components. is it easy to remove this?

wormhole-connect/src/hooks/useComputeQuote.ts
```diff
@@ -1,16 +1,12 @@
-import { useEffect } from 'react';
```
artursapek, line 1: is the plan to get rid of this and switch entirely to `useComputeQuoteV2`? what are the outstanding items before we can do that?

wormhole-connect/src/utils/index.ts
```diff
@@ -30,6 +30,12 @@ export function trimAddress(address: string, max = 6): string {
   );
 }
 
+export function trimTxHash(txHash: string, max = 6): string {
+  const start = txHash.slice(0, max);
+  const end = txHash.slice(txHash.length - 4, txHash.length);
```
artursapek, line 35: what is the `max` argument supposed to do here? is that the max length of the output? why are we using that for the first part, but hard-coding 4 for the second part?

wormhole-connect/src/utils/sdkv2.ts
artursapek: this is just the native gas amount the user is receiving, correct?

the way I wrote that parsing code was I just did a transfer, paused my devtools, and inspected the attestation. I can help add this to the other routes.

wormhole-connect/src/utils/transferValidation.ts
```diff
@@ -351,8 +351,17 @@ export const useValidate = () => {
   }, [debouncedStateForValidation, dispatch]);
 };
 
+export const minutesAndSecondsWithPadding = (
+  minutes: number,
+  seconds: number,
+) => {
+  const minutesDisp = `${minutes < 10 ? `0${minutes}` : minutes}`;
+  const secondsDisp = `${seconds < 10 ? `0${seconds}` : seconds}`;
```
artursapek: you could use `padStart` here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/index.tsx
artursapek: since we can only resume txs that are finalized on the source chain & attested, the ETA for a resume transaction is just the destination chain finality time. you can actually just get this using `finality` in the SDK: https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/connect/src/protocols/tokenBridge/tokenTransfer.ts#L720

artursapek: since we can only resume transfers that have been finalized on the source chain & attested, there's no real ETA to show for them. as soon as the destination transaction goes through, the transfer is complete. so I think there's no need for an ETA for those.

wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
emreboga: @artursapek isCctp is an optional parameter. I can remove it from here so it won't check this case:
```
if (isCctp && numAmount >= CCTP_MAX_TRANSFER_LIMIT) { ... }
```
If that's ok, I can remove it?

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/index.tsx
emreboga: @artursapek What if the user took note of the tx hash and came back a few mins later to resume it? Can we ignore that case? It'd still be in source finality I believe.

wormhole-connect/src/utils/index.ts
emreboga: Good catch :) This was a copy/paste/modify from `trimAddress` above. I believe that's the max for the first part and the second part is always 4 chars. I don't think we need it TBH. I will fix it to `6 chars ... 4 chars`.

wormhole-connect/src/utils/sdkv2.ts
emreboga: @artursapek I did the same to figure out where to get those amounts in the attestation object. Which network/token configurations will show me CCTP and NTT routes?

wormhole-connect/src/hooks/useComputeQuote.ts
emreboga: @artursapek That was the original plan, but it seems until we do a redesign of the app state management (either replacing or improving Redux), we need to pass the selected route's quote props to Redux state.
One alternative is I can combine useComputeQuote and useComputeQuoteV2 and pass a param whether to update Redux as well. So it won't update Redux for the case of SingleRoute and just return the quote props.

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/index.tsx
emreboga: I think this is good for internal release. I've opened an item to revisit this:
https://www.notion.so/wormholelabs/UI-Feature-Do-we-need-a-transaction-creation-date-in-the-SDK-a8969f1344dc4583bd50eddb62009694?pvs=4

wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
emreboga: Removed the CCTP check in the redesign code.

wormhole-connect/src/hooks/useComputeQuote.ts
emreboga: I'll handle this as a tech-debt post-beta.
https://www.notion.so/wormholelabs/UI-TECHDEBT-Converge-useComputeQuote-and-useComputeQuoteV2-into-a-single-hook-9bc8587c679b419eb57b928520317e12?pvs=4

wormhole-connect/src/utils/sdkv2.ts
emreboga: I've also added an item to track this work for prod release:
https://www.notion.so/wormholelabs/UI-Feature-Add-receiveNativeAmount-and-relayerFee-to-CCTP-and-NTT-parsers-in-utils-sdkv2-ts-b9c0018830da4160b962535e2e76f20a?pvs=4

artursapek (approved):

PR #2373, opened 2024-08-13
Removed ChainName from Connect; use Chain from SDK instead

kev1n-peters:
Also removed some dead code


wormhole-connect/src/config/converter.ts
```diff
@@ -18,23 +19,6 @@ export class SDKConverter {
...
-  toChainNameV1(chain: v2.Chain) {
-    return this.wh.toChainName(this.toChainIdV1(chain));
-  }
-
-  toChainV2(chain: v1.ChainName | v1.ChainId): v2.Chain {
-    if (typeof chain !== 'number' && typeof chain !== 'string') {
-      throw new Error(JSON.stringify(chain));
-    }
-    return v2.toChain(this.wh.toChainId(chain));
-  }
```
artursapek, line 36: love it

wormhole-connect/src/sdklegacy/config/TESTNET.ts
```diff
@@ -4,29 +4,29 @@ import { WormholeConfig, Context, ChainConfig } from '../types';
...
+  Xlayer: 37,
+  Wormchain: 3104,
+  Osmosis: 20,
+  Cosmoshub: 4000,
+  Evmos: 4001,
+  Kujira: 4002,
+  Sepolia: 10002,
+  ArbitrumSepolia: 10003,
+  BaseSepolia: 10004,
+  OptimismSepolia: 10005,
```
artursapek: can we get rid of this entirely and rely on the new SDK? https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/core/base/src/constants/chains.ts#L6

artursapek (dismissed):

artursapek (approved):

PR #2382, opened 2024-08-14
[Redesign] Fixing the information displayed on a Single Route

emreboga:
This fixes the bug reported here:
https://www.notion.so/wormholelabs/UI-BUG-Route-card-info-needs-to-be-updated-d79d5b6662c74ab382753a2d4a54ffa4?pvs=4

Please see the loom below:
https://www.loom.com/share/13afe3843fce4b3fa138f413cb5ec17b?sid=53804e4d-39a1-4769-b14b-21a5c92ca6e4


artursapek (approved):
the changes look fine but are we really getting rid of the route name completely in the UI? what happens when there's multiple automatic routes (Mayan + automatic CCTP for example)? are we only ever showing one of automatic and one manual option at a time?

PR #2383, opened 2024-08-15
[Redesign] Updating route information after switching source/destination network/token config

emreboga:
Fixes the issue:
https://www.notion.so/wormholelabs/UI-BUG-Switch-dest-network-token-after-selecting-a-route-90130cf4421e4a658393f0667fe9df62?pvs=4

Please see for the fix:
https://www.loom.com/share/1f78384fd20a46c4bd937ed0865c1678?sid=60b9c52c-01b1-4b9c-a354-862bc2132851

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
@@ -126,7 +126,7 @@ const AmountInput = (props: Props) => {
     }
 
     setValidationResult(amountValidation);
-  }, [tokenAmount, validationResult]);
+  }, [tokenAmount, tokenBalance, route, validationResult]);
 
   const maxButton = useMemo(() => {
     return (
--- a/wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
@@ -60,7 +60,7 @@ const SingleRoute = (props: Props) => {
 
   const { prices: tokenPrices } = useFetchTokenPricesV2();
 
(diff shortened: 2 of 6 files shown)


artursapek (dismissed):

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
```diff
@@ -242,7 +242,7 @@ const SingleRoute = (props: Props) => {
   }
 
   return (
-    <div key={routeName} className={classes.container}>
+    <div key={name} className={classes.container}>
```
emreboga, line 245: We need to use the name prop here as route can be same for automatic and manual of the same route.

wormhole-connect/src/views/v2/Bridge/index.tsx
```diff
@@ -321,16 +320,6 @@ const Bridge = () => {
     );
   }, []);
 
-  const routeList = useMemo(() => {
```
emreboga, line 324: We don't need a separate memoized variable for Routes.

wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
```diff
@@ -126,7 +126,7 @@ const AmountInput = (props: Props) => {
     }
 
     setValidationResult(amountValidation);
-  }, [tokenAmount, validationResult]);
+  }, [tokenAmount, tokenBalance, route, validationResult]);
```
emreboga, line 129: These were missing in validation checks, hence not checking amount when route and token changes (which causes a change in tokenBalance value).

artursapek (approved):

PR #2387, opened 2024-08-15
Fix: Possible to start a transfer with disabled route

yuli-ferna:
Issue: https://github.com/XLabs/portal-bridge-ui/issues/999#issuecomment-2288539778

https://github.com/user-attachments/assets/528db471-b2bb-4efb-ac58-6997eee5d21c

### DIFF
--- a/wormhole-connect/src/components/ConnectWallet.tsx
+++ b/wormhole-connect/src/components/ConnectWallet.tsx
@@ -99,9 +99,7 @@ function ConnectWallet(props: Props) {
     (popupState?: any) => {
       if (disabled) return;
       if (popupState) popupState.close();
-      if (side === 'destination' && manualAddressTarget) {
-        dispatch(setManualAddressTarget(false));
-      }
+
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/components/ConnectWallet.tsx
```diff
@@ -99,9 +99,7 @@ function ConnectWallet(props: Props) {
     (popupState?: any) => {
       if (disabled) return;
       if (popupState) popupState.close();
-      if (side === 'destination' && manualAddressTarget) {
```
sebastianscatularo, line 102: Have you checked if the close without connection works after this change?

sebastianscatularo (approved):

wormhole-connect/src/components/ConnectWallet.tsx
yuli-ferna: if you mean that when you change wallet and just close it, the current state is maintained, yes.

PR #2390, opened 2024-08-15
[Redesign] Update the UI state when a (selected) route is not available

emreboga:
Fixes the issue: 
https://www.notion.so/wormholelabs/UI-BUG-Unavailable-route-should-not-be-auto-selected-c69f512ecefb4e2ba17d37df9c313c20?pvs=4

Please see the loom for details:
https://www.loom.com/share/e9731715392d496ea3b85d0fb6a5b2c9?sid=fab032c1-63c6-4d32-ad3c-9522266777f2

### DIFF
--- a/wormhole-connect/.eslintrc.json
+++ b/wormhole-connect/.eslintrc.json
@@ -14,19 +14,13 @@
     "comma-dangle": ["error", "always-multiline"],
     "semi": ["error", "always"],
 
-    "@typescript-eslint/no-explicit-any": "off",
-
-    "@typescript-eslint/explicit-module-boundary-types": ["error"],
-    "@typescript-eslint/explicit-module-boundary-types": "off",
-
-    "@typescript-eslint/no-non-null-assertion": ["error"],
-    "@typescript-eslint/no-non-null-assertion": "off",
-
     "@typescript-eslint/ban-ts-comment": "off",
+    "@typescript-eslint/explicit-module-boundary-types": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-non-null-assertion": "off",
+    "@typescript-eslint/no-unused-vars": ["warn", { "args": "none" }],
 
     "no-constant-condition": "off",
-
-    "no-unused-vars": "off",
-    "@typescript-eslint/no-unused-vars": ["warn", { "args": "none" }]
+    "no-unused-vars": "off"
   }
 }
--- a/wormhole-connect/src/routes/sdkv2/route.ts
+++ b/wormhole-connect/src/routes/sdkv2/route.ts
@@ -175,7 +175,9 @@ export class SDKv2Route {
       console.error(`Error thrown in isRouteAvailable`, e);
       // TODO is this the right place to try/catch these?
       // or deeper inside SDKv2Route?
-      return false;
+
+      // Re-throw for the caller to handle and surface the error message
+      throw e;
     }
 
     return true;
--- a/wormhole-connect/src/store/transferInput.ts
+++ b/wormhole-connect/src/store/transferInput.ts
@@ -104,7 +104,7 @@ export type RouteState = {
   name: string;
   supported: boolean;
   available: boolean;
-  availabilityError: string;
+  availabilityError?: string;
 };
 
 export interface TransferInputState {
--- a/wormhole-connect/src/store/transferInput.ts
+++ b/wormhole-connect/src/store/transferInput.ts
@@ -104,6 +104,7 @@ export type RouteState = {
   name: string;
   supported: boolean;
   available: boolean;
+  availabilityError: string;
 };
 
 export interface TransferInputState {
--- a/wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenWarnings.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenWarnings.tsx
@@ -27,18 +27,18 @@ const useStyles = makeStyles()((theme: any) => ({
   },
   link: {
     textDecoration: 'underline',
-    opacity: '0.8',
+    opacity: 0.8,
     padding: '4px 0',
     cursor: 'pointer',
     '&:hover': {
-      opacity: '1',
+      opacity: 1,
     },
   },
   disabled: {
     cursor: 'not-allowed',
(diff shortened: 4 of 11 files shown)


wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenWarnings.tsx
```diff
@@ -27,18 +27,18 @@ const useStyles = makeStyles()((theme: any) => ({
   },
   link: {
     textDecoration: 'underline',
-    opacity: '0.8',
+    opacity: 0.8,
```
emreboga, line 30: I've changed the opacity value to be between 0-1 (instead of a percentage) and a number (instead of string) across the redesign code for consistency. You will see more instances of these below.

wormhole-connect/src/views/v2/Bridge/index.tsx
```diff
@@ -136,10 +136,10 @@ const Bridge = () => {
 
     const routeState = routeStates?.find((rs) => rs.name === route);
 
-    if (routeState?.available && routeState?.supported && !selectedRoute) {
+    if (routeState?.supported && routeState?.available) {
```
emreboga, line 139: We don't need to check selectedRoute as the side-effect will no longer triggered its changes. I've added this check to make sure it will init the selectedRoute at the mount of this component, but I think we no longer need it because:
1. route will update only when user explicitly selects a route.
2. routeStates will update when user changes the chain/token configuration.

wormhole-connect/src/views/v2/Bridge/index.tsx
```diff
@@ -369,7 +369,7 @@ const Bridge = () => {
           ) {
             const route = routeStates.find((rs) => rs.name === selectedRoute);
 
-            if (route?.available) {
+            if (route?.supported && route?.available) {
```
emreboga, line 372: Although useAvailableRoutes won't set available prop if a route is not supported, I thought it won't hurt to add a check here as well.

wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -180,7 +180,9 @@ export class SDKv2Route {
       console.error(`Error thrown in isRouteAvailable`, e);
       // TODO is this the right place to try/catch these?
       // or deeper inside SDKv2Route?
-      return false;
+
+      // Re-throw for the caller to handle and surface the error message
+      throw e;
```
emreboga, line 180: @artursapek @kev1n-peters I need your input here. It seems `isRouteAvailable` function here eats up the exception, but we need that to surface the error. I first thought maybe we can return the error message instead of re-throwing; but that messes up the API signature (a method called isRouteAvailable returning a string error message, IDK 🤔). Do you see any issues with the re-throw? There is one more call to this in `routes/operator.ts`, do you think we need another try/catch there as well?

wormhole-connect/.eslintrc.json
emreboga, line 1: Re-ordered and removed duplicate rules here, no other changes.

wormhole-connect/src/routes/sdkv2/route.ts
kev1n-peters: I don't see an issue with re-throwing at the moment

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
```diff
@@ -165,12 +165,33 @@ const SingleRoute = (props: Props) => {
...
+    }
+
+    return (
+      <>
+        <Divider flexItem sx={{ marginTop: '8px' }} />
+        <Stack direction="row" alignItems="center">
+          <WarningIcon htmlColor={theme.palette.error.main} />
+          <Stack sx={{ padding: '16px' }}>
+            <Typography color={theme.palette.error.main} fontSize={14}>
+              {props.error}
```
artursapek, line 182: I don't think exposing the raw `error.message` from `isRouteAvailable` in the UI is a good idea. there's no guarantee it's human-friendly. it could be a crazy looking network error. we should have some code that tries parsing it into a human-friendly string, and defaults to something generic like "Route is unavailable, try again later."

wormhole-connect/src/store/transferInput.ts
```diff
@@ -104,6 +104,7 @@ export type RouteState = {
   name: string;
   supported: boolean;
   available: boolean;
+  availabilityError: string;
```
artursapek: I know empty strings are "falsy" but in general I like being explicit when something is optional and using `undefined` rather than an empty string when there is no error.

```suggestion
  availabilityError?: string;
```

artursapek (changes requested):

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
emreboga: I see your point! I've checked and there is no specific error code from unavailability exceptions. Let me repro those ones and come up with a simple parser to at least detect network errors.

wormhole-connect/src/store/transferInput.ts
emreboga: Fair enough.

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
emreboga: @artursapek the ideal solution seems to be SDK adding error codes (or anything) to distinguies human readable ones from others. Until then I have two alternatives to detect whether an error message is non-html:

Using the DOMParser:
```
var errorRepsonse = new DOMParser().parseFromString(
  e.message,
  'text/html',
);
const isHtml = Array.from(errorRepsonse.body.childNodes).some(
  (node) => node.nodeType === 1,
);
```

Or just simple RegEx:
```
const hasHTMLTags = /<\/?[a-z][\s\S]*>/i.test(e.message);
```

WDYT?

artursapek: (discussed privately - SDK needs to normalize errors to a standard set. until then, use generic message to avoid leaking technical errors that confuse users)

artursapek (approved):

PR #2392, opened 2024-08-16
Pre-select network when asset picker is opened

nikarm22:
Fixes: https://www.notion.so/wormholelabs/UI-BUG-tapping-on-select-chain-and-token-should-open-full-menu-8f363535e6e54acf8936524b663ec6ef


emreboga (approved):
LGTM!

PR #2407, opened 2024-08-20
[bug] governor limit message

danielisaacgeslin:
[update governor limit for polygon big tx limit to 500k not 200k for polygon](https://github.com/XLabs/portal-bridge-ui-issues/issues/128)

https://github.com/user-attachments/assets/0ac21796-0bc9-4020-b3b9-db1c13a0d4d9

### DIFF
--- a/wormhole-connect/src/hooks/useIsTransferLimited.ts
+++ b/wormhole-connect/src/hooks/useIsTransferLimited.ts
@@ -28,13 +28,13 @@ interface TokenList {
 
 interface AvailableNotionalByChainEntry {
   chainId: number;
-  remainingAvailableNotional: number;
+  availableNotional: number;
   notionalLimit: number;
-  bigTransactionSize: number;
+  maxTransactionSize: number;
 }
 
 interface AvailableNotionalByChain {
-  entries: AvailableNotionalByChainEntry[];
+  data: AvailableNotionalByChainEntry[];
 }
 
 export interface ChainLimits {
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/hooks/useIsTransferLimited.ts
```diff
@@ -105,14 +105,12 @@ const useIsTransferLimited = (): IsTransferLimitedResult => {
...
-              rpcHost.endsWith('/') ? '' : '/'
-            }v1/governor`;
+            const baseUrl = `${rpcHost}${rpcHost.endsWith('/') ? '' : '/'}`;
             const [tokenListResponse, availableNotionalByChainResponse] =
               await Promise.all([
-                axios.get<TokenList>(`${baseUrl}/token_list`),
+                axios.get<TokenList>(`${baseUrl}v1/governor/token_list`),
                 axios.get<AvailableNotionalByChain>(
-                  `${baseUrl}/available_notional_by_chain`,
+                  `${baseUrl}api/v1/governor/limit`,
```
sebastianscatularo, line 113: rpcHost could be wormholescan or public guardian api, and `api/v1/governor/limit` is not compatible with guardian api, check here for instance [https://wormhole-v2-mainnet-api.mcf.rocks/api/v1/governor/limit](https://wormhole-v2-mainnet-api.mcf.rocks/api/v1/governor/limit). The fix should come from wormholescan API

sebastianscatularo (changes requested):

wormhole-connect/src/hooks/useIsTransferLimited.ts
sebastianscatularo: or alternatively we should add logic to handle the case, but I think that not a good path to follow

PR #2412, opened 2024-08-20
Remove `Route` enum; enable route plugins

artursapek:
- Removes `Route` enum type in Connect, 
  - Makes it possible to add arbitrary SDKv2 routes that Connect doesn't know about
- Changes `WormholeConnectConfig` interface to accept an array of SDKv2 `RouteConstructor`s
- Changes `RouteOperator` to be stateful, not a global singleton class
- Removes proxy pass-through methods from `RouteOperator`


wormhole-connect/package.json
```diff
@@ -17,6 +17,7 @@
     "dist"
   ],
   "dependencies": {
+    "@artursapek/vite-plugin-checker": "^0.7.3",
```
artursapek, line 20: a fork which contains this fix https://github.com/fi3ework/vite-plugin-checker/pull/389

wormhole-connect/package.json
```diff
@@ -154,6 +154,9 @@
     },
     "@wormhole-foundation/wormhole-connect": {
       "aptos": "1.5.0"
+    },
+    "vite-plugin-dts": {
```
artursapek, line 158: just some stupid dephell i had to resolve

wormhole-connect/src/routes/operator.ts
```diff
@@ -1,34 +1,59 @@
...
+  routes.AutomaticTokenBridgeRoute,
+  routes.TokenBridgeRoute,
+];
+
+export default class RouteOperator {
+  preference: string[];
+  routes: Record<string, SDKv2Route>;
+
+  constructor(routesConfig: routes.RouteConstructor<any>[] = DEFAULT_ROUTES) {
+    let routes = {};
```
kev1n-peters: Should use `const` here and below (linter errors)

wormhole-connect/src/hooks/useAvailableRoutes.ts
```diff
@@ -28,54 +26,51 @@ const useAvailableRoutes = (): void => {
     let isActive = true;
 
     const getAvailable = async () => {
-      const routes: RouteState[] = [];
-      for (const value of config.routes) {
-        const r = value as Route;
-
+      let routes: RouteState[] = [];
+      await config.routes.forEach(async (name, route) => {
```
kev1n-peters, line 30: I think we should use `Promise.all(routes.map(...))` or a `for ... of` loop to avoid issues calling async functions in `forEach`: https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971#do-not-use-foreach-with-async-await since

kev1n-peters (dismissed):

kev1n-peters (approved):

PR #2417, opened 2024-08-21
[Redesign] Fixing the UI to work with Mayan route

emreboga:
There are two main changes here:
- A temporary change to add Mayan in a static priority list of routes, so we can unblock the UI testing.
- Removing v2 hooks for source/dest token computations to continue using Redux state. We will revisit this question of redux vs. local state later on, but for now continue on the legacy pattern.


wormhole-connect/src/store/transferInput.ts
```diff
@@ -273,6 +273,7 @@ const establishRoute = (state: TransferInputState) => {
     Route.NttManual,
     Route.Relay,
     Route.Bridge,
+    Route.Mayan,
```
emreboga, line 276: This is a temporary change to unblock the UI testing as we eventually get rid of `routeOrderOfPreference` and rank order the routes according to ETA.

artursapek (approved):

PR #2419, opened 2024-08-22
Route priority sorting

nikarm22:
Prefetching quotes for routes.
Sorting Routes prioritizing:
1. Automatics over manuals
2. Faster over slower
3. Cheaper over expensive
In that priority order.


wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        const quote = result.value;
+        const receiveAmount = sdkAmount
+          .whole(quote.destinationToken.amount)
+          .toString();
+        const receiveNativeAmount = quote.destinationNativeGas
+          ? sdkAmount.whole(quote.destinationNativeGas)
+          : 0;
+        const eta = quote.eta ?? 0;
```
artursapek: I think using `0` for `eta` to indicate we don't know the ETA is error-prone because we'll eventually sort by these values. I don't like using zero-values like `0` or `""` to express an unknown value. Can we make `eta` optional and just make it `undefined` if we don't get one from the quote, and handle that correctly downstream?

See https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/connect/src/routes/types.ts#L80

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -0,0 +1,150 @@
...
+      !params.sourceChain ||
+      !params.sourceToken ||
+      !params.destChain ||
+      !params.destToken ||
+      !params.amount
+    ) {
+      return;
+    }
+
+    const promises = routes.map((route) =>
```
artursapek: I like the way you wrote this code, but I feel like it would be nice to move this part into a method on the `RouteOperator`, so this hook can be shorter and easier to read. `RouteOperator` is the class where we keep [other iterative logic like this](https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/routes/operator.ts#L63-L119).

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -0,0 +1,150 @@
...
+  sourceChain: Chain | undefined;
+  sourceToken: string;
+  destChain: Chain | undefined;
+  destToken: string;
+  amount: number;
+  nativeGas: number;
+};
+
+export type RouteQuote = {
+  error: string;
```
artursapek: It would be clearer to make `error` optional, and keep it `undefined` in the case where there is no error.
```suggestion
  error?: string;
```

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -0,0 +1,150 @@
...
+type RoutesQuotesBulkParams = {
+  sourceChain: Chain | undefined;
+  sourceToken: string;
+  destChain: Chain | undefined;
+  destToken: string;
+  amount: number;
+  nativeGas: number;
+};
+
+export type RouteQuote = {
```
artursapek: Why'd you choose to define a new type for this, rather than just using the [`Quote`](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/connect/src/routes/types.ts#L40) type that every SDK `Route` will give you? Is that one hard to work with because the numbers are not string-ified (or some other reason)?

wormhole-connect/src/views/v2/Bridge/Routes/index.tsx
```diff
@@ -52,11 +65,70 @@ const Routes = (props: Props) => {
     return routeStates.filter((rs) => rs.supported);
   }, [routeStates]);
 
+  const supportedRoutesNames = useMemo(
+    () => supportedRoutes.map((r) => r.name),
+    [supportedRoutes],
+  );
+
+  const { quotesMap, isFetching } = useRoutesQuotesBulk(supportedRoutesNames, {
+    amount: Number.parseFloat(amount),
```
artursapek: We can improve this. Here, for some reason [`computeQuote` expects a `number`](https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/routes/sdkv2/route.ts#L395) (I think because it was mimicking legacy code)... but all it does is [immediately call `toString()` on it](https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/routes/sdkv2/route.ts#L406).

Can you actually change that class so we are always dealing with string amounts? We should never be using floating point numbers for amounts - they are imprecise and not good for financial data. We use [`amount.Amount`](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/core/base/src/utils/amount.ts#L5) for this reason. Use strings or `amount.Amount`s everywhere.

wormhole-connect/src/views/v2/Bridge/Routes/index.tsx
```diff
@@ -52,11 +65,70 @@ const Routes = (props: Props) => {
...
+
+          // 3. Compare relay fees
+          if (quoteA.relayerFee > quoteB.relayerFee) {
+            return 1;
+          } else if (quoteA.relayerFee < quoteB.relayerFee) {
+            return -1;
+          }
+        }
+
+        // Don't swap when routes match by all criteria or don't have quotas
```
artursapek, line 135: If one has a quote and the other doesn't, shouldn't we prefer the one that does?

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
nikarm22: Fixed, keeping undefined for non existing value.

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
nikarm22: Moved multiCompute to `RouteOperator` and cleaned up parsing.

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
nikarm22: This makes much easier down the line for rendering, renamed to `ParsedQuote` for better readability, but using `routes.Quote` will repeat a lot of parsing in Components, hence the new type in the hook.

wormhole-connect/src/views/v2/Bridge/Routes/index.tsx
nikarm22: Added final priority, routes with a quote over w/o one.

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -0,0 +1,142 @@
...
+  relayerFee?: number;
+};
+
+type HookReturn = {
+  quotesMap: Record<string, ParsedQuote | undefined>;
+  isFetching: boolean;
+};
+
+const defaultQuote: ParsedQuote = {
+  error: '',
```
artursapek: `undefined`?

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -0,0 +1,142 @@
...
+  eta: 0,
```
artursapek: `undefined`?

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -0,0 +1,144 @@
...
+    ) {
+      return;
+    }
+
+    // Forcing TS to infer that fields are non-optional
+    const rParams = params as Required<RoutesQuotesBulkParams>;
+    const promises = config.routes.computeMultipleQuotes(routes, rParams);
+
+    setIsFetching(true);
+    Promise.allSettled(promises).then((results) => {
```
artursapek: IMO you could move more of this logic into the `RouteOperator` method. I also notice you didn't type the return value of `computeMultipleQuotes`. What do you think about this?

- move `ParsedQuote` type into `routes/operator.ts` and import it from there (since multiple components import it)
- move the parsing logic into there as well
- make it so the return type of `computeMultipleQuotes` is `Async<ParsedQuote[]>` so you can simply just `await` that and have your list of quotes in here.

artursapek (approved):

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -0,0 +1,82 @@
...
+    // Forcing TS to infer that fields are non-optional
+    const rParams = params as Required<RoutesQuotesBulkParams>;
+
+    setIsFetching(true);
+    config.routes
+      .computeMultipleQuotes(routes, rParams)
+      .then((quoteResults) => {
+        if (!unmounted) {
+          setQuotes(quoteResults);
+          setIsFetching(false);
```
emreboga, line 50: You need to set isFetching to false in an error case too. Like .finally()? Or even a try/catch to wrap the whole routes call.

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -0,0 +1,82 @@
...
+    params.nativeGas,
+  ]);
+
+  const quotesMap = useMemo(
+    () =>
+      routes.reduce((acc, route, index) => {
+        acc[route] = quotes[index];
+        return acc;
+      }, {} as Record<string, QuoteResult | undefined>),
+    [routes.join(), quotes],
```
emreboga, line 73: What's the difference to use `routes.join()` vs. `routes` itself here as a dependency?

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
nikarm22: This is left intentionally, as `Promise.allSettled` can't reject per [spec](https://tc39.es/ecma262/#sec-promise.allsettled). 
Can add a comment or add a finally block just for readability.

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
nikarm22: `routes` is an array, while joining it's becoming a string, essentially `useMemo` becomes agnostic of `routes` reference changes, and only re-calculates when actual value is changed, helps preserve `quote` references in the components that rely on them. It's a linear time operation, but implementation is native in browsers and is incredibly fast. (Actual string concatenation is executed in native runtime, not in JS VM)

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
emreboga: I see. Another thing is the promise callback comes here only once right? Shouldn't we call `setIsFetching(false)` outside the if statement?

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
emreboga: I see, so when it's an array/object dependencies are compared by reference:

> React will compare each dependency with its previous value using the [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison.

PR #2436, opened 2024-08-26
Programatically compute foreign asset decimals

kev1n-peters:
Foreign asset decimals can be computed programatically which eliminates the need to configure them.

- Removed DecimalsMap from TokenConfig
- Removed decimals from foreignAssets config
- Removed some dead code


artursapek:
oh lawd

wormhole-connect/src/utils/index.ts
```diff
@@ -121,16 +121,36 @@ export function getTokenDecimals(
...
+  if (platform === tokenPlatform) return decimals;
+
+  // Otherwise, return the minimum of the token's decimals and the platform's max decimals (token bridge)
+  // See: https://github.com/wormhole-foundation/wormhole/blob/main/whitepapers/0003_token_bridge.md#handling-of-token-amounts-and-decimals
+  const maxWrappedDecimals = {
+    Evm: 18,
+    Solana: 8,
+    Sui: 8,
+    Aptos: 8,
+  } as const;
```
artursapek: it be clearer to write this. "if platform is Evm, decimals is 18; otherwise decimals is 8"

did you not do that because eventually it's more complicated when we add Sui/Sei/Cosmos?

wormhole-connect/tests/ci/config.test.ts
```diff
@@ -35,6 +35,7 @@ envTests.forEach((env) => {
         if (!token.tokenId) {
           expect(token.wrappedAsset).toBeTruthy();
           const wrappedToken = TOKENS[token.wrappedAsset!];
+          console.log(token.key, 'wrapped', wrappedToken);
```
artursapek: did u mean to leave this in?

wormhole-connect/src/config/devnet/chains.ts
```diff
@@ -12,40 +12,6 @@ export const DEVNET_CHAINS: ChainsConfig = {
     gasToken: 'ETH',
     chainId: 1,
     icon: Icon.ETH,
-    automaticRelayer: false,
-    maxBlockSearch: 0,
-  },
-  Osmosis: {
```
artursapek, line 18: why are you removing these chains?

wormhole-connect/src/utils/index.ts
kev1n-peters: I'm not sure what the decimals is on Sei, Cosmos, or Near, but it will be a constant, probably 8 or 18. I could look them up

wormhole-connect/tests/ci/config.test.ts
kev1n-peters: No was debugging

wormhole-connect/src/config/devnet/chains.ts
kev1n-peters: They're not supported so I decided to remove them rather than comment them out. I could add them back if you want

wormhole-connect/src/utils/index.ts
kev1n-peters: It looks like wrapped assets are capped at 8 decimals except on EVM which caps them at 18

wormhole-connect/src/config/devnet/chains.ts
kev1n-peters: added them back

artursapek (approved):

PR #2438, opened 2024-08-27
Modernize config types

artursapek:
- get rid of legacy `Network` types, consolidate on new one from SDKv2
- rename `BridgeDefaults` properties to not use the term "network" in a confusing way


wormhole-connect/src/config/utils.ts
```diff
@@ -157,65 +157,65 @@ export const validateDefaults = (
...
     const requiredConfig = chains[defaults.requiredNetwork];
     if (!requiredConfig) {
       error(
         `Invalid network value "${defaults.requiredNetwork}" specified for bridgeDefaults.requiredNetwork`,
       );
     }
     if (
-      defaults.toNetwork !== defaults.requiredNetwork &&
-      defaults.fromNetwork !== defaults.requiredNetwork
+      defaults.toChain !== defaults.requiredNetwork &&
```
kev1n-peters: Should we rename `requiredNetwork` to `requiredChain`?

artursapek: ah... good catch thanks. will do

kev1n-peters (dismissed):

kev1n-peters (approved):

PR #2445, opened 2024-08-27
Filter out frankenstein tokens from token list

kev1n-peters:

PR #2461, opened 2024-08-28
redeem view should show token display name instead of key

kev1n-peters:


kev1n-peters:
depends on https://github.com/wormhole-foundation/wormhole-connect/pull/2445

PR #2463, opened 2024-08-28
Push equivalent wrapped token to top of token list

artursapek:
As observed by @tsadovska, adding the Mayan route causes wrapped assets to not show up in the destination token list. This PR makes two changes:

- still includes a wrapped destination token if it's the "equivalent token" for the source token (eg `wrappedAsset` key in `TokenConfig`)
- prioritizes this token to the top of the destination tokens list

On the first point, Mayan caused this bug because suddenly we had tokens native to the destination chain available (because Mayan supports several of them).

Before: 
image

After (notice WAVAX at the top of the list):
image


emreboga (approved):
LGTM!

PR #2466, opened 2024-08-28
use suiscan explorer

kev1n-peters:


wormhole-connect/src/utils/index.ts
```diff
@@ -334,3 +334,16 @@ export const isEmptyObject = (value: object | null | undefined) => {
 
   return true;
 };
+
+export const getExplorerUrl = (chain: Chain, address: string) => {
+  const chainConfig = config.chains[chain]!;
+  let explorerUrl = '';
+
+  if (chain === 'Sui') {
```
emreboga, line 342: Nitpick: Adding a condition here for a chain is kind of against the chain/token agnostic UI, but do you think there was an alternative? Like adding the `coin/` and `address/` paths to the chain configs?

kev1n-peters: If we extend this function in the future to support creating URLs for transactions, wallet addresses, etc. then we'll have to concatenate a different path anyway. I think the `explorerUrl` in the chain config should just be the "base URL"

kev1n-peters: here's how explorer URLs are constructed on old Connect for example: https://github.com/wormhole-foundation/wormhole-connect/blob/80625136db752d373ac8c1e0f26759c8b6514718/wormhole-connect/src/views/Redeem/ExplorerLink.tsx#L27

emreboga: I was thinking more of an optional prop in chain config, like `explorerUrlPath: "coin/"`, so can be appended if exists. This is fine, we can definitely revisit if we have more chain specific instances  👍

emreboga (approved):
LGTM!

PR #2469, opened 2024-08-28
[Redesign] Fixing the amount input and when chains/tokens are swapped and handling couple of error cases

emreboga:
- We need to clear the amount input when chains/tokens are swapped.
- We also need to add a couple of checks for error cases.

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { makeStyles } from 'tss-react/mui';
 import { useDebounce } from 'use-debounce';
@@ -66,7 +66,6 @@ const AmountInput = (props: Props) => {
     route,
   } = useSelector((state: RootState) => state.transferInput);
 
-  const [tokenAmount, setTokenAmount] = useState(amount);
   const [validationResult, setValidationResult] = useState('');
 
   // Debouncing validation to prevent false-positive results while user is still typing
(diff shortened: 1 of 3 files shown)


wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
```diff
@@ -66,7 +66,6 @@ const AmountInput = (props: Props) => {
     route,
   } = useSelector((state: RootState) => state.transferInput);
 
-  const [tokenAmount, setTokenAmount] = useState(amount);
```
emreboga, line 69: I've decided to remove the local state var and directly use amount in Redux . That's why I also moved the validation in useEffect into the amount change handler.

wormhole-connect/src/views/v2/Bridge/SwapInputs/index.tsx
```diff
@@ -33,12 +33,13 @@ function SwapInputs() {
     fromChain &&
     !config.chains[fromChain]?.disabledAsDestination &&
     toChain &&
-    !config.chains[toChain]?.disabledAsDestination;
+    !config.chains[toChain]?.disabledAsSource;
```
emreboga, line 36: I think this was a copy/paste error, or am I missing anything here?

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
```diff
@@ -224,6 +224,10 @@ const SingleRoute = (props: Props) => {
   }, [props.route.name]);
 
   const providerText = useMemo(() => {
+    if (!sourceToken) {
```
emreboga, line 227: This is a very rare case (I've seen it when switching chains), but good to be safe.

wormhole-connect/src/views/v2/Bridge/SwapInputs/index.tsx
artursapek: came from here https://github.com/wormhole-foundation/wormhole-connect/commit/5c13b8d1aea40d1ee896deb4c0e7a375cffdbb02#diff-50062b7f121564dcafd21f3ee9350314aae20a3a4b2487eff668042c0d358587R36

artursapek (approved):

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
artursapek: came from here https://github.com/wormhole-foundation/wormhole-connect/commit/5c13b8d1aea40d1ee896deb4c0e7a375cffdbb02#diff-50062b7f121564dcafd21f3ee9350314aae20a3a4b2487eff668042c0d358587R36

Issue #2486, opened 2024-08-29
[UI-Feature] Filter out all Wormhole-wrapped assets in asset picker


Wesleyleung:
## Description and context

Wormhole-wrapped assets should be filtered out from the asset picker except when the user has had their wallet connected and they have a balance of the specific Wormhole-wrapped asset. Otherwise, these assets should not appear. 

Cases:
1. User's wallet is not connected:
- User selects a network. 
- Only native tokens like stables, gas tokens, or whatever tokens are configured for that integrators are shown. 
- No wormhole-wrapped assets appear. For example, even if the user searches "USDC" they will only see the native USDC or USDC.e for the relevant chain, not Wormhole-wrapped USDC.

2. User's wallet is connected and they have a balance of the Wormhole-wrapped asset in their wallet
- show the relevant asset in the asset picker for both "To" and "From"
- if the user selected the network where the Wormhole-wrapped asset originated from that would unwrap the token, show the native unwrapped version  (eg. Wormhole-wrapped Sui on Solana. User selected Sui as the network, then they will see Sui as the output)

3. User's wallet is connected and they do not have any Wormhole-wrapped balances
- do not show any Wormhole-wrapped assets in "From"
- the user can still mint new Wormhole-wrapped assets by choosing a native asset and then bridging it to a new chain using tokenbridge.

PR #2492, opened 2024-08-29
[UI-Feature] Filter out all Wormhole-wrapped assets in asset picker #…

kev1n-peters:
…2486

Addresses #2486

### DIFF
--- a/wormhole-connect/src/utils/index.ts
+++ b/wormhole-connect/src/utils/index.ts
@@ -373,6 +373,10 @@ export const isFrankensteinToken = (token: TokenConfig, chain: Chain) => {
   return (
     nativeChain !== chain &&
     !(['Ethereum', 'Sepolia'] as Chain[]).includes(nativeChain) &&
-    ['ETH', 'WETH', 'wstETH', 'USDT', 'USDC'].includes(symbol)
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/utils/index.ts
```diff
@@ -373,6 +373,10 @@ export const isFrankensteinToken = (token: TokenConfig, chain: Chain) => {
   return (
     nativeChain !== chain &&
     !(['Ethereum', 'Sepolia'] as Chain[]).includes(nativeChain) &&
-    ['ETH', 'WETH', 'wstETH', 'USDT', 'USDC'].includes(symbol)
+    ['ETH', 'WETH', 'wstETH', 'USDT', 'USDC', 'USDC.e'].includes(symbol)
   );
 };
+
+export const isWrappedToken = (token: TokenConfig, chain: Chain) => {
+  return token.nativeChain !== chain;
```
emreboga, line 381: @kev1n-peters are all non-native tokens considered wrapped tokens?

emreboga (approved):

wormhole-connect/src/utils/index.ts
kev1n-peters: that's right - if a token is not native to the current chain, then it's wormhole-wrapped in Connect

PR #2495, opened 2024-08-30
fix: event handler config overrides by parsedConfig

jovinlidan:
Fix issue where event handler configurations were being overridden by parsedConfig when users supplied configurations in the hosted build


artursapek (approved):
Nice catch

PR #2506, opened 2024-08-30
Fetch token prices excess network calls / rate limited fix

kev1n-peters:
We should fetch the token prices once when mounted and refresh them periodically. Doing this instead of fetching them on demand in multiple components helps prevent us from getting rate limited and making redundant network calls.


wormhole-connect/src/views/v2/Redeem/TransactionDetails/index.tsx
```diff
@@ -91,7 +91,7 @@ const TransactionDetails = () => {
             {amount} {sourceTokenConfig.symbol}
           </Typography>
           <Typography color={theme.palette.text.secondary} fontSize={14}>
-            {isFetchingTokenPrices ? (
+            {tokenPrices.isFetching ? (
```
artursapek, line 96: I don't think any code actually sets this `isFetching` boolean to true anymore. From what I can tell it's set using `setFetchingPrices` in `stores/tokenPrices.ts` but I don't see that function called anywhere. Maybe we can default `isFetching` to `true` inside the `initialState` value?

wormhole-connect/src/views/v2/Bridge/index.tsx
```diff
@@ -184,6 +185,9 @@ const Bridge = () => {
   // Call to initiate transfer inputs validations
   useValidate();
 
+  // Fetch token prices
+  useFetchTokenPrices();
```
artursapek, line 189: Nice improvement calling this hook only in the top level here.

wormhole-connect/src/views/v2/Redeem/TransactionDetails/index.tsx
kev1n-peters: I decided to call `dispatch(setFetchingPrices())` once in `fetchTokenPrices` instead, wdyt

artursapek: yeah I think once at the beginning makes sense!

artursapek (approved):

PR #2508, opened 2024-08-30
Backport USDT Bridge route tokens

kev1n-peters:
Mayan supports these


artursapek (approved):
double checked addresses for u, lgtm :)

PR #2516, opened 2024-09-03
Don't disable chains; Connect already handles wallet switching

artursapek:
This removes the behavior that a chain is disabled if you have a wallet connected for a different platform. Connect already can disconnect the wallet (and reconnect to the right wallet if you have previously connected one).

This was frustrating/confusing because the new design menu auto-selects Ethereum when I open it, which then immediately made Solana and other non-EVM chains disabled right away.

https://github.com/user-attachments/assets/00133467-fbf9-44c4-b8c5-cd2282d71b1d


kev1n-peters (approved):

emreboga (approved):
LGTM!

PR #2521, opened 2024-09-03
config: Replaced `TokenConfig.foreignAssets` with `wrappedTokens` config

kev1n-peters:
Wormhole-wrapped token addresses are specific to the Token Bridge and should not be a property of `TokenConfig`. Introduced a new top-level config, `wrappedTokens`, which is a map keyed by token keys and contains chain-to-wormhole-wrapped token addresses.


wormhole-connect/src/config/index.ts
```diff
@@ -133,10 +142,10 @@ export function buildConfig(
         : true;
     }),
 
-    // For token bridge ^_^
+    // For token bridge =^_^=
```
artursapek, line 146: lmfao

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -94,7 +94,10 @@ const TokenList = (props: Props) => {
             (t) =>
               t.key === destTokenKey &&
               // Only add the wrapped token if it actually exists on the destination chain
-              !!t.foreignAssets?.[selectedChainConfig.key],
+              !!config.wrappedTokenAddressCache.get(
```
artursapek: same comment here

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenItem.tsx
```diff
@@ -53,7 +54,9 @@ function TokenItem(props: TokenItemProps) {
   const address =
     chain === token.nativeChain
       ? token.tokenId?.address
-      : getWrappedToken(token)?.foreignAssets?.[chain];
```
artursapek, line 56: so we're no longer using `getWrappedToken` which comes from `utils`... but I feel like it would be a good idea to maintain that method or one like it and hiding the implementation detail of `config.wrappedTokenAddressCache` existing.

we have [`getTokenBridgeWrappedTokenAddress`](https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/utils/sdkv2.ts#L65) in `utils/sdkv2.ts` but it's async... maybe we could add one like it that's synchronous and only hits this cache.

I just think it would be nice to have a util for this and not have to reference this cache (again, an implementation detail) in our React code like this

kev1n-peters: agree that we should hide impl details

artursapek (approved):

PR #2531, opened 2024-09-04
Resume transfer token address parsing bug fix

kev1n-peters:
We were not using the token's origin chain when converting its universal address to native format. This was preventing us from resuming transfers for certain tokens.


kev1n-peters:
bug that was happening before when resuming a TB transfer of SOL from Solana 
image

artursapek (approved):

PR #2542, opened 2024-09-05
Redesign wallet above asset picker

nikarm22:
Implementation is somewhere in between design and current one. (Preserved arrow and wallet icon)

UI:
Screenshot from 2024-09-05 20-17-47


emreboga (approved):
LGTM!

PR #2545, opened 2024-09-05
[Redesign] Transaction History

emreboga:
This PR adds the Transaction history view, currently displaying token bridge, NTT and CCTP transactions.

Please see the loom for visuals and more details:
https://www.loom.com/share/1478e981b0fc4b069c1db0dbc6fb5c05?sid=b4877189-43d6-476a-9be1-30368f3f4f9f

### DIFF
--- a/wormhole-connect/src/config/index.ts
+++ b/wormhole-connect/src/config/index.ts
@@ -102,22 +102,22 @@ export function buildConfig(
     ),
     graphql: Object.assign({}, networkData.graphql, customConfig?.graphql),
     wormholeApi: {
-      mainnet: 'https://api.wormholescan.io/',
-      testnet: 'https://api.testnet.wormholescan.io/',
-      devnet: '',
+      Mainnet: 'https://api.wormholescan.io/',
+      Testnet: 'https://api.testnet.wormholescan.io/',
+      Devnet: '',
     }[network],
     wormholeRpcHosts: {
-      mainnet: [
+      Mainnet: [
         'https://wormhole-v2-mainnet-api.mcf.rocks',
         'https://wormhole-v2-mainnet-api.chainlayer.network',
         'https://wormhole-v2-mainnet-api.staking.fund',
       ],
-      testnet: [
+      Testnet: [
         'https://guardian.testnet.xlabs.xyz',
         'https://guardian-01.testnet.xlabs.xyz',
         'https://guardian-02.testnet.xlabs.xyz',
       ],
-      devnet: ['http://localhost:7071'],
+      Devnet: ['http://localhost:7071'],
     }[network],
     coinGeckoApiKey: customConfig?.coinGeckoApiKey,
 
--- a/wormhole-connect/src/hooks/useFetchTransactionHistory.ts
+++ b/wormhole-connect/src/hooks/useFetchTransactionHistory.ts
@@ -1,8 +1,8 @@
 import { useCallback, useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import {
-  chainIdToChain,
   amount as sdkAmount,
+  chainIdToChain,
   resolveWrappedToken,
 } from '@wormhole-foundation/sdk';
 
@@ -11,7 +11,7 @@ import config from 'config';
 import type { RootState } from 'store';
 import type { Chain } from '@wormhole-foundation/sdk';
 import type { RelayerFee } from 'store/relay';
-import { getTokenDecimals, getWrappedTokenId } from 'utils';
+import { getTokenById, getTokenDecimals, getWrappedTokenId } from 'utils';
 
 export interface Transaction {
   // Transaction hash
@@ -30,10 +30,10 @@ export interface Transaction {
   // Source token address
   tokenAddress: string;
   tokenKey: string;
-  tokenDecimals: number;
+  tokenDecimals?: number;
 
   // Destination token
-  receivedTokenKey: string;
+  receivedTokenKey?: string;
   receiveAmount?: string;
   relayerFee?: RelayerFee;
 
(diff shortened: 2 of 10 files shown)


wormhole-connect/src/hooks/useFetchTransactionHistory.ts
```diff
@@ -0,0 +1,243 @@
...
+    const receivedTokenKey =
+      tokenConfig.nativeChain === toChain
+        ? tokenConfig.key
+        : getWrappedToken(tokenConfig)?.key;
+
+    // Reverse engineering the conversion rate:
+    //   data.tokenAmount is the actual value and standarizedProperties.amount is the denomination.
+    //   There has been unexpected conversaion rates in API responses. Therefore we are
+    //   sniffing out the decimals from the conversion rate of between these two values.
+    const conversionRate =
```
emreboga: I'm not 100% sure if this is a good way to get the decimals. TBH the data API returns changes from 6 to 9 decimals. What would be the best way to calculate these decimals for a chain/token?

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
```diff
@@ -0,0 +1,243 @@
...
+  const [isFetching, setIsFetching] = useState(false);
+  const [hasMore, setHasMore] = useState(true);
+
+  const { sending: sendingWallet } = useSelector(
+    (state: RootState) => state.wallet,
+  );
+
+  const { page = 0, pageSize = 30 } = props;
+
+  const parseTokenBridgeTx = (tx) => {
```
artursapek: can we type this? can we create an interface for the data structure that Wormholescan API gives us?

it's nicer to put the types into the function signature and just return the right object

```ts
const parseTokenBridgeTx = (tx: WormholeScanTransaction): Transaction => {

  ...

  return {
    txHash: sourceChain.transaction?.txHash,
    ...
  }
}
```

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
artursapek: We have `getTokenDecimals` in `utils` which can give you the decimals for a given token. This is just based on which chain it's on. Does this work?

```ts
const tokenId = {
  chain: fromChain,
  address: standarizedProperties.tokenAddress,
};
const tokenConfig = getTokenById(tokenId);
const tokenDecimals = getTokenDecimals(tokenId);
const { amount, fee } = standardizedProperties;
const receiveAmount = BigInt(amount) - BigInt(fee);
const receiveAmountDisplay = sdkAmount.display({
  amount: amountOutBaseUnits.toString(),
  decimals: tokenDecimals
});
```

We should use BigInts for arithmetic whenever possible

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
```diff
@@ -0,0 +1,243 @@
...
+      appIds.forEach((appId) => {
+        if (typeof SUPPORTED_APPIDS[appId] === 'function') {
+          txParser = SUPPORTED_APPIDS[appId];
+        }
+      });
+
+      return txParser?.(tx);
+    });
+
+    return parsedTxs.filter((tx) => !!tx);
```
artursapek: Feel free to take this or leave it, but you could make this logic easier to read and run a little faster

```suggestion
    return allTxs.map((tx) => {
      const appIds = tx.content?.standarizedProperties?.appIds || [];
      for (const appId of appIds) {
        const parser = SUPPORTED_APPIDS[appId];
        if (parser) return parser(tx);
      }
    }).filter((tx) => !!tx);
```

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
```diff
@@ -0,0 +1,243 @@
...
+        data = await res.json();
+      } catch (error) {
+        if (!cancelled) {
+          setError(`Error fetching transaction history: ${error}`);
+        }
+
+        setIsFetching(false);
+      }
+
+      if (!cancelled) {
```
artursapek: IMO it's confusing how this block runs even when we caught an error above. can you return early in the `catch` or maybe put this under the `try`?

If you put this under the `try`, you could call `setIsFetching(false);` for both cases by putting it inside a `finally`

```ts
try {
  // happy path
} catch (e) {
  // sad path
} finally {
  setIsFetching(false);
}
```

wormhole-connect/src/views/v2/TxHistory/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+              <CircularProgress />
+            </Stack>
+          }
+          loadMore={() => setPage((page) => page + 1)}
+          pageStart={0}
+          useWindow={false}
+        >
+          <div className={joinClass([classes.spacer])}>
+            {transactions.map((tx) => {
+              return <TxHistoryItem data={tx} tokenPrices={tokenPrices.data} />;
```
kev1n-peters: Should have a `key` prop since we're rendering a list of components

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
emreboga: @artursapek that gives me the decimals in token config, which are usually larger than the actual. I can use bigInt type, but the second value is decimal, so it will be a bit more tricky. Let me give it a try.

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
emreboga: I agree, helps with the productivity as well.

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
emreboga: @artursapek I think this one seems to be working:
```
const decimals = Math.min(
      8,
      getTokenDecimals(
        fromChain,
        tokenConfig.nativeChain === fromChain ? 'native' : tokenConfig.tokenId,
      ),
    );
```

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
emreboga: I'll make this more simple 👍

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
emreboga: If we caught and error inside try/catch, data will be falsy and this block won't run. I wanted like to keep try/catch minimal to the things that we actually want to monitor. That being said I can apply the recommended pattern to increase readability.

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
```diff
@@ -0,0 +1,310 @@
...
+  // Amount of native gas being received, in destination gas token units
+  // For example 1.0 is 1.0 ETH, not 1 wei
+  receiveNativeAmount?: number;
+
+  // Timestamps
+  senderTimestamp?: string;
+  receiverTimestamp?: string;
+}
+
+interface WormholeScanTransaction {
```
emreboga, line 45: I don't think we need to export as we don't need to use it anywhere else (unless we decide to move out the parsing logic to somewhere else in the future).

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
```diff
@@ -0,0 +1,310 @@
...
+    transaction: {
+      txHash: string;
+    };
+    from: string;
+    status: string;
+    fee: string;
+    gasTokenNotional: string;
+    feeUSD: string;
+  };
+  targetChain?: {
```
emreboga, line 88: `targetChain` seems to be the only optional property.

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
artursapek: yeah data being falsy is what I'm saying makes this a little hard to read. I'm not hung up on this one, just a comment. you can leave it as is if you prefer.

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
artursapek: OK. FYI, the string `'native'` is meant to represent the native gas token on a chain. I don't think you are using it correctly here; `tokenConfig.nativeChain` just represents which chain the token is native to; it doesn't indicate that the token is the _native gas token_. It's kind of confusing.

Maybe a dumb question, but is it possible Wormholescan's API just always uses decimals of `8`?

artursapek: Also, the comment about BigInt is that it's just better to use BigInt when subtracting the fee from the original amount, to get the receive amount. We're working with base units which are integers so we should avoid using the default `Number`, which is a floating point, to do arithmetic with base units.

emreboga: It might be just 8 for all decimals from the API. I've already asked about this and I believe we can get a confirmation on Monday 👍

emreboga: I've also pushed a change to use BigInt for operations.

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
```diff
@@ -0,0 +1,333 @@
...
+        );
+
+        const data = await res.json();
+
+        if (!cancelled) {
+          if (data?.operations?.length > 0) {
+            setTransactions((txs) => {
+              const parsedTxs = parseTransactions(data.operations);
+              if (txs && txs.length > 0) {
+                return txs.concat(parsedTxs);
```
artursapek: It's unlikely, but possible that a user would be infinite scrolling while one of their transactions gets indexed. This would result in every entry in their history being bumped down by 1, and possibly cause you to get a duplicate entry as you page through the history. You could avoid this by de-duplicating them by some kind of unique ID, like the source tx hash.

artursapek (dismissed):

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
emreboga: Thanks for pointing this out! Yes, it's unlikely and also if ever happens it'll be a duplicate transactions instead of a missing one, which is not too bad. But all that being said, let me add one more iteration over the list of transactions to ensure its uniqueness 👍

wormhole-connect/src/config/index.ts
```diff
@@ -102,22 +102,22 @@ export function buildConfig(
     ),
     graphql: Object.assign({}, networkData.graphql, customConfig?.graphql),
     wormholeApi: {
-      mainnet: 'https://api.wormholescan.io/',
-      testnet: 'https://api.testnet.wormholescan.io/',
-      devnet: '',
+      Mainnet: 'https://api.wormholescan.io/',
```
emreboga, line 105: I just realized these have recently changed to first letter being capital.

artursapek: yes. we switched to the new SDK's type for this recently. :+1:

wormhole-connect/src/hooks/useFetchTransactionHistory.ts
```diff
@@ -296,7 +296,14 @@ const useFetchTransactionHistory = (
             setTransactions((txs) => {
               const parsedTxs = parseTransactions(data.operations);
               if (txs && txs.length > 0) {
-                return txs.concat(parsedTxs);
+                const uniqList = {};
+                txs.concat(parsedTxs).forEach((tx) => {
+                  if (tx?.txHash) {
+                    uniqList[tx.txHash] = tx;
+                  }
+                });
```
artursapek, line 305:
```ts
const existingEntries = new Set();
for (const tx of txs) {
  existingEntries.add(tx.txHash);
}
return txs.concat(parsedTxs.filter((newTx) => !existingEntries.has(newTx.txhash)));
```

artursapek (dismissed):

artursapek (approved):

PR #2547, opened 2024-09-06
Run amount validation on token/netowork/wallet changes.

nikarm22:
[Ticket](https://www.notion.so/wormholelabs/UI-Bug-Errors-need-to-dynamically-update-based-on-new-inputs-c1dbb35abc3b41aaa198ac20d2fd4117?pvs=23)

PR #2551, opened 2024-09-09
[Redesign] Update amount only when it's valid to prevent extra api calls

emreboga:
I'm adding the local-state for amount input to set the redux state only when the amount in valid. This will prevent the unnecessary API calls.

Fixes https://www.notion.so/wormholelabs/BUG-Unnecessary-API-requests-96c28242a89449a18872be917479784c?pvs=4


emreboga:
I've added the fix from https://github.com/wormhole-foundation/wormhole-connect/pull/2547 as well. They were going to conflict otherwise.

kev1n-peters (approved):

PR #2554, opened 2024-09-09
Fix token amount decimals in final view.

nikarm22:
UI: 
Screenshot from 2024-09-09 20-14-42

### DIFF
--- a/wormhole-connect/src/store/transferInput.ts
+++ b/wormhole-connect/src/store/transferInput.ts
@@ -49,6 +49,13 @@ export const formatBalance = (
   });
 };
 
+export const formatStringAmount = (amountStr = '0'): string => {
+  const amountNum = parseFloat(amountStr);
+  return amountNum.toLocaleString('en', {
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/store/transferInput.ts
```diff
@@ -49,6 +49,13 @@ export const formatBalance = (
   });
 };
 
+export const formatStringAmount = (amountStr = '0'): string => {
+  const amountNum = parseFloat(amountStr);
+  return amountNum.toLocaleString('en', {
```
emreboga, line 54: You can also use [toFixed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) instead of locale formatting.

emreboga: Actually, it's not the same, you are looking for maximum decimals, not fixed.

emreboga (approved):
LGTM!

PR #2555, opened 2024-09-09
Mayan USD size limit & bug fixes in `Routes`/`SingleRoute`

artursapek:
- Adds temporary $1000 USD limit to Mayan Swift txs
- Fixes route auto-selection: don't auto-select a route if we got an error trying to fetch a quote
- Fixes some visual treatment of routes with a quote error
- Remove preference of CCTP over Mayan

Based off https://github.com/wormhole-foundation/wormhole-connect/pull/2535 - which needs to merge first

image

### DIFF
--- a/wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
+++ b/wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
@@ -1,6 +1,9 @@
 import { useState, useEffect, useMemo } from 'react';
+import { useSelector } from 'react-redux';
+import type { RootState } from 'store';
 import { Chain, routes } from '@wormhole-foundation/sdk';
 import { QuoteParams } from 'routes/operator';
+import { calculateUSDPriceRaw } from 'utils';
 
 import config from 'config';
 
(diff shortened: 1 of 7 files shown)


wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -24,6 +27,16 @@ const useRoutesQuotesBulk = (routes: string[], params: Params): HookReturn => {
   const [isFetching, setIsFetching] = useState(false);
   const [quotes, setQuotes] = useState<QuoteResult[]>([]);
 
+  // TODO temporary
+  // Calculate USD amount for temporary $1000 Mayan limit
```
emreboga, line 31: I think this comment belongs to line 85 below?

emreboga (approved):
One minor comment, rest is LGTM!

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
artursapek: it belongs in both places. I'll copy it

PR #2556, opened 2024-09-09
add phantom and backpack evm wallet explicitly

sebastianscatularo:
Add support fro eip 6963 on wallet aggregator to allow multiple evm wallets to work at the same time

depends on: https://github.com/XLabs/wallet-aggregator-sdk/pull/62

https://github.com/user-attachments/assets/38c7cac4-6135-415c-81e8-63181c10fa8a


yuli-ferna (approved):

wormhole-connect/src/utils/wallet/index.ts
```diff
@@ -339,6 +339,10 @@ const mapWallets = (
   skip: string[] = [],
 ): WalletData[] => {
   return Object.values(wallets)
+    .filter(
+      (wallet, index, self) =>
+        index === self.findIndex((o) => o.getName() === wallet.getName()),
```
emreboga, line 344: @sebastianscatularo I don't think I've fully understood this filter :-/
Are we filtering out the ones that doesn't match the name at the same index from the iterator?

sebastianscatularo: we are mixing two wallet detection mechanism, so, both could detect the same wallet, to avoid duplicated, we are filtering out. 

if you check at https://github.com/wormhole-foundation/wormhole-connect/pull/2556/files#diff-db002acd3260c36c2dfcc51bbbe8be8b0edf626feebb296bc3ff35b8b1860eb9R19 
you can see that we are listing explicit the wallets, so, I've added Injected connector for the wallets that are detected by injected, and are not listed explicit

emreboga (approved):
LGTM!

PR #2558, opened 2024-09-10
Resume CCTP transfers to Solana fix

kev1n-peters:
If the ATA didn't exist then it wasn't possible to resume a manual CCTP transfer to Solana.


wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -273,6 +274,13 @@ const Redeem = () => {
   // Checks whether the receiving wallet is currently connected
   const isConnectedToReceivingWallet = useMemo(() => {
     if (!recipient) {
+      // For Solana transfers, the associated token account (ATA) might not exist,
+      // preventing us from retrieving the recipient wallet address.
+      // In such cases, when resuming transfers, we allow the user to connect a wallet
+      // to claim the transfer, which will create the ATA.
+      if (isResumeTx && toChain === 'Solana' && receivingWallet.address) {
```
artursapek: so if they connect to the wrong Solana wallet, are you just relying on the transaction failing?

wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -285,7 +293,7 @@ const Redeem = () => {
       walletAddress === walletCurrentAddress &&
       walletAddress === recipientAddress
     );
-  }, [receivingWallet, recipient]);
+  }, [receivingWallet, recipient, toChain]);
```
artursapek: do you need to add `isResumeTx` to this list?

wormhole-connect/src/views/v2/Redeem/index.tsx
kev1n-peters: yeah, if the wrong wallet is connected then the simulation will fail. We don't have a way to determine the correct wallet address since the ATA can't be mapped back to it

wormhole-connect/src/views/v2/Redeem/index.tsx
kev1n-peters: fixed

wormhole-connect/src/views/v2/Redeem/index.tsx
kev1n-peters: fwiw this scenario (resuming a CCTP transfer to Solana where the ATA doesn't exist) is likely a rare edge case

kev1n-peters: Going to make this more robust and warn the user if the wrong wallet is connected (we can derive the ATA and see if it matches what's on the VAA)

artursapek: yeah I was gonna suggest that but maybe it's not necessary.

artursapek: I'd like to avoid having Connect ever look at VAAs 😬

artursapek (approved):

PR #2561, opened 2024-09-10
[Redesign] Adding Dark theme colors and other styling fixes

emreboga:
This PR:
- Adds the [palette colors](https://www.figma.com/design/CdUgpewRdHY1YkZMxSIteh/WH-Connect-Redesign-Q2-2024?node-id=1626-18212&node-type=frame&t=eCH5ZE3MeZ8zNRUV-0) to the dark theme (which is our default theme).
- Replaces any hardcoded color with the theme counterpart.
- Also has several UX fixes.

Screenshot 2024-09-10 at 8 32 27 PM

Screenshot 2024-09-10 at 8 32 38 PM

Screenshot 2024-09-10 at 8 28 12 PM

Screenshot 2024-09-10 at 8 40 28 PM

### DIFF
--- a/wormhole-connect/src/components/v2/Button.tsx
+++ b/wormhole-connect/src/components/v2/Button.tsx
@@ -3,28 +3,24 @@ import { styled } from '@mui/material';
 import { default as MUIButton, ButtonProps } from '@mui/material/Button';
 
 const PrimaryButton = styled(MUIButton)<ButtonProps>(({ theme }) => ({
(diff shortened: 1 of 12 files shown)


Wesleyleung:
I still like #C1BBF6 as a themed default purple though vs the more neon version here. Is there a way to cleanly override the purple used in the palette such that`primary.main` renders as #C1BBF6 like the below? 

Screenshot 2024-09-10 at 4 15 12 PM

artursapek (approved):

PR #2562, opened 2024-09-10
Approximate fee for Mayan Swift quotes

artursapek:
Adds special logic for Mayan Swift xfers that calculates a USD-denominated fee. Also hides the token symbol for Mayan quotes' network cost, because we're using USDC to represent dollars.

image

Depends on https://github.com/mayan-finance/wormhole-sdk-route/pull/40

### DIFF
--- a/wormhole-connect/src/components/DemoApp/index.tsx
+++ b/wormhole-connect/src/components/DemoApp/index.tsx
@@ -53,7 +53,7 @@ const parseConfig = (config: string): WormholeConnectConfig => {
       /* @ts-ignore */
       window.TokenBridgeRoute = routes.TokenBridgeRoute;
       /* @ts-ignore */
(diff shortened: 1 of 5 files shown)


wormhole-connect/src/components/DemoApp/index.tsx
```diff
@@ -53,7 +53,7 @@ const parseConfig = (config: string): WormholeConnectConfig => {
       /* @ts-ignore */
       window.TokenBridgeRoute = routes.TokenBridgeRoute;
       /* @ts-ignore */
-      window.ManualCCTPRoute = routes.ManualCCTPRoute;
+      window.CCTPRoute = routes.CCTPRoute;
```
artursapek, line 56: just fixing mistake I made before which I didnt notice

kev1n-peters (approved):

PR #2570, opened 2024-09-11
finalize transfers support

kev1n-peters:
the redeem page allows transfers that are queued on the destination chain to be completed after the delay

currently only the ntt routes use this state

### DIFF
--- a/wormhole-connect/src/views/v2/Redeem/index.tsx
+++ b/wormhole-connect/src/views/v2/Redeem/index.tsx
@@ -6,7 +6,13 @@ import Box from '@mui/material/Box';
 import CircularProgress from '@mui/material/CircularProgress';
 import Stack from '@mui/material/Stack';
 import Typography from '@mui/material/Typography';
-import { isRedeemed, routes, TransferState } from '@wormhole-foundation/sdk';
+import {
+  isCompleted,
+  isDestinationQueued,
+  isRedeemed,
+  routes,
+  TransferState,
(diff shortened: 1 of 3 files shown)


kev1n-peters:
image

wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -461,7 +500,12 @@ const Redeem = () => {
 
   // Main CTA button which has separate states for automatic and manual claims
   const actionButton = useMemo(() => {
-    if (!isTxComplete && !isTxRefunded && !isTxFailed && !isAutomaticRoute) {
+    if (
+      !isTxComplete &&
+      !isTxRefunded &&
+      !isTxFailed &&
+      (isTxDestQueued || !isAutomaticRoute)
```
artursapek, line 507: maybe this would be easier to read if we checked for what we _do_ expect it to be, rather than what we do not expect it to be?

isn't it basically just `(state === TransferState.Attested && !isAutomatic) || isTxDestQueued`?

I realize this was already like this before your change... just thinking out loud here.

artursapek (approved):

wormhole-connect/src/views/v2/Redeem/index.tsx
kev1n-peters: i agree that this is ugly, but going to leave it for now since we know that it works. we should probably refactor the redeem view in the future as it's becoming quite complex

PR #2574, opened 2024-09-11
frankenstein token fixes

kev1n-peters:
There are a number of places that call `addToken` so the frankenstein and wrapped token checks should be moved into it

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
@@ -74,6 +74,24 @@ const TokenList = (props: Props) => {
     const tokens: Array<TokenConfig> = [];
 
     const addToken = (tokenConfig: TokenConfig) => {
+      // Exclude frankenstein tokens with no balance
+      const balance = Number(balances?.[tokenConfig.key]?.balance);
+      if (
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -74,6 +74,24 @@ const TokenList = (props: Props) => {
...
+      if (
+        isFrankensteinToken(tokenConfig, selectedChainConfig.key) &&
+        !balance
+      ) {
+        return;
+      }
+
+      // Exclude wormhole-wrapped tokens with no balance
+      if (
+        props.isSource &&
```
artursapek, line 88: is this backwards? did you mean to hide them in the destination list? the source list should already not show tokens without a balance.

kev1n-peters: No this is right, we want to be able to select wormhole-wrapped tokens on the destination chain without a balance (otherwise we can't transfer them). We just disable these tokens when they have 0 balance (but they're still shown). This makes it so that they're not even shown

artursapek: got it

artursapek (approved):

PR #2584, opened 2024-09-12
frankenstein fixes

kev1n-peters:
a wormhole-wrapped token on a chain where there is a native representation of that same token is a frankenstein token, even if it's brided from Ethereum (was a a special case before)


emreboga (approved):
LGTM!

PR #2587, opened 2024-09-12
search tokens by address

kev1n-peters:


artursapek (dismissed):
nice

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -179,31 +179,34 @@ const TokenList = (props: Props) => {
...
 
         const queryLC = query.toLowerCase();
-        return Boolean(
-          token.symbol?.toLowerCase().includes(queryLC) ||
-            getDisplayName(token, props.selectedChainConfig.key)
-              .toLowerCase()
-              .includes(queryLC),
-        );
+
+        const symbolMatch = token.symbol?.toLowerCase().includes(queryLC);
```
emreboga, line 197: Nitpick: If there is a symbolMatch, you can return here without checking other matches. Same as displayNameMatch below. However, if you think having one return at the bottom for all cases makes the code more organized and readable, I'm all for that too 👍

kev1n-peters: Change to make the code more performant. I don't think readability suffers

emreboga (approved):
LGTM!

PR #2591, opened 2024-09-13
[Redesign] Merging Mayan transactions into Transaction History view

emreboga:
This PR adds the hook to fetch and parse Mayan transactions as well as the logic to merge them with WHScan transactions into the Transaction History view.

Please see the loom here:
https://www.loom.com/share/3b849138a5e44adab4461a3add8fd2bb

### DIFF
--- a/wormhole-connect/src/hooks/useTransactionHistory.ts
+++ b/wormhole-connect/src/hooks/useTransactionHistory.ts
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 
 import useTransactionHistoryWHScan from 'hooks/useTransactionHistoryWHScan';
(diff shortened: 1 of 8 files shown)


wormhole-connect/src/hooks/useTransactionHistory.ts
```diff
@@ -0,0 +1,193 @@
...
+  const { page = 0, pageSize = 30 } = props;
+
+  // Keeping separate indexes to track the last rendered item in respective transaction sets
+  const [mayanIndex, setMayanIndex] = useState(0);
+  const [whScanIndex, setWHScanIndex] = useState(0);
+
+  // We need to keep the last fetched pages from each APIs
+  // as the number of items in a page from each API can be different when sorted by time.
+  const [mayanPage, setMayanPage] = useState(page);
+  const [whScanPage, setWHScanPage] = useState(page);
```
artursapek, line 32: it seems like you could reduce the repetitive code here a lot by creating some kind of `HistoryFetcher` utility class that stores its page and other necessary state internally. Ideally we would have code that can handle any arbitrary number of merge-able histories.

I do want to move all of this into SDK in the not too distant future; each `Route` should be able to return transaction history. So perhaps that would be a good time to refactor it... and leave it as-is for now? Just wanted to point it out at least.

emreboga: Hey @artursapek , I totally agree and have thought about a separate adaptor to handle the page and indexes for each API fetch. If you think that'd not be a throw away code and can be moved into the SDK for the Transaction history implementation there, I already had this in mind. Otherwise I think it's better to keep all that implementation inside the hook where we can retire easily once SDK had the functionality. LMK 🙏

artursapek (dismissed):

kev1n-peters (approved):

PR #2595, opened 2024-09-13
Isolate chain badge background color so main background can be transparent

artursapek:
Trying to avoid this:
image

### DIFF
--- a/wormhole-connect/src/theme.ts
+++ b/wormhole-connect/src/theme.ts
@@ -33,6 +33,7 @@ export type WormholeConnectPartialTheme = {
   divider?: string;
   background?: {
     default: string;
+    badge: string;
   };
   text?: {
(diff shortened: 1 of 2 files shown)


yuli-ferna (approved):

Wesleyleung:
looks fine to me in Connect (below) fwiw, but it is a problem in Portal

Screenshot 2024-09-13 at 5 52 02 PM

artursapek:
> looks fine to me in Connect (below) fwiw, but it is a problem in Portal
> 
> Screenshot 2024-09-13 at 5 52 02 PM

That's why I made this change. Portal can now specify that they want a transparent background in general, but still keep a particular opaque color for these badges.

It's not a problem in the Netlify previews because those use an opaque background.

PR #2600, opened 2024-09-13
Increase polling frequency as we approach the ETA

artursapek:
This will make Connect poll faster (every 1 second) for ETAs < 1 minute

### DIFF
--- a/wormhole-connect/src/hooks/useTrackTransfer.ts
+++ b/wormhole-connect/src/hooks/useTrackTransfer.ts
@@ -19,10 +19,11 @@ const useTrackTransfer = (): void => {
   useEffect(() => {
     let isActive = true;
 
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/hooks/useTrackTransfer.ts
```diff
@@ -60,9 +69,25 @@ const useTrackTransfer = (): void => {
         } catch (e) {
           console.error('Error tracking transfer:', e);
```
kev1n-peters, line 71: My only concern is if the `sleepTime` is near 1 second and we are getting rate limited and keep on hitting the RPC. Should we backoff if an exception is thrown? That might complicate this, though. Just wanted to raise this point

kev1n-peters (approved):

PR #2601, opened 2024-09-13
Log commit hash in telemetry events

artursapek:
This adds a more specific build identifier (which git commit hash was used) to all Connect telemetry events.

This is generally useful, but particularly for immediately checking which commit hash a given build is from, using the `load` event that Connect fires when it first loads:

image

### DIFF
--- a/wormhole-connect/vite.config.ts
+++ b/wormhole-connect/vite.config.ts
@@ -1,4 +1,5 @@
 import path from 'path';
+import { execSync } from 'child_process';
 import { defineConfig, loadEnv } from 'vite';
 import react from '@vitejs/plugin-react-swc';
 import checker from '@artursapek/vite-plugin-checker';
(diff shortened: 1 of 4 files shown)


kev1n-peters (dismissed):

wormhole-connect/vite.config.ts
```diff
@@ -13,6 +14,18 @@ const packagePath = __dirname.endsWith('wormhole-connect')
   : './wormhole-connect/package.json';
 const { version } = require(packagePath);
 
+let gitHash = 'unknown';
+try {
+  gitHash = execSync("git log | head -n 1 | awk '{ print $2 }'").toString();
```
kev1n-peters: Is the second field that `awk` prints always the commit hash? Another way is `git log -1 --format=%H` but I defer

artursapek: that's cleaner because it doesn't depend on `head` and `awk`. updated.

kev1n-peters (approved):

PR #2608, opened 2024-09-16
Refactor AmountInput and Routes error handling.

nikarm22:
UI:
[Screencast from 2024-09-16 23-21-40.webm](https://github.com/user-attachments/assets/992dbdea-c408-46e1-8a53-71c9608f7482)


wormhole-connect/src/hooks/useSortedRoutesWithQuotes.ts
```diff
@@ -0,0 +1,118 @@
...
+        if (quote?.success) {
+          return {
+            route,
+            quote,
+          };
+        } else {
+          return undefined;
+        }
+      })
+      .filter(Boolean) as RouteWithQuote[];
```
artursapek, line 72: we should still show routes that received a `MinAmountError` - so the user can see that there are other routes available if they increase their size

nikarm22: Min amount error is unified with others and displayed bellow amount input.

wormhole-connect/src/hooks/useAmountValidation.ts
```diff
@@ -0,0 +1,109 @@
...
+        if (!minAmountError?.min) {
+          return minAmount;
+        }
+
+        if (!minAmount) {
+          return minAmountError.min;
+        }
+
+        const minAmountNum = parseFloat(minAmountError.min.amount);
+        const existingMin = parseFloat(minAmount.amount);
```
artursapek: `parseFloat` is valid here and works, but I want to avoid using floats as much as possible in our amount-related code... just to set a precedent. so do you mind changing this to `BigInt(minAmount.amount)` etc?

wormhole-connect/src/hooks/useAmountValidation.ts
```diff
@@ -0,0 +1,109 @@
...
+  const minAmount = useMemo(
+    () =>
+      Object.values(props.quotesMap).reduce((minAmount, quoteResult) => {
+        if (quoteResult?.success) {
+          return minAmount;
+        }
+
+        const minAmountError = quoteResult?.error as routes.MinAmountError;
+
+        if (!minAmountError?.min) {
```
artursapek: this could be more concise and easier to follow. I'd first establish that `quoteResult` is defined, then use `instanceof`:

```
if (!quoteResult || quoteResult.success) {
  return minAmount;
}

if (quoteResult.error instanceof routes.MinAmountError) {
  // ... do the thing
} else {
  return minAmount;
}
```

`instanceof` handles undefined properties just fine:

image

wormhole-connect/src/hooks/useSortedRoutesWithQuotes.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (routeA.quote.eta && routeB.quote.eta) {
+        if (routeA.quote.eta > routeB.quote.eta) {
+          return 1;
+        } else if (routeA.quote.eta < routeB.quote.eta) {
+          return -1;
+        }
+      }
+
+      // 3. Compare relay fees
+      if (routeA.quote.relayFee && routeB.quote.relayFee) {
```
artursapek: I think it's probably better to use the _amount out_ instead (`quote.destinationToken.amount.amount`) because some routes like Mayan could have more than one reason for a difference in the amount out, which might not be encapsulated in the `relayFee` (a swap or auction process).

wormhole-connect/src/hooks/useAmountValidation.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  const numAmount = Number.parseFloat(amount);
+  // Input errors
+  if (Number.isNaN(numAmount)) {
+    return {
+      error: 'Amount must be a number.',
+    };
+  }
+  if (numAmount <= 0) {
```
artursapek: maybe this is me, but I think it's not even worth showing an error in this case. like, duh. all it does is flash an error while someone is keying in `0.001` and haven't pushed `1` yet etc. nobody is going to be confused that they can't bridge zero.

wormhole-connect/src/hooks/useAmountValidation.ts
nikarm22: Indeed, that was my first implementation, but for some reason I consistently getting false negative results.
Here is the output of the error, MinQuoteError and instance of:
Screenshot from 2024-09-17 01-01-25

I suspect MinQuoteError is being transpiled/bundled to 2 different JS chunks, and since inctanceof does the check by reference, if fails. Tho, I'm not sure about it, didn't dug into it. I'll add a 'workaround' comment in code.

wormhole-connect/src/hooks/useAmountValidation.ts
nikarm22: Tunnel visioned this part, I forgot that `amount.Amount` stores pure integer.

wormhole-connect/src/hooks/useAmountValidation.ts
artursapek: oh got it. interesting.

one other thing we could do is add a little utility function like `isMinAmountError` which type-guards it. you could put it in `utils/sdkv2.ts`. could be similar to this: https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/connect/src/types.ts#L137-L141

you would do that instead of casting using `as`. wdyt?

artursapek (approved):

PR #2626, opened 2024-09-17
Minor UI tweaks

artursapek:
Addresses UI feedback:

- Reduce padding on history icon
- Remove colons for From, To, and Amount
- Fix typography so `"Inter"` font is used everywhere by default
- Balance should use contrast grey color


emreboga (approved):
LGTM!

PR #2656, opened 2024-09-19
[Redesign] Transaction in-progress widget

emreboga:
This PR adds the in-progress widget to the Bridge view for recently created transactions from the current browser and still are in progress.

Please see the loom here:
https://www.loom.com/share/4350b421e1bc40ea81709771813980c5?sid=05c9d701-172c-427e-9b4f-020f873678d7

Adding up-to-date screenshots for multiple transactions both in-progress and completed:

Screenshot 2024-09-23 at 10 43 02 AM

Screenshot 2024-09-23 at 10 44 03 AM


wormhole-connect/src/hooks/useTransactionHistoryMayan.ts
```diff
@@ -105,6 +106,7 @@ const useTransactionHistoryMayan = (
       tokenAddress: toTokenAddress,
       senderTimestamp: initiatedAt,
       explorerLink: `https://explorer.mayan.finance/swap/${sourceTxHash}`,
+      inProgress: clientStatus?.toLowerCase() !== 'completed',
```
emreboga: We need to check for `REFUNDED` as well. I'll add it here.

wormhole-connect/package.json
```diff
@@ -63,6 +63,7 @@
     "ethers": "^6.4.2",
     "lz-string": "^1.5.0",
     "material-ui-popup-state": "^5.0.4",
+    "poll": "^3.2.2",
```
artursapek: can we just copy this into `utils` and avoid adding more dependencies? it's like 10 loc: https://github.com/kleinfreund/poll/blob/main/src/poll.js

I am speaking from perspective of avoiding future "dependency hell" like we went through this summer.

wormhole-connect/src/utils/localStorage.ts
```diff
@@ -0,0 +1,56 @@
...
+  data: object, // Item data
+  prefix: string, // Prefix to use for the item key
+  max?: number, // Max number of items allowed with the same prefix
+) => {
+  const ls = window.localStorage;
+  const items: Array<string> = [];
+
+  // Get all current items with the prefix
+  for (let i = 0; i < ls.length; i++) {
+    const itemKey = ls.key(i);
```
artursapek: > The order of keys is user-agent defined, so you should not rely on it.

https://developer.mozilla.org/en-US/docs/Web/API/Storage/key

wormhole-connect/src/utils/localStorage.ts
```diff
@@ -0,0 +1,56 @@
...
+      items.push(itemKey);
+    }
+  }
+
+  // Remove the oldest items that puts the length at or over the limit
+  if (max && items.length >= max) {
+    for (let i = 0; i < items.length - max + 1; i++) {
+      ls.removeItem(items[i]);
+    }
+  }
```
artursapek: are you just trying to keep a chronological array in localStorage, and trim it to a max length? why not just use a single key and store an array? you could trim it by using `slice`. that way the order is actually reliable. I don't understand the reason for doing it this way

Wesleyleung:
Instead of showing "Any time now!" and removing the status bar, can we keep the status bar except leave it in indeterminate form https://mui.com/material-ui/react-progress/#linear-indeterminate?. As for the text, perhaps change it to "Wrapping up..." instead of "Any time now!"?

Then, when the txn actually confirms complete, continue to keep the status bar but show it in its fully filled in completed state.

The widget size right now shifts in height due to the status bar, and I feel its better to maintain its size throughout.

wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
```diff
@@ -0,0 +1,281 @@
...
+      setIsFetching(true);
+
+      try {
+        const res = await fetch(explorerInfo.apiUrl);
+        const resPayload = await res.json();
+
+        if (!cancelled) {
+          let txInProgress = true;
+          let txSenderTimestamp = '';
+          if (isMayanTx) {
```
artursapek: can we normalize these into a common data structure? you could just store a boolean called `inProgress`

I really hate adding code referencing mayan like this. I will keep nagging about this: Connect is supposed to be route-agnostic. there's no reason this widget should have to know what Mayan is.

emreboga:
@Wesleyleung if we wanna change "Any time now!" to "Wrapping up!", you wanna change it in Redeem view as well?

wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
emreboga: @artursapek I agree, but it needs to know about Mayan at some level, if we want to parse the completed status from its API response. One mitigation we can do is to move this logic in a parser utility, so it'll be at least not in the component.
Also, ideally, this can also be moved into the SDK to get the status of a transaction. Then we can remove all these logic from Connect.

Wesleyleung:
> @Wesleyleung if we wanna change "Any time now!" to "Wrapping up!", you wanna change it in Redeem view as well?

Yes, no exclamation though "Wrapping up...".  There are several things that I'd like to update in the redeem view too (problem for another day :)) https://docs.google.com/presentation/d/1ftqm2OVKuaQa99h-KnwA5aDf9QO8qOr6SQoj6l32pAA/edit#slide=id.g2d2f13a2ff9_0_92)

wormhole-connect/src/utils/localStorage.ts
emreboga: We can't even depend on a 0-based indexed order 😢 I'll add a timestamp to the localStorage and check that one instead.

wormhole-connect/src/utils/localStorage.ts
emreboga: I assumed I could depend on the insertion order. I was thinking maybe a timestamp, but using an array with a single entry sounds better.

wormhole-connect/src/utils/localStorage.ts
emreboga: Scratch that, I'll use a single entry with an array as you mentioned below.

wormhole-connect/package.json
emreboga: This is hilarious! I was writing our own polling hook, then wanted to just test the API before building the component and opted for something ready to use, which turns out being 10 lines 🤦🏼😄  By the way it's been a long time seeing someone using do-while loop in JS.

wormhole-connect/src/utils/localStorage.ts
```diff
@@ -1,56 +1,71 @@
 import { TransactionLocal } from 'config/types';
 
-// Adds an object to localStorage
-export const addItemToLocalStorage = (
-  id: string, // Key for the item
-  data: object, // Item data
-  prefix: string, // Prefix to use for the item key
-  max?: number, // Max number of items allowed with the same prefix
-) => {
+const LOCAL_STORAGE_KEY = 'wormhole-connect:transactions:inprogress';
```
artursapek: seems like either this file or this const should have a more specific name... this isnt the only feature Connect uses `localStorage` for. I'd personally just rename this file to `txHistoryCache.ts` or something like that.

https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/utils/wallet/index.ts#L132-L134

wormhole-connect/src/utils/localStorage.ts
```diff
@@ -1,56 +1,71 @@
...
-  if (max && items.length >= max) {
-    for (let i = 0; i < items.length - max + 1; i++) {
-      ls.removeItem(items[i]);
-    }
+// Adds a TransactionLocal object to localStorage
+export const addTxToLocalStorage = (
+  data: TransactionLocal, // Item data
+) => {
+  const ls = window.localStorage;
+  const items: Array<TransactionLocal> | undefined = getTxsFromLocalStorage();
```
artursapek: in general it's preferable to add the return type to the function signature (`getTxsFromLocalStorage`) rather than leaving the function return value untyped and having to annotate the type in the assignment here.

wormhole-connect/src/utils/localStorage.ts
emreboga: Initially it had more generic functions but now very tx history specific. I'll rename the file 👍

wormhole-connect/src/utils/localStorage.ts
emreboga: Oh the return type is already there in the function definition, I forgot to remove the types from the callers.

artursapek (approved):

Issue #2671, opened 2024-09-20
[Bug] WalletConnect popup shows up unprompted


tonyjin:
## Description and context
WalletConnect popup shows up unprompted when WalletConnect is selected as the wallet option for a chain, but before "Connect source wallet" button is clicked.

## Steps to reproduce
1. Load Connect V2 on Portal
2. Choose source chain/token, select Ethereum
3. Connect source wallet, select WalletConnect, but do not choose an option or connect via QR
4. Reload page
5. Choose source chain/token
6. Note that WalletConnect popup shows up beneath the source chain/token popup automatically

Note: This does not appear to be consistently reproducible even with the steps above. It seems like if the localstorage property`wormhole-connect:wallet:Ethereum` is set to `WalletConnect`, this will occur.

## Experienced behavior
WalletConnect popup appears unexpectedly

## Expected behavior

WalletConnect popup should not appear until "Connect source wallet" button is clicked.

Attached screen recording: 
https://github.com/user-attachments/assets/0caa2a1e-3e97-4fc1-aac7-b55395335276

sebastianscatularo:
should be fixed once https://github.com/wormhole-foundation/wormhole-connect/pull/2686 gets merged

emreboga:
I can confirm that I don't see the issue after the above PR is merged.
@tonyjin you wanna double-check?

tonyjin:
I wasn't able to reliably reproduce the issue originally, but I also could not reproduce it in the development build post-merging of 2686.

PR #2686, opened 2024-09-20
#2671 - avoid auto connect when last used wallet was WalletConnect

sebastianscatularo:
- if the previous connected wallet was WalletConnect, avoid auto connecting it

### DIFF
--- a/wormhole-connect/src/utils/wallet/index.ts
+++ b/wormhole-connect/src/utils/wallet/index.ts
@@ -132,7 +132,8 @@ export const connectLastUsedWallet = async (
   const lastUsedWallet = localStorage.getItem(
     `wormhole-connect:wallet:${chainConfig.context}`,
   );
-  if (lastUsedWallet) {
+  // if the last used wallet is not WalletConnect, try to connect to it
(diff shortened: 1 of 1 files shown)


kev1n-peters:
Prior to this change, I was able to auto-reconnect WalletConnect w/o the QR code popup appearing. Is it worth investigating why the QR code would appear for some users, but not others when auto-reconnecting WC?

yuli-ferna (approved):

sebastianscatularo:
@kev1n-peters, this was in V1:

- https://github.com/wormhole-foundation/wormhole-connect/blob/release/wormhole-connect/src/utils/wallet/index.ts#L132

Using a mobile wallet as an example, if you disconnect your mobile app or if the session has expired, once you open a Connect instance again, WalletConnect will try to reuse the session. However, since it’s no longer valid, the WalletConnect popup will appear, and without a clear context, it may seem confusing.

I’ll check if there’s an easy way to detect whether WalletConnect still has a valid active session. But even if that’s the case, I’m not sure if it would be secure to auto-connect, as you could be accessing someone else’s wallet session

kev1n-peters (approved):

PR #2687, opened 2024-09-20
[Redesign] Automatic route selection should overwrite invalid routes

emreboga:
I've been seeing this case where a previous source/dest token selection's route might be invalid for a new token pair. In that case we should auto-select a new valid route.

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/index.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/index.tsx
@@ -165,9 +165,14 @@ const Bridge = () => {
       setSelectedRoute('');
     } else {
       const autoselectedRoute = route || validRoutes[0].route.name;
+      const isSelectedRouteValid =
+        validRoutes.findIndex((r) => r.route.name === selectedRoute) > -1;
 
-      // avoids overwriting selected route
-      if (!autoselectedRoute || !!selectedRoute) return;
+      // If no route is autoselected or we already have a valid selected route,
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/views/v2/Bridge/index.tsx
```diff
@@ -165,9 +165,14 @@ const Bridge = () => {
       setSelectedRoute('');
     } else {
       const autoselectedRoute = route || validRoutes[0].route.name;
+      const isSelectedRouteValid =
+        validRoutes.findIndex((r) => r.route.name === selectedRoute) > -1;
```
kev1n-peters, line 169: I think we need to include `selectedRoute` as a dependency

emreboga: Actually, we don't want to run this side-effect whan selected route changes. Only when the `route` has an its initial value or when we have the list of supported routes, so we can initially select a route among them. I tried to explain that in the comment above the side-effect, but please let me know if it's not clear 🙏 
```
// Set selectedRoute if the route is auto-selected
// After the auto-selection, we set selectedRoute when user clicks on a route in the list
```

kev1n-peters: Makes sense to me. We should really enable the linter check for hook dependencies and add an exception for this dependency (+ document why), but that's out of scope :)

kev1n-peters (approved):

PR #2696, opened 2024-09-23
Add `transfer.success` event to v2 Redeem view

artursapek:
image

image

### DIFF
--- a/wormhole-connect/src/views/v2/Redeem/index.tsx
+++ b/wormhole-connect/src/views/v2/Redeem/index.tsx
@@ -102,7 +102,7 @@ const Redeem = () => {
 
   const [claimError, setClaimError] = useState('');
   const [isClaimInProgress, setIsClaimInProgress] = useState(false);
-  const [isTransferComplete, setIsTranferComplete] = useState(false);
+  const [isTransferComplete, setIsTransferComplete] = useState(false);
   const [etaExpired, setEtaExpired] = useState(false);
 
   const [isWalletSidebarOpen, setIsWalletSidebarOpen] = useState(false);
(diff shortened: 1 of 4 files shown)


wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -130,10 +132,34 @@ const Redeem = () => {
...
     receiveAmount,
     eta = 0,
   } = txData!;
 
+  const getUSDAmount = useUSDamountGetter();
+
+  useEffect(() => {
+    // When we see the transfer was complete for the first time,
+    // fire a transfer.success telemetry event.
+    if (isTxComplete && !isTransferComplete) {
```
emreboga: Nitpick: Having both isTxComplete and isTransferComplete may confuse people looking at this later. Maybe we can name the second one something like `successEventTriggered`?

emreboga (approved):
LGTM!

wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -130,10 +133,34 @@ const Redeem = () => {
...
     receiveAmount,
     eta = 0,
   } = txData!;
 
+  const getUSDAmount = useUSDamountGetter();
+
+  useEffect(() => {
+    // When we see the transfer was complete for the first time,
+    // fire a transfer.success telemetry event.
+    if (isTxComplete && !transferSuccessEventFired) {
```
kev1n-peters, line 146: Do we need to include `transferSuccessEventFired` in the dep array? I guess `isTxComplete` will only change once (from `false` to `true`) so might not matter

PR #2711, opened 2024-09-24
[bug] binance errors

danielisaacgeslin:
https://github.com/user-attachments/assets/f575ef4a-8444-4986-877a-0736f5e2d47b

### DIFF
--- a/wormhole-connect/src/hooks/useBalanceChecker.ts
+++ b/wormhole-connect/src/hooks/useBalanceChecker.ts
@@ -7,8 +7,7 @@ import { useIsMounted } from './useIsMounted';
 export const useBalanceChecker = (
   quote?: routes.Quote<
     routes.Options,
-    routes.ValidatedTransferParams<routes.Options>,
-    any
+    routes.ValidatedTransferParams<routes.Options>
   >,
 ): {
   feeSymbol?: string;
--- a/wormhole-connect/src/hooks/useBalanceChecker.ts
+++ b/wormhole-connect/src/hooks/useBalanceChecker.ts
@@ -56,7 +56,8 @@ export const useBalanceChecker = (
   return {
     feeSymbol,
     isCheckingBalance,
-    hasSufficientBalance: !!state && state?.balance > state?.cost,
+    hasSufficientBalance:
+      !quote?.relayFee || (!!state && state?.balance > state?.cost),
     walletBalance: state?.balance,
     networkCost: state?.cost,
   };
--- a/wormhole-connect/src/hooks/useBalanceChecker.ts
+++ b/wormhole-connect/src/hooks/useBalanceChecker.ts
@@ -12,7 +12,7 @@ export const useBalanceChecker = (
 ): {
   feeSymbol?: string;
   isCheckingBalance: boolean;
-  isBalanceEnough: boolean;
+  hasSufficientBalance: boolean;
   walletBalance?: number;
   networkCost?: number;
 } => {
@@ -35,8 +35,10 @@ export const useBalanceChecker = (
     setCheckingBalance(true);
     (async () => {
       try {
+        if (!quote?.relayFee?.amount) return;
+
         const wallet = getWalletConnection(TransferWallet.SENDING);
-        if (!wallet || !quote?.relayFee?.amount) return;
+        if (!wallet) return;
 
         const cost = amount.whole(quote.relayFee.amount);
         const balance = parseFloat(await wallet.getBalance());
@@ -54,7 +56,7 @@ export const useBalanceChecker = (
   return {
     feeSymbol,
     isCheckingBalance,
-    isBalanceEnough: !!state && state?.balance > state?.cost,
+    hasSufficientBalance: !!state && state?.balance > state?.cost,
     walletBalance: state?.balance,
     networkCost: state?.cost,
   };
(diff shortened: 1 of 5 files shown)


wormhole-connect/src/hooks/useIsMounted.ts
emreboga, line 1: @danielisaacgeslin Thanks for separating this hook out, I believe we can use it in others as well 👍

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/WalletBalanceWarning.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import AlertBanner from 'components/v2/AlertBanner';
+import { Stack, Typography, useTheme } from '@mui/material';
+import { toFixedDecimals } from 'utils/balance';
+
+export type WalletBalanceWarningProps = {
+  isCheckingBalance: boolean;
+  isBalanceEnough: boolean;
```
emreboga: Nitpick: Would you consider a different naming, `hasSufficientBalance`?

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/WalletBalanceWarning.tsx
```diff
@@ -0,0 +1,58 @@
...
+  walletBalance?: number;
+  networkCost?: number;
+  symbol?: string;
+};
+
+export default function WalletBalanceWarning({
+  isCheckingBalance,
+  isBalanceEnough,
+  walletBalance,
```
emreboga, line 10: Nitpick: Is this the token's balance in the wallet? In that case do you it'd be better to name it `tokenBalance`?

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/WalletBalanceWarning.tsx
```diff
@@ -0,0 +1,58 @@
...
+  networkCost,
+  symbol,
+}: WalletBalanceWarningProps) {
+  const theme = useTheme();
+  const content = isCheckingBalance
+    ? `Checking if wallet balance is enough...`
+    : `Your wallet balance is not enough to cover the network cost. Please add more funds to your wallet.`;
```
emreboga: "Insufficient token balance to cover the network cost. Please add more funds to your wallet." ?

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/WalletBalanceWarning.tsx
```diff
@@ -0,0 +1,58 @@
...
+export default function WalletBalanceWarning({
+  isCheckingBalance,
+  isBalanceEnough,
+  walletBalance,
+  networkCost,
+  symbol,
+}: WalletBalanceWarningProps) {
+  const theme = useTheme();
+  const content = isCheckingBalance
+    ? `Checking if wallet balance is enough...`
```
emreboga: "Checking token balance for network costs." ?

wormhole-connect/src/hooks/useBalanceChecker.ts
```diff
@@ -0,0 +1,61 @@
...
+import config from 'config';
+import { useIsMounted } from './useIsMounted';
+
+export const useBalanceChecker = (
+  quote?: routes.Quote<
+    routes.Options,
+    routes.ValidatedTransferParams<routes.Options>
+  >,
+): {
+  feeSymbol?: string;
```
emreboga, line 13: Aren't these return props very similar (if not the same) to `WalletBalanceWarningProps` below? Can you import it here as well? Or maybe move it to `wormhole-connect/src/config/types.ts`?

wormhole-connect/src/hooks/useBalanceChecker.ts
```diff
@@ -0,0 +1,61 @@
...
+      Object.values(config.tokens),
+    )?.symbol;
+  }, [quote]);
+
+  useEffect(() => {
+    setCheckingBalance(true);
+    (async () => {
+      try {
+        const wallet = getWalletConnection(TransferWallet.SENDING);
+        if (!wallet || !quote?.relayFee?.amount) return;
```
emreboga: You can check and return for `!quote?.relayFee?.amount` even before getting the wallet.

wormhole-connect/src/hooks/useBalanceChecker.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        const cost = amount.whole(quote.relayFee.amount);
+        const balance = parseFloat(await wallet.getBalance());
+
+        if (isMounted.current) setState({ balance, cost });
+      } catch (e) {
+        console.error(e);
+        if (isMounted.current) setState(null);
+      } finally {
+        if (isMounted.current) setCheckingBalance(false);
```
emreboga, line 51: Is there a way to test this case? Like getting the isMounted hook unmounted while isCheckingBalance is true? It's a very very edge case I know, but do you think would it be possible to unmount the hook without the component still being mounted? In that case it might get stuck in checking state.

danielisaacgeslin: it wouldn't be possible to unmount the hook while the compo is mounted in a real case. Hooks can't be conditional and they're bound together in the same render branch therefore hooks would only unmount along with the compo that called them.

That being said, a realistic way to test this (nvm fast fingers) would be to set an arbitrary long wait in try clause whilst unmounting through the UI (pressing the back button i think)

wormhole-connect/src/views/v2/Bridge/ReviewTransaction/WalletBalanceWarning.tsx
danielisaacgeslin: leaving this one as is as per slack comments

emreboga (approved):

PR #2712, opened 2024-09-24
Show canonical tokens even when wallet is not connected

### LINKED ISSUE #2690: Add Wormhole-wrapped USDC on Sui to builtin token list
## Description and context

I cannot select USDC on Sui in either the To or From field at all (unless I have a balance). 

[Wormhole-wrapped USDC](https://suiscan.xyz/mainnet/coin/0x5d4b302506645c37ff133b98c4b50a5ae14841659738d6d733d59d0d217a93bf::coin::COIN/txs) on Sui is not a builtin token (even though it is de facto canonical). Add it so that it shows up in the token selection list (even when the user has not yet connected their wallet).

Screenshot 2024-09-22 at 8 47 37 AM

## Definition of done

Wormhole-wrapped USDC on Sui shows in both input fields for Sui (alongside the token SUI) even when your source wallet isn't connected.

kev1n-peters:
Wormhole-wrapped Ethereum USDC was not showing with Sui as the source chain, because it's a Wormhole-wrapped token, and we only display those if the user has a balance. However, this token is canonical on Sui, so we want to display it even when the users's wallet isn't connected.

Added an `isCanonical` function that the `TokenList` uses.

Fixes #2690

### DIFF
--- a/wormhole-connect/src/utils/index.ts
+++ b/wormhole-connect/src/utils/index.ts
@@ -408,6 +408,12 @@ export const isWrappedToken = (token: TokenConfig, chain: Chain) => {
   return token.nativeChain !== chain;
 };
 
+// Canonical tokens may be Wormhole-wrapped tokens that are canonical on the chain
+// e.g., Wormhole-wrapped Ethereum USDC is canonical on Sui
+export const isCanonicalToken = (token: TokenConfig, chain: Chain) => {
+  return token.key === 'USDCeth' && chain === 'Sui';
+};
+
 export const millisToHumanString = (ts: number): string => {
   if (ts > 60000) {
     const minutes = Math.ceil(ts / 60000);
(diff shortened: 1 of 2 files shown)


Wesleyleung:
In the ticket, I also noted I'm not able to bridge USDC into Sui at all after wallets are connected (example used was selecting native USDC on Solana for the origin and then seeing no options on Sui for the destination). Does this change also address that?

kev1n-peters:
> In the ticket, I also noted I'm not able to bridge USDC into Sui at all after wallets are connected (example used was selecting native USDC on Solana for the origin and then seeing no options on Sui for the destination). Does this change also address that?

This does not, because Wormhole-wrapped Solana USDC is considered a Frankenstein token on Sui

wormhole-connect/src/utils/index.ts
```diff
@@ -408,6 +408,12 @@ export const isWrappedToken = (token: TokenConfig, chain: Chain) => {
   return token.nativeChain !== chain;
 };
 
+// Canonical tokens may be Wormhole-wrapped tokens that are canonical on the chain
+// e.g., Wormhole-wrapped Ethereum USDC is canonical on Sui
+export const isCanonicalToken = (token: TokenConfig, chain: Chain) => {
+  return token.key === 'USDCeth' && chain === 'Sui';
```
emreboga, line 414: @kev1n-peters Is canonical token concept only valid for Sui?

kev1n-peters: It's a general concept, but this is the only token I know of now that needs to be special-cased like this

kev1n-peters: We may need to expand this list in the future

emreboga (approved):
LGTM!

PR #2715, opened 2024-09-24
Allow all flavors of USDC to be transferred on Sui

### LINKED ISSUE #2706: [Bug] Solana to Sui route doesn't exist for USDC? Possibly other routes too
## Steps to reproduce

1. Select USDC on Solana as From
2. USDC on Sui as To (this is strangely only possible if you go back to select USDC on Ethereum first in From, select USDC on Sui, and switch back to USDC on Solana in From)
3. Get zero routes and no feedback why

## Experienced behavior

Screenshot 2024-09-23 at 11 16 25 PM
Screenshot 2024-09-23 at 11 36 34 PM

kev1n-peters:
Should be removed once CCTP is launched on Sui and Wormhole-wrapped USDC flavors are deprecated.

Fixes #2706

### DIFF
--- a/wormhole-connect/src/utils/index.ts
+++ b/wormhole-connect/src/utils/index.ts
@@ -427,7 +427,7 @@ export const isCanonicalToken = (token: TokenConfig, chain: Chain) => {
 };
 
 export const isStableCoin = (token: TokenConfig) => {
-  return token.symbol === 'USDC' || token.symbol === 'USDT';
+  return ['USDC', 'USDT', 'DAI'].includes(token.symbol);
 };
 
 export const millisToHumanString = (ts: number): string => {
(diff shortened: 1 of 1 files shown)


kev1n-peters:
image

Wesleyleung:
Assuming these other ones like USDCarb, etc are also able to be shown then I think it looks good to me.

Screenshot 2024-09-24 at 7 14 55 PM

wormhole-connect/src/utils/index.ts
```diff
@@ -414,6 +426,10 @@ export const isCanonicalToken = (token: TokenConfig, chain: Chain) => {
   return token.key === 'USDCeth' && chain === 'Sui';
 };
 
+export const isStableCoin = (token: TokenConfig) => {
+  return token.symbol === 'USDC' || token.symbol === 'USDT';
```
emreboga: We don't have DAI as stablecoin?

emreboga (approved):
LGTM!

PR #2722, opened 2024-09-25
[improvements] Cleanup some errors

danielisaacgeslin:
* Added missing forward ref in button compos. It was causing some console errors and MIGHT have affected tooltip positioning
* Changed a couple of kebab-case styles causing some console errors that MIGHT have caused incorrect styling


sebastianscatularo (approved):
looks good

PR #2731, opened 2024-09-25
TokenBridge route: Only support configured wrapped tokens

kev1n-peters:
Only allow wrapped tokens to be selected on the destination chain if they're actually configured. Otherwise users have no way to select them when transferring back.


emreboga (approved):
LGTM!

PR #2748, opened 2024-09-26
Bug/binance errors 2

danielisaacgeslin:
* code previously reverted
* fixes to cover additional cases

https://preview.portalbridge.com/5a97995403/


wormhole-connect/src/hooks/useBalanceChecker.ts
```diff
@@ -0,0 +1,73 @@
...
+          await wallet.getBalance(quote.relayFee.token.address?.toString()),
+        );
+
+        if (typeof balance !== 'number') {
+          throw new Error(
+            `Expected balance to be a number, but got "${balance}"`,
+          );
+        }
+
+        if (isMounted.current) setState({ balance, cost });
```
artursapek: I don't think this is idiomatic React. you should return [cleanup code](https://react.dev/reference/react/useEffect#usage) in your `useEffect` instead of using this `isMounted` thing. set a boolean `canceled = false` and return a function from your hook that sets it to true. then just check for `if (!canceled)` here instead.

artursapek (changes requested):

wormhole-connect/src/hooks/useBalanceChecker.ts
danielisaacgeslin: https://github.com/wormhole-foundation/wormhole-connect/pull/2748/files#diff-b9861d794f4711dee7eb03464dbf080cd493cffe2c3b5ca46d4e0d8b43a01afd

danielisaacgeslin: @artursapek have you had time to re-check this? ty

artursapek: I don't understand. what I'm saying is `useIsMounted` shouldn't even exist.

wormhole-connect/src/hooks/useBalanceChecker.ts
```diff
@@ -0,0 +1,73 @@
...
+      try {
+        if (!quote?.relayFee?.amount) return;
+
+        const wallet = getWalletConnection(TransferWallet.SENDING);
+        if (!wallet) return;
+
+        const cost = amount.whole(quote.relayFee.amount);
+
+        const balance = parseFloat(
+          await wallet.getBalance(quote.relayFee.token.address?.toString()),
```
kev1n-peters: What if the relayFee.token is a token that doesn't exist on this chain?

wormhole-connect/src/hooks/useBalanceChecker.ts
kev1n-peters: We should set a flag in the hook instead of using `useIsMounted`:
```
  useEffect(() => {
    let isActive = true;

    setCheckingBalance(true);
    (async () => {
      try {
        if (!quote?.relayFee?.amount) return;

        const wallet = getWalletConnection(TransferWallet.SENDING);
        if (!wallet) return;

        const cost = amount.whole(quote.relayFee.amount);

        const balance = parseFloat(
          await wallet.getBalance(quote.relayFee.token.address?.toString())
        );

        if (typeof balance !== 'number') {
          throw new Error(
            `Expected balance to be a number, but got "${balance}"`
          );
        }

        if (isActive) setState({ balance, cost });
      } catch (e) {
        console.error(e);
        if (isActive) setState(null);
      } finally {
        if (isActive) setCheckingBalance(false);
      }
    })();

    return () => {
      isActive = false;
    };
  }, [quote]);
  ```

File: wormhole-connect/src/hooks/useBalanceChecker.ts (hunk 2 above)
Comment (kev1n-peters) [reply wormhole-connect/src/hooks/useBalanceChecker.ts]: We should also handle the case where `relayFee.token.address` is set to `"native"` (native gas token)

File: wormhole-connect/src/hooks/useBalanceChecker.ts (hunk 1 above)
Comment (danielisaacgeslin) [reply wormhole-connect/src/hooks/useBalanceChecker.ts]: the is mounted hook has that same logic to be reutilised

File: wormhole-connect/src/hooks/useBalanceChecker.ts (hunk 2 above)
Comment (danielisaacgeslin) [reply wormhole-connect/src/hooks/useBalanceChecker.ts]: > What if the relayFee.token is a token that doesn't exist on this chain?

there's an early return 8 lines up (line 38) for this case

> We should also handle the case where `relayFee.token.address` is set to `"native"` (native gas token)

added an extra check for this

File: wormhole-connect/src/hooks/useBalanceChecker.ts (hunk 1 above)
Comment (artursapek) [reply wormhole-connect/src/hooks/useBalanceChecker.ts]: there's no need for a new hook at all. we're showing you the convention that you should follow.

Comment (danielisaacgeslin) [reply wormhole-connect/src/hooks/useBalanceChecker.ts]: pushed those changes

File: wormhole-connect/src/hooks/useBalanceChecker.ts [hunk 3]
```diff
@@ -0,0 +1,77 @@
...
+    (async () => {
+      try {
+        if (!quote?.relayFee?.amount) return;
+
+        const wallet = getWalletConnection(TransferWallet.SENDING);
+        if (!wallet) return;
+
+        const cost = amount.whole(quote.relayFee.amount);
+
+        const feeTokenAddress = quote.relayFee.token.address?.toString();
```
kev1n-peters, line 50: If token.chain != chain we're checking the balance on then we should bail out

wormhole-connect/src/hooks/useBalanceChecker.ts
```diff
@@ -0,0 +1,77 @@
...
+            `Expected balance to be a number, but got "${balance}"`,
+          );
+        }
+
+        if (isActive) setState({ balance, cost });
+      } catch (e) {
+        console.error(e);
+        if (isActive) setState(null);
+      } finally {
+        if (isActive) setCheckingBalance(false);
```
kev1n-peters, line 61: Should we also reset the state here?

wormhole-connect/src/hooks/useBalanceChecker.ts
danielisaacgeslin: added that change

artursapek:
You know, now that I read this again, I realize you don't need to add a new hook for this at all. Why not just use `useGetTokenBalances` to fetch the native currency balance, and compare that to the quote inside of `ReviewTransaction.tsx`?

https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/views/v2/Bridge/index.tsx#L202-L206

I don't like that we're adding a new hook that calls `wallet.getBalance`... we should be able to rely on our existing balance fetching hook for all balance fetching logic so we don't have multiple implementations of the same thing.

I could imagine this looking more like this

```ts
  // ReviewTransaction/index.tsx
  //
  // this file already has quote, fromChain, and sendingWallet in scope

  const [hasMinimumBalance, setHasMinimumBalance] = useState(true);

  const { balances, isFetching: isFetchingBalances } = useGetTokenBalances(
    sendingWallet?.address || '',
    fromChain,
    ['native'],
  );

  useEffect(() => {
    // do your balance check in here
    if (balances['native'] < quote.fee) { // etc etc this is pseudo code
      setHasMinimumBalance(false);
    }
  })
```

Does that make sense?

artursapek (changes requested):
You know, now that I read this again, I realize you don't need to add a new hook for this at all. Why not just use `useGetTokenBalances` to fetch the native currency balance, and compare that to the quote inside of `ReviewTransaction/index.tsx`?

https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/views/v2/Bridge/index.tsx#L202-L206

I don't like that we're adding a new hook that calls `wallet.getBalance`... we should be able to rely on our existing balance fetching hook for all balance fetching logic so we don't have multiple implementations of the same thing.

I could imagine this looking more like this

```ts
  // ReviewTransaction/index.tsx
  //
  // this file already has quote, fromChain, and sendingWallet in scope

  const [hasMinimumBalance, setHasMinimumBalance] = useState(true);

  const { balances, isFetching: isFetchingBalances } = useGetTokenBalances(
    sendingWallet?.address || '',
    fromChain,
    ['native'],
  );

  useEffect(() => {
    // do your balance check in here
    if (balances['native'] < quote.fee) { // etc etc this is pseudo code
      setHasMinimumBalance(false);
    }
  })
```

Does that make sense?

PR #2761, opened 2024-09-27
Fix missing OKX wallet from default

danielisaacgeslin:
Fix missing OKX wallet from default


artursapek (dismissed):

wormhole-connect/src/utils/wallet/evm.ts
```diff
@@ -23,6 +24,7 @@ const eip6963Wallets = Object.entries(Eip6963Wallets).reduce(
 
 export const wallets = {
   ...eip6963Wallets,
+  okxwallet: new InjectedWallet({ name: InjectedWallets.OKXWallet }),
```
sebastianscatularo: since InjectedConnector works over `window.ethereum` global variable, when a different wallet overrides it, and you click on OKX it will open the other wallet, a better alternative would be implementing it through https://github.com/ethereum/EIPs/blob/master/EIPS/eip-6963.md#motivation

https://github.com/user-attachments/assets/4da17db0-4c0c-4651-b6c9-b73a86c40c30

sebastianscatularo (changes requested):

sebastianscatularo (approved):
Looks good :)

Issue #2740, opened 2024-09-26
[QA] Amount displayed in "More routes available" message should have more decimal places


tsadovska:
## Describe the current behavior in detail
Checked on https://preview.portalbridge.com/712d941311/.

Steps:
1. Prepare to send 0.000001 WBTC from Ethereum (MetaMask) to BSC (MetaMask)
2. Check amount in "More routes available" message 

Actual result: Amount is displayed as 0 WBTC.
Expected result: Amount has more decimals placed. 

## Notes/ Attachments

![Screenshot 2024-09-26 at 09 35 44](https://github.com/user-attachments/assets/d4706eba-1bd0-405d-a417-6f9eb7d58f54)

PR #2762, opened 2024-09-27
[QA] Amount displayed in "More routes available" message should have more decimal places

danielisaacgeslin:
[[QA] Amount displayed in "More routes available" message should have more decimal places](https://github.com/wormhole-foundation/wormhole-connect/issues/2740)

image

### DIFF
--- a/wormhole-connect/src/hooks/useAmountValidation.ts
+++ b/wormhole-connect/src/hooks/useAmountValidation.ts
@@ -83,7 +83,7 @@ export const useAmountValidation = (props: Props): HookReturn => {
   if (allRoutesFailed) {
     if (minAmount) {
       const formattedAmount = sdkAmount.whole(minAmount).toLocaleString('en', {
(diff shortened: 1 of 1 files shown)


artursapek (changes requested):
Why are we assuming 6 is always the correct precision? We should just use `sdkAmount.display` here instead. It will format it to a string with the correct precision without trailing zeroes. https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/core/base/src/utils/amount.ts#L161-L163

danielisaacgeslin:
> Why are we assuming 6 is always the correct precision? We should just use `sdkAmount.display` here instead. It will format it to a string with the correct precision without trailing zeroes. https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/core/base/src/utils/amount.ts#L161-L163

@artursapek changes made

artursapek (approved):

PR #2767, opened 2024-09-30
tokens with large balances not selectable fix

kev1n-peters:
Calling `Number(s)` where `s` is an amount string with commas returns NaN and was preventing tokens from large balances from being selected. Also we should be using relaceAll instead of relace when replacing commas.


emreboga (approved):
LGTM!

artursapek:
two comma club 😎

PR #2769, opened 2024-09-30
New approach to CDN-hosted Connect

artursapek:
- Stops using HTML attributes to pass config & theme into Connect (now using global variables)
- Adds `wormholeConnectHosted` utility function, exported from the package, which attaches a hosted version of Connect to any parent component in the parent application
- Switches from unpkg.com to jsdelivr.net

Hosted version is now used like this:

```ts
import {
  wormholeConnectHosted,
  MayanRoute
} from '@wormhole-foundation/wormhole-connect';

wormholeConnectHosted(document.getElementById('connect')!, {
  config: {
    routes: [MayanRoute],
    eventHandler: (e) => {
      console.log('Connect event', e);
    }
  }
});
```

### DIFF
--- a/wormhole-connect/src/main.tsx
+++ b/wormhole-connect/src/main.tsx
@@ -2,9 +2,6 @@ import React from 'react';
 import ReactDOM from 'react-dom/client';
 import WormholeConnect from './WormholeConnect';
 import ErrorBoundary from './components/ErrorBoundary';
-import { WormholeConnectConfig } from 'config/types';
-import { WormholeConnectPartialTheme } from './theme';
-import { WormholeConnectEvent } from 'telemetry/types';
 export * from './theme';
 
 // This is the entry point that runs when integrators add the Connect widget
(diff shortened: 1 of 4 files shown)


wormhole-connect/src/main.tsx
```diff
@@ -26,46 +23,10 @@ if (!container) {
...
-  if (themeAttr) {
-    const parsedTheme = JSON.parse(themeAttr);
-    theme = parsedTheme;
-  }
-} catch (e) {
-  console.error(`Error parsing custom config: ${e}`);
-  // Ignore
-}
+/* @ts-ignore */
+const config = window.__CONNECT_CONFIG as WormholeConnectConfig;
```
emreboga, line 27: @artursapek Aren't these available in window when `wormholeConnectHosted` is initiated? I don't see it being called here, am I missing something?

artursapek: The idea of `wormholeConnectHosted` is it adds a `script` tag to a page, which loads a separate all-encompassing bundle which includes React and runs Connect on the page. This file `main.tsx` is the entry point for that bundle.

So an integrator using a different framework like Vue.js for example can import `wormholeConnectHosted` in their bundle and run that function, which sets these globals and appends that `script` tag. This code will run in their main Vue.js app bundle. Then, this second all-encompassing Connect bundle loads from that script tag and runs with React, reading the config and theme from those globals.

The config used to be passed into this second bundle via HTML attributes `data-config="{...}"` but this approach is limited to JSON-serializable data, which our config is not anymore.

kev1n-peters (approved):

kev1n-peters (approved):

kev1n-peters (dismissed):

emreboga (approved):
LGTM!

PR #2778, opened 2024-10-01
bug fix: Assert signer connected to right EVM chain before sending

kev1n-peters:


artursapek (dismissed):
chad

kev1n-peters:
Still investigating how it was possible to land a TX on a different chain since it looks like ethers already does this check: https://github.com/ethers-io/ethers.js/blob/858a81b81173d9f5771517b375d05a157d3f59e8/src.ts/providers/abstract-signer.ts#L112

EDIT: Ignore the above function, it's not actually called. What ends up happening is that the wallet (e.g. Metamask) overrides the `chainId` at signing time to whichever network it's connected to.

artursapek (approved):

PR #2785, opened 2024-10-02
Display DestinationCapacityWarning

kev1n-peters:


kev1n-peters:
image

kev1n-peters:
image

wormhole-connect/src/utils/index.ts
```diff
@@ -462,3 +462,15 @@ export const millisToRelativeTime = (ts: number): string => {
     return '~1 minute ago';
   }
 };
+
+export const formatDuration = (seconds: number) => {
```
emreboga, line 466: Thought bubble: The number of time formatting utilities is growing :) I think we just need dayjs:
https://day.js.org/docs/en/display/format

emreboga (approved):
LGTM

artursapek (approved):

wormhole-connect/src/utils/index.ts
kev1n-peters: Good idea, I'll open an issue for it since the scope is wider than this targeted PR

PR #2787, opened 2024-10-02
[Redesign] UI polish for Redeem view

emreboga:
This PR addresses the UI polish requests which includes a redesign of the Circular Progress bar for both determinate and indeterminate cases (see below). It also cleans up a few more React warnings/errors.

You can see the update for smoother progress bar transition here:
https://www.loom.com/share/67634ad8fceb4d99a82804b39cefb3eb?sid=ba85d219-3077-4940-befc-6a48227d0af9

Screenshots for all UI states:

Screenshot 2024-10-02 at 3 42 11 PM

Screenshot 2024-10-02 at 3 49 46 PM

Screenshot 2024-10-02 at 5 10 23 PM

Screenshot 2024-10-02 at 5 15 23 PM

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
@@ -207,7 +207,6 @@ const AmountInput = (props: Props) => {
                 e.currentTarget.blur();
               },
               step: '0.1',
-              pattern: '[0-9]+([.|,][0-9]{1,2})?',
             }}
             placeholder="0"
             variant="standard"
--- a/wormhole-connect/src/views/v2/Redeem/index.tsx
+++ b/wormhole-connect/src/views/v2/Redeem/index.tsx
@@ -652,9 +652,7 @@ const Redeem = () => {
           {isClaimInProgress || !isTxAttested ? (
             <Stack direction="row" alignItems="center">
               <CircularProgress size={24} />
-              <Typography marginLeft="4px" textTransform="none">
-                Transfer in progress
-              </Typography>
+              <Typography textTransform="none">Transfer in progress</Typography>
             </Stack>
           ) : (
             <Typography textTransform="none">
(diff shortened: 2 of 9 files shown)


wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
```diff
@@ -207,7 +207,6 @@ const AmountInput = (props: Props) => {
                 e.currentTarget.blur();
               },
               step: '0.1',
-              pattern: '[0-9]+([.|,][0-9]{1,2})?',
```
emreboga, line 210: We don't need this anymore as we are checking the input type/format in the inner Input control.

wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -276,19 +292,101 @@ const Redeem = () => {
...
+    }
+
+    return (timePassed / eta) * 100;
+  }, [eta, txTimestamp, seconds]);
+
+  // In-progress circular progress bar
+  const etaCircularProgress = useMemo(() => {
+    return (
+      <>
+        <svg width={0} height={0}>
```
emreboga, line 330: This is the linear gradient we define in DOM to use in the circular progress bar below for gradient effect. That's why the svg doesn't need to be shown, it's only the def we need.

wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -276,19 +292,101 @@ const Redeem = () => {
...
+    // Check if eta has already
+    if (timePassed >= eta) {
+      return 100;
+    }
+
+    return (timePassed / eta) * 100;
+  }, [eta, txTimestamp, seconds]);
+
+  // In-progress circular progress bar
+  const etaCircularProgress = useMemo(() => {
```
emreboga, line 327: I wasn't sure but can move this progress bar in its own file to make this one smaller. It's not used anywhere else though.

wormhole-connect/src/views/v2/Redeem/index.tsx
artursapek: this is pretty 1337 I gotta say well done

artursapek (approved):

PR #2792, opened 2024-10-02
[Redesign] Custom config feature flag for in-progress widget

emreboga:
Adds the prop showInProgressWidget under UI custom props to hide in-progress widgets when not set to true.

```
{
  network: 'mainnet',
  ui: { showInProgressWidget: true },
  ...
}
```


artursapek (approved):
thank you for putting it under `ui` <3

PR #2793, opened 2024-10-02
Always use a transparent background

artursapek:
This is almost always desired behavior. If an integrator wants to put Connect into an opaque box of a unique color, they can easily do that by styling its parent.

image


yuli-ferna (approved):
👏🏼

emreboga (approved):
LGTM!

PR #2812, opened 2024-10-07
Simpler custom theme type

artursapek:
This introduces a new type `WormholeConnectTheme` which is a more integrator-friendly way of creating custom themes. It allows integrators to provide a simple mapping of whichever colors they want to modify in our default themes, and uses MUI's color derivation to do the rest.

```ts
export type WormholeConnectTheme = {
  // "dark" or "light"
  mode: PaletteMode;
  // Color of input fields, like asset picker and dropdowns
  input?: string;
  // Primary brand color
  primary?: string;
  // Secondary brand color
  secondary?: string;
  // Primary text color
  text?: string;
  // Secondary text color
  textSecondary?: string;
  // Error message color
  error?: string;
  // Success message color
  success?: string;
  // Background color for badges in asset picker
  badge?: string;
  // Font family
  font?: string;
};


```

This mirrors our approach to configs: there is an external-facing `WormholeConnectConfig` type for integrators to provide, and an `InternalConfig` type that we create from that and use for all of the code internally.

|        | External type for integrators | Internal Type  |
|--------|-------------------------------|----------------|
| Config | `WormholeConnectConfig`         | `InternalConfig` |
| Theme  | `WormholeConnectTheme`          | `InternalTheme`  |

### DIFF
--- a/wormhole-connect/src/theme.ts
+++ b/wormhole-connect/src/theme.ts
@@ -4,7 +4,7 @@ import green from '@mui/material/colors/green';
 import orange from '@mui/material/colors/orange';
 import purple from '@mui/material/colors/purple';
 import red from '@mui/material/colors/red';
-import { PaletteMode } from '@mui/material';
+import { PaletteMode, PaletteColorOptions, Theme } from '@mui/material';
 import { OPACITY } from './utils/style';
 
 export type PaletteColor = {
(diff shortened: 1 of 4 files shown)


wormhole-connect/src/theme.ts
```diff
@@ -300,15 +323,47 @@ export const dark: WormholeConnectTheme = {
...
+  // Override built-in theme with whichever custom values we've been provided
+  if (customTheme) {
+    if (customTheme.input) {
+      theme.modal = {
+        background: customTheme.input,
+      };
+    }
+    if (customTheme.primary) {
+      theme.primary = {
+        main: customTheme.primary,
```
artursapek, line 339: MUI derives all the other values for `PaletteColor` when just provided with `main`.

https://mui.com/material-ui/customization/theming/#api

emreboga (dismissed):
LGTM!

emreboga (approved):
LGTM!

PR #2822, opened 2024-10-09
Pet peeve: handle "." as a numeric input

artursapek:
This change lets me type ".01" instead of having to type "0.01" in the amount field.


emreboga (approved):
LGTM!

PR #2828, opened 2024-10-10
pre select destination and preferred route

sebastianscatularo:
Added `toTokenKey` and `preferredRouteName` params to `ui` object config to allow integrators to pre-select target token and mark a configured route as preferred

If the toTokenKey does not exist or is invalid, it will fail silent
If the preferred route does not exist or is invalid, it will fail silent

Sample config piece

```json
"defaultInputs": {
  "fromChain": "Ethereum",
  "toChain": "Arbitrum",
  "tokenKey": "USDCeth",
  "toTokenKey": "USDCarbitrum",
  "preferredRouteName": "ManualCCTP"
}
```

https://github.com/user-attachments/assets/9fb444f8-b4da-4868-9435-ec8f34b38634

### DIFF
--- a/wormhole-connect/src/hooks/useComputeDestinationTokens.ts
+++ b/wormhole-connect/src/hooks/useComputeDestinationTokens.ts
@@ -11,6 +11,7 @@ import { Chain } from '@wormhole-foundation/sdk';
 type Props = {
   sourceChain: Chain | undefined;
   sourceToken: string;
+  destToken: string;
   destChain: Chain | undefined;
   route?: string;
 };
@@ -20,7 +21,7 @@ type ReturnProps = {
 };
 
 const useComputeDestinationTokens = (props: Props): ReturnProps => {
-  const { sourceChain, destChain, sourceToken } = props;
+  const { sourceChain, destChain, sourceToken, destToken } = props;
 
   const dispatch = useDispatch();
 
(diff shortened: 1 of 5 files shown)


wormhole-connect/src/views/v2/Bridge/index.tsx
```diff
@@ -165,7 +167,11 @@ const Bridge = () => {
     if (validRoutes.length === 0) {
       setSelectedRoute('');
     } else {
-      const autoselectedRoute = route || validRoutes[0].route.name;
+      const preferredRoute = validRoutes.find(
+        (route) => route.route.name === preferredRouteName,
+      );
+      const autoselectedRoute =
+        route ?? validRoutes[0].route.name ?? preferredRoute?.route.name;
```
artursapek: Shouldn't the preferred route come before the `validRoutes[0].route.name`?

emreboga (dismissed):
LGTM!

wormhole-connect/src/hooks/useComputeDestinationTokens.ts
```diff
@@ -54,8 +55,15 @@ const useComputeDestinationTokens = (props: Props): ReturnProps => {
...
       setIsFetching(false);
 
-      if (destChain && supported.length === 1) {
-        if (!canceled) {
+      if (active && destChain && destToken !== '') {
+        // check if the pre selected destToken is supported
+        const isTokenSupported = supported.some((t) => t.key == destToken);
+        if (!isTokenSupported) {
+          // if not, clear the destToken
+          dispatch(setDestToken(''));
```
artursapek: Why is this necessary? Can't we just let the UI show no quotes if there are no routes?

sebastianscatularo: you're right, removed

wormhole-connect/src/views/v2/Bridge/index.tsx
sebastianscatularo: yes, I've tried to optimize it, and that is why I've moved to the last position, but, you are right, updated

artursapek: wdym by "optimize it"?

artursapek (approved):
the diff in `wormhole-connect/src/hooks/useComputeDestinationTokens.ts` is pointless now but lgtm

tsadovska:
Checked on Netlify https://wormhole-connect-mainnet.netlify.app/, `toTokenKey` and `preferredRouteName` configs work correctly. If token or route does not exist, then `toTokenKey` or `preferredRouteName` are ignored silently as expected.

Issue #2830, opened 2024-10-11
Add Governor enqueue warning to route quote


Wesleyleung:
Use a similar design to the NTT rate limit warnings 
![372875597-32c8df76-aceb-4ac4-972d-4515acc9f985](https://github.com/user-attachments/assets/66a0eeaa-5ab1-4ef8-bfb3-3c856a57179f)

Your transfer to <dest chain> may be delayed due to security rate limits set by the [Governor](https://wormholescan.io/#/governor). If your transfer is delayed, you will need to return after 24 hours to complete your transfer. Please consider this before proceeding.

artursapek:
we actually do have code that displays this exact warning string:

https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx#L279

the problem is, it looks for warning type `"DestinationCapacityWarning"` warning... but the SDK never emits this anywhere. it's instead [emitting `"GovernorLimitWarning"`](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/connect/src/protocols/tokenBridge/tokenTransfer.ts#L717-L731)

PR #2831, opened 2024-10-11
Handle governor warnings similarly to NTT queue

artursapek:
Addresses #2830

Adapts the existing transfer delay warning logic to also cover `GovernorLimitWarning` with a hard-coded 24 hour period.

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
@@ -10,7 +10,7 @@ import Typography from '@mui/material/Typography';
 import Stack from '@mui/material/Stack';
 import WarningIcon from '@mui/icons-material/Report';
 import { makeStyles } from 'tss-react/mui';
-import { amount, routes } from '@wormhole-foundation/sdk';
+import { amount, routes, Chain } from '@wormhole-foundation/sdk';
 
 import config from 'config';
 import TokenIcon from 'icons/TokenIcons';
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
```diff
@@ -262,26 +262,36 @@ const SingleRoute = (props: Props) => {
...
+      messages.push(
+        <TransferDelayWarning
+          chain={destChain}
+          token={config.tokens[destToken].symbol}
+          duration={maxDelay}
+        />,
+      );
+    }
+
     for (const warning of quote?.warnings || []) {
```
kev1n-peters, line 290: remove

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
```diff
@@ -459,4 +469,31 @@ const SingleRoute = (props: Props) => {
...
+      <Divider flexItem sx={{ marginTop: '8px' }} />
+      <Stack direction="row" alignItems="center">
+        <WarningIcon htmlColor={theme.palette.warning.main} />
+        <Stack sx={{ padding: '16px' }}>
+          <Typography color={theme.palette.warning.main} fontSize={14}>
+            {`Your transfer ${
+              props.chain ? `to ${props.chain} ` : ''
+            }may be delayed due to rate limits set by ${
+              props.token
+            }. If your transfer is delayed, you will need to return after ${duration} to complete the transfer. Please consider this before proceeding.`}
```
kev1n-peters, line 491: If this is an AutomaticTokenBridge transfer and it gets delayed by the governor, the transfer will be relayed once the VAA is emitted 24 hours later. The user doesn't need to return to complete it manually.

Wesleyleung: I think this copy is ok. Might be too complicated/confusing to have different copy for manual and automatic.

Wesleyleung:
@emreboga for governor warnings or NTT warnings, would these show up as history as "in-progress"? Or in the widget as well?

wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx
aadam-10: I think the text should say "after at most 24 hours", because enough room may free up in the rolling window before 24h (unless the tx is over the big tx limit but I guess we don't need to complicate this iteration)

as for later, I believe Wormholescan is able to supply the ETA when a tx is scheduled to be released so future iterations can provide better UX by telling the user how much they'll need to wait.

Wesleyleung: imo having a clear and consistent expectation (24 hours) even if its conservative is better than something that is not. As a user, I would be fine just checking back exactly the next day and seeing it done one time, vs thinking it *might* be done less than 24 hours and repeatedly checking in at random times throughout the day to be disappointed. If it happens to be done sooner and I check in sooner, great, then expectations underpromised overdelivered.

PR #2835, opened 2024-10-14
[Redesign] Replace explorer API with SDK route.track for in-progress widget

emreboga:
This PR replaces API calls (WHScan and Mayan) with the SDK track() functionality. The goal here is to be able to track CCTP_MANUAL transactions (currently not supported by WHScan API) and also move all the calls through SDK for the long term goal.

Please see below loom for details:
https://www.loom.com/share/a42a43cbc70d43f7aafe45bcae132c2f?sid=053d3550-09f8-4241-bd21-f94b50c1fc1c


wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
```diff
@@ -203,18 +146,18 @@ const WidgetItem = (props: Props) => {
...
-    if (!isRunning && inProgress && etaRemaining) {
+    if (!isRunning && !isCompleted && etaRemaining) {
       // Start only when:
       //   1- the timer hasn't been started yet and
       //   2- transaction is in progress and
       //   3- we have the remaining eta
       restart(new Date(Date.now() + etaRemaining), true);
     }
-  }, [etaRemaining, inProgress]);
+  }, [etaRemaining, isCompleted]);
```
kev1n-peters: Do we need to include `isRunning` in the deps here

wormhole-connect/src/utils/inProgressTxCache.ts
```diff
@@ -3,6 +3,12 @@ import { TransactionLocal } from 'config/types';
 const LOCAL_STORAGE_KEY = 'wormhole-connect:transactions:inprogress';
 const LOCAL_STORAGE_MAX = 3;
 
+// Bigint types cannot be serialized to a string
+// That's why we need to provide a replacer function to handle it separately
+// Please see for details: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/BigInt_not_serializable
+const JSONReplacer = (_, value: any) =>
+  typeof value === 'bigint' ? value.toString() : value;
```
artursapek, line 10: does it matter that we don't parse these back into `bigint` later when we read the value?

wormhole-connect/src/hooks/useTrackTransferInProgress.ts
```diff
@@ -0,0 +1,111 @@
+import { useEffect, useState } from 'react';
```
artursapek, line 1: https://github.com/wormhole-foundation/wormhole-connect/issues/2850

artursapek (dismissed):

wormhole-connect/src/utils/inProgressTxCache.ts
emreboga: Good point! At the time of `JSON.parse` we do not know which property we have previously converted to string (while calling `JSON.stringify`). We can track this in different ways, but track() function doesn't seem to be complaining about that when pass the receipt object, so it may end up being an unnecessary complication. WDYT?

wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
emreboga: Good catch!

wormhole-connect/src/utils/inProgressTxCache.ts
artursapek: I think you could avoid storing the receipt in localStorage by simply using [`resumeIfManual`](https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/routes/sdkv2/route.ts#L605) to fetch it for the first time when loading state from localStorage. that way you can store less data and not have to worry about `bigints` at all.

did you already try that and it didn't work?

artursapek: I really want to get Connect to a place where it never has to look at, or store, attestations at all. the SDK should be able to provide the right abstractions for us. adding this change where we're storing the receipt in localStorage is a step backwards.

emreboga: @artursapek `route.resume()` is currently supported for manual txs (bridge and CCTP). That's why I wasn't able to use it for resuming the in-progress tx when user clicks on the widget (that's another work).

artursapek (approved):

PR #2837, opened 2024-10-14
Hide terrible quotes; warn about mediocre ones

artursapek:
This PR makes two changes:

- Never show a quote that returns <=90% of the input amount, for transfers over $1000 USD
- Show a warning for any other quotes that return a USD value that's >=$20 _less_ than the input amount:

image

(Screenshot was taken with a threshold of $2 instead, for testing purposes)

I also un-broke the loading UI for `Routes`... which was broken recently by someone in a different change.

TODO:

- add % to warning

### DIFF
--- a/wormhole-connect/src/hooks/useAmountValidation.ts
+++ b/wormhole-connect/src/hooks/useAmountValidation.ts
@@ -58,7 +58,7 @@ export const useAmountValidation = (props: Props): HookReturn => {
   const numAmount = Number.parseFloat(amount);
 
   // Don't show errors when no amount is set or it's loading
-  if (!amount || !numAmount || props.isLoading || props.disabled) {
+  if (!amount || !numAmount || props.disabled) {
     return {};
   }
 
--- a/wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
+++ b/wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
@@ -43,16 +43,17 @@ const useRoutesQuotesBulk = (routes: string[], params: Params): HookReturn => {
   const [quotes, setQuotes] = useState<QuoteResult[]>([]);
 
   // TODO temporary
-  // Calculate USD amount for temporary $1000 Mayan limit
-  const tokenConfig = config.tokens[params.sourceToken];
+  // Calculate USD amount for temporary $10,000 Mayan limit
+  const sourceTokenConfig = config.tokens[params.sourceToken];
+  const destTokenConfig = config.tokens[params.destToken];
   const { usdPrices } = useSelector((state: RootState) => state.tokenPrices);
   const { isTransactionInProgress } = useSelector(
     (state: RootState) => state.transferInput,
   );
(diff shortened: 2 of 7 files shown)


wormhole-connect/src/hooks/useAmountValidation.ts
```diff
@@ -58,7 +58,7 @@ export const useAmountValidation = (props: Props): HookReturn => {
   const numAmount = Number.parseFloat(amount);
 
   // Don't show errors when no amount is set or it's loading
-  if (!amount || !numAmount || props.isLoading || props.disabled) {
+  if (!amount || !numAmount || props.disabled) {
```
artursapek, line 61: Armen had added this change which broke the Routes component loading state... undoing it here

kev1n-peters (approved):

kev1n-peters (approved):
lgtm, don't see any obvious issues

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -120,6 +121,26 @@ const useRoutesQuotesBulk = (routes: string[], params: Params): HookReturn => {
...
+    if (quote !== undefined && quote.success) {
+      const usdValueOut = calculateUSDPriceRaw(
+        amount.whole(quote.destinationToken.amount),
+        usdPrices.data,
+        destTokenConfig,
+      );
+
+      if (usdValue && usdValueOut) {
+        const valueRatio = usdValueOut / usdValue;
+        if (usdValue >= 1000 && valueRatio <= 0.9) {
```
emreboga, line 137: if usdValue is < 1000, you can skip the whole loop, right?

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -120,6 +121,26 @@ const useRoutesQuotesBulk = (routes: string[], params: Params): HookReturn => {
...
+  for (const name in quotesMap) {
+    const quote = quotesMap[name];
+    if (quote !== undefined && quote.success) {
+      const usdValueOut = calculateUSDPriceRaw(
+        amount.whole(quote.destinationToken.amount),
+        usdPrices.data,
+        destTokenConfig,
+      );
+
+      if (usdValue && usdValueOut) {
```
emreboga, line 135: is there a possibility that the value out can be 0?

tsadovska:
> Never show a quote that returns <=90% of the input amount, for transfers over $1000 USD

Scenario is not checked, because of insufficient amount. 

> Show a warning for any other quotes that return a USD value that's >=$20 less than the input amount.

Warning "Output amount is much lower than input amount" is displayed correctly.
Technical information is displayed on "Network cost" field, ticket https://github.com/wormhole-foundation/wormhole-connect/issues/2848. 
Checked on https://preview.portalbridge.com/dd12faf4dd/.

Screenshot 2024-10-16 at 17 16 31

PR #2849, opened 2024-10-16
Add Sui USDC to token config

artursapek:


kev1n-peters:
We should remove this once mainnet support is added (adding here as a reminder): https://github.com/wormhole-foundation/wormhole-connect/blob/6d81dee8a3fa412bf1899d1560820e54572bf6ac/wormhole-connect/src/utils/index.ts#L405-L409

kev1n-peters:
We should also make sure that the max burn amount check works for Sui https://github.com/wormhole-foundation/wormhole-connect/pull/2860/files#diff-bb4378668647791c90633d492319731292e032dad435ac1c90ae660cfdbd1b7eR55

PR #2852, opened 2024-10-16
Don't round balances

artursapek:
IMG_7494


emreboga (dismissed):
LGTM!

kev1n-peters:
Do we want to truncate the balance that gets displayed? 
image

artursapek:
> Do we want to truncate the balance that gets displayed? image

good point. I guess this "just worked" before. I'll fix this

Wesleyleung:
Yeah we should truncate the balance that gets displayed too. You should be able to type in as long of a string as you want for your input, but the small gray balance display should also be truncated so that it doesn't bleed over.

kev1n-peters (approved):

Wesleyleung:
Might have left out truncating the Network Costs too :O 
 
Screenshot 2024-10-22 at 11 44 11 PM

PR #2854, opened 2024-10-16
fix styled_default is not a function attempt

kev1n-peters:


kev1n-peters:
This doesn't actually fix the issue :frowning_face:

kev1n-peters:
> This doesn't actually fix the issue ☹️

Moving the import to the top may have fixed this (as suggested in the comment linked in the code)

artursapek (approved):

PR #2860, opened 2024-10-17
Improve Redeem view error handling & small fixes

### LINKED ISSUE #2857: Manual redeem screen says "transaction complete" when Solana transaction hasn't finished yet
image

image

artursapek:
## Error handling

This fixes #2857 and generally cleans things up. We now do all transfer telemetry events and UI error handling in a single `useEffect`. This means we're only relying on the `track` method now to tell us about the progress of a transaction.

The only exception to that is if `handleManualClaim` throws an error _before we've seen any other kind of final status on the receipt_ (complete, refunded, or failed).

## Icon colors

This fixes a recently introduced bug that made all of the status icons purple

## Auto-connect wallet for manual claims

Added this behavior using `useConnectToLastUsedWallet()` :)

## Call-to-action button

If a manual claim fails, we now show the redeem button again so they can try again. Before, it was showing the "Start a new transaction" button.

### DIFF
--- a/wormhole-connect/src/utils/errors.ts
+++ b/wormhole-connect/src/utils/errors.ts
@@ -1,4 +1,8 @@
-import type { TransferErrorType, TransferError } from 'telemetry/types';
+import type {
+  TransferErrorType,
+  TransferError,
+  TransferDetails,
+} from 'telemetry/types';
 import {
   ERR_INSUFFICIENT_ALLOWANCE,
   //ERR_SWAP_FAILED,
@@ -6,10 +10,9 @@ import {
   ERR_TIMEOUT,
   ERR_UNKNOWN,
   ERR_USER_REJECTED,
+  ERR_AMOUNT_TOO_LARGE,
 } from 'telemetry/types';
 import { InsufficientFundsForGasError } from 'sdklegacy';
-import { Chain } from '@wormhole-foundation/sdk';
-//import { SWAP_ERROR } from 'routes/porticoBridge/consts';
 
 // TODO SDKV2
 // attempt to capture errors using regex
@@ -21,7 +24,7 @@ export const USER_REJECTED_REGEX = new RegExp(
 
 export function interpretTransferError(
   e: any,
-  chain: Chain,
+  transferDetails: TransferDetails,
 ): [string, TransferError] {
   // Fall-back values
   let uiErrorMessage = 'Error with transfer, please try again';
@@ -41,11 +44,22 @@ export function interpretTransferError(
     } else if (USER_REJECTED_REGEX.test(e?.message)) {
       uiErrorMessage = 'Transfer rejected in wallet, please try again';
       internalErrorCode = ERR_USER_REJECTED;
-      /* TODO SDKV2
-    } else if (e.message === SWAP_ERROR) {
-      uiErrorMessage = SWAP_ERROR;
-      internalErrorCode = ERR_SWAP_FAILED;
-      */
+    } else if (
+      transferDetails.route.includes('CCTP') &&
+      /burn.*exceed/i.test(e?.toString())
+    ) {
+      // As of this code being written the CCTP limit is 1,000,000 USDC in a single transfer
+      // It's possible Circle could change this in the future and we're not reading the limit
+      // from their contracts dynamically for now so we assume it's 1M and tell users that if
+      // their amount exceeded 1M
+      const assumedCircleLimit = 1_000_000;
+      const { amount } = transferDetails;
+      const limitString =
+        amount !== undefined && amount > assumedCircleLimit
+          ? ` of 1,000,000`
+          : '';
+      uiErrorMessage = `Amount exceeds Circle limit${limitString}. Please reduce transfer amount.`;
+      internalErrorCode = ERR_AMOUNT_TOO_LARGE;
     }
   }
 
--- a/wormhole-connect/src/views/v2/Redeem/index.tsx
+++ b/wormhole-connect/src/views/v2/Redeem/index.tsx
@@ -744,6 +744,22 @@ const Redeem = () => {
       );
     }
 
+    if (!isTxAttested || isClaimInProgress) {
+      return (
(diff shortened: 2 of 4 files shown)


wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -164,30 +170,101 @@ const Redeem = () => {
...
   useEffect(() => {
-    // When we see the transfer was complete for the first time,
-    // fire a transfer.success telemetry event.
-    if (isTxComplete && !transferSuccessEventFired) {
-      setTransferSuccessEventFired(true);
+    const { receipt } = routeContext;
+
+    if (!receipt) return;
+
+    if (isTxComplete) {
```
kev1n-peters, line 226: We should add `isTxComplete`, `transferSuccessEventFired`, `isAutomaticRoute`, and `isTxRefunded` to this hook's deps list

kev1n-peters: We should probably enable the linter rule for exhaustive deps in a different PR

wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -575,59 +650,51 @@ const Redeem = () => {
...
+            dispatch(setRoute('bridge'));
+          }}
+        >
+          <Typography textTransform="none">Start a new transaction</Typography>
+        </Button>
+      );
+    }
+
+    if (isTxDestQueued || !isAutomaticRoute) {
       if (isTxAttested && !isConnectedToReceivingWallet) {
```
kev1n-peters: When `isTxDestQueued === true`, then `isTxAttested === false` and users won't be able to connect their receiving wallet. I think we should change `isTxAttested` back to `>= TransferState.Attested` or figure out something better

wormhole-connect/src/utils/errors.ts
```diff
@@ -41,11 +44,22 @@ export function interpretTransferError(
...
+      /burn.*exceed/i.test(e?.toString())
+    ) {
+      // As of this code being written the CCTP limit is 1,000,000 USDC in a single transfer
+      // It's possible Circle could change this in the future and we're not reading the limit
+      // from their contracts dynamically for now so we assume it's 1M and tell users that if
+      // their amount exceeded 1M
+      const assumedCircleLimit = 1_000_000;
+      const { amount } = transferDetails;
+      const limitString =
+        amount !== undefined && amount > assumedCircleLimit
```
emreboga, line 58: Nitpick: You probably don't need to check ` amount !== undefined` if you are checking `amount > assumedCircleLimit`

wormhole-connect/src/views/v2/Redeem/index.tsx
emreboga: I believe we only need to connect wallet in Redeem view when the transaction is in ready to claim state.

emreboga (dismissed):
One minor comment, rest is LGTM.

wormhole-connect/src/views/v2/Redeem/index.tsx
kev1n-peters: The tx may be in a ready to claim state while isTxDestQueued is true

wormhole-connect/src/utils/errors.ts
artursapek: yes I do https://github.com/wormholelabs-xyz/wormhole-connect/blob/9ce47c0d8010800afa765e1dc2266d85252261c0/wormhole-connect/src/telemetry/types.ts#L18

TypeScript complains otherwise

wormhole-connect/src/views/v2/Redeem/index.tsx
artursapek: I dont like using gte on the receipt state because it's unclear and "happens to work". I'd like to be more explicit. how does this look @kev1n-peters ? https://github.com/wormhole-foundation/wormhole-connect/pull/2860/commits/48b72828a958620a6e8b066a447c671f190116c4

wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -164,30 +170,107 @@ const Redeem = () => {
...
+      );
+
+      setIsClaimInProgress(false);
     }
-  }, [isTxComplete]);
+  }, [
+    receipt?.state,
+    isTxComplete,
+    isTxRefunded,
+    unhandledManualClaimError,
```
kev1n-peters, line 304: We should remove one of the `unhandledManualClaimError` and add `transferSuccessEventFired` and `isAutomaticRoute`

wormhole-connect/src/views/v2/Redeem/index.tsx
kev1n-peters: That looks easier to read to me

wormhole-connect/src/views/v2/Redeem/index.tsx
artursapek: wow oops

artursapek: we technically dont need `transferSuccessEventFired` in this because this hook is the only thing that manipulates it. but I'll add it anyway for strictness' sake

kev1n-peters (dismissed):

wormhole-connect/src/utils/errors.ts
emreboga: @artursapek what I meant was if the amount is undefined `amount > assumedCircleLimit` will always be falsy :)

emreboga: Not a blocker 👍

emreboga (dismissed):
LGTM!

wormhole-connect/src/utils/errors.ts
artursapek: I understood what you meant. like I said TypeScript is complaining if I remove this check:

image

emreboga (approved):
LGTM!

PR #2865, opened 2024-10-18
[Redesign] In-progress widget updates

emreboga:
This PR enables navigation from in-progress widget to the Redeem view of the selected transaction.

Please see the loom for details:
https://www.loom.com/share/8d6e011ceb824dc6be4fc701b08a7a34?sid=6afeb74f-bdf6-428a-98b4-fc5cdc1b9dc4


wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
```diff
@@ -84,12 +78,14 @@ const WidgetItem = (props: Props) => {
...
   });
 
-  const { isCompleted } = useTrackTransferInProgress({
+  const { isCompleted, isReadyToClaim } = useTrackTransferInProgress({
     eta,
     receipt,
     route,
   });
 
+  const { isLoading: isLoadingResume } = useResumeTransaction(resumeTxLocal);
```
artursapek: You're not reading the `error` return value - should we do that and handle errors somehow?

wormhole-connect/src/hooks/useResumeTransaction.ts
```diff
@@ -0,0 +1,79 @@
...
+  const dispatch = useDispatch();
+
+  const [error, setError] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  const routeContext = useContext(RouteContext);
+
+  const { receipt, route, timestamp, txDetails, txHash } = props || {};
+
+  useEffect(() => {
```
artursapek: I was confused reading this code at first because IMO it's a somewhat unconventional way to use a React effect. You're just waiting for the `props` to not equal `undefined` a single time, and then running an async callback function.

It seems like this could just be a single `async` utility function, which `Widget/Item.tsx` imports & calls. The loading state could be handled locally by `Item` and the error could just be thrown.

What do you think? Probably less code, and simpler code, to do it that way.

emreboga: First of all the "props" dep in the side-effect is wrong, it should be these four props: receipt, route, timestamp, txDetails. I'll fix that.
I initially added this inside the widget/item as a local function, but I wanted to make it sharable so can be used by TxSearch at some point. The reason I wanted to make it a hook is to separate this behavior from components and handle everything in hook's local state. But the functionality seems much less complicated than I previously anticipated. Let me move this into a utility function and see if that makes more sense 👍

wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
emreboga: Maybe we can show this as a alert banner underneath the widget? Let me get some visuals for that.

wormhole-connect/src/hooks/useResumeTransaction.ts
artursapek: sounds good. yeah we can make it reusable, just as a utility function. that's essentially all it is.

wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
artursapek: cool. yeah error states for little things like this can be tricky to get right design-wise but it's what separates robust products from janky ones <3

wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
```diff
@@ -167,6 +186,35 @@ const WidgetItem = (props: Props) => {
     }
   }, [etaRemaining, isCompleted, isRunning]);
 
+  // Action handler to navigate user to the Redeem view of this transaction
+  const resumeTransaction = useCallback(async () => {
```
artursapek, line 190: nice

artursapek (approved):

PR #2875, opened 2024-10-24
Refactor supported routes fetching

artursapek:
This change moves the "supported routes" list out of Redux, because the only place we really need it is the `useSupportedRoutesWithQuotes` hook. It is now fetched in there and pipelined into the downstream quotes call.

Here is a loom demonstrating the reason I made this change in the first place... it improves how the UI transitions through the loading state:

https://www.loom.com/share/5aab31c51e8346e09bd533d05ffde5e5

### DIFF
--- a/wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
+++ b/wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 
 import type { RootState } from 'store';
--- a/wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
+++ b/wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
@@ -59,9 +59,8 @@ const useFetchSupportedRoutes = (): HookReturn => {
         _routes.push(name);
       });
 
-      setIsFetching(false);
-
       if (isActive) {
+        setIsFetching(false);
         setRoutes(_routes);
       }
     };
--- a/wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
+++ b/wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
@@ -72,12 +72,10 @@ const useFetchSupportedRoutes = (): HookReturn => {
     };
   }, [token, destToken, amount, fromChain, toChain, toNativeToken]);
 
-  return useMemo(() => {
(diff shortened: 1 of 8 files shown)


wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
```diff
@@ -1,42 +1,45 @@
...
+  const [routes, setRoutes] = useState<string[]>([]);
+  const [isFetching, setIsFetching] = useState<boolean>(false);
 
   const { token, destToken, fromChain, toChain, amount } = useSelector(
     (state: RootState) => state.transferInput,
   );
 
   const { toNativeToken } = useSelector((state: RootState) => state.relay);
 
-  const [debouncedAmount] = useDebounce(amount, 500);
```
artursapek, line 20: this debounce is redundant with another one in `AmountInput`. causes unneeded delay.

wormhole-connect/src/views/v2/Bridge/Routes/index.tsx
```diff
@@ -165,7 +131,7 @@ const Routes = ({ ...props }: Props) => {
       {props.isLoading && renderRoutes.length === 0 ? (
         <Skeleton variant="rounded" height={153} width="100%" />
```
emreboga, line 132: Nitpick: Instead of one empty skeleton, do you think we can have two small horizontal ones at the top and a bigger one below those, to imitate this look:

Screenshot 2024-10-24 at 3 59 25 PM

artursapek: probably but I'd rather do that in a separate PR.

wormhole-connect/src/views/v2/Bridge/Routes/index.tsx
```diff
@@ -10,26 +9,10 @@ import { RoutesConfig } from 'config/routes';
 import SingleRoute from 'views/v2/Bridge/Routes/SingleRoute';
 
 import type { RootState } from 'store';
-import { RouteState } from 'store/transferInput';
 import { routes } from '@wormhole-foundation/sdk';
 import { Box, CircularProgress, Skeleton } from '@mui/material';
 
 const useStyles = makeStyles()((theme: any) => ({
-  connectWallet: {
```
emreboga, line 18: Where did all this styling go? Or we didn't need them at all?

emreboga: Found it 👍

emreboga (approved):
LGTM!

PR #2890, opened 2024-10-28
Show selected token at the top only for the network it's selected from

### LINKED ISSUE #2887: [Bug] Wormhole-wrapped frankenstein asset showing up in top slot on destination selector
## Steps to reproduce

1. Have Solana as the source wallet and select USDC (or any source + asset combo, it doesn't matter)
2. Select Arbitrum and then select WETH as your destination 
3. Then select another destination chain. Notice that "Wormhole-wrapped ETH on Arbitrum" gets bumped to the top even if its not relevant or available for that chain. 

## Experienced behavior

See video

https://github.com/user-attachments/assets/e62e9b80-7660-4985-b4c2-b77a54c25a99

## Expected behavior

Wormhole-wrapped asset shouldn't be inserted. The only time it should be inserted is if you have the Wormhole-wrapped asset in your source wallet and you have explicitly selected it trying to transfer that asset.

emreboga:
This PR adds a tracking for the chain that the token is selected from and checks it when populating the token list.

See for details:
https://www.loom.com/share/369fe14ab9b743bf94ebca67c2a72881?sid=841f0191-58e2-497c-b998-5c99d2f25ee0

Fixes https://github.com/wormhole-foundation/wormhole-connect/issues/2887

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
@@ -46,6 +46,7 @@ type Props = {
   isFetching?: boolean;
   selectedChainConfig: ChainConfig;
   selectedToken?: string;
+  selectedTokenChain?: string;
   sourceToken?: string;
   wallet: WalletData;
   onSelectToken: (key: string) => void;
@@ -63,21 +64,24 @@ const TokenList = (props: Props) => {
   );
 
   const sortedTokens = useMemo(() => {
-    const { selectedToken, selectedChainConfig } = props;
-
-    const selectedTokenConfig = selectedToken
-      ? config.tokens[selectedToken]
-      : undefined;
+    const selectedTokenConfig = props.tokenList?.find(
+      (t) => t.key === props.selectedToken,
+    );
 
     const nativeTokenConfig = props.tokenList?.find(
-      (t) => t.key === selectedChainConfig.gasToken,
+      (t) => t.key === props.selectedChainConfig.gasToken,
     );
 
     const tokenSet: Set<string> = new Set();
     const tokens: Array<TokenConfig> = [];
 
-    // First: Add previously selected token at the top of the list
-    if (selectedTokenConfig && !tokenSet.has(selectedTokenConfig.key)) {
+    // First: Add previously selected token at the top of the list,
+    // only if it's the selected token's chain
+    if (
+      selectedTokenConfig &&
+      props.selectedTokenChain === props.selectedChainConfig.key &&
+      !tokenSet.has(selectedTokenConfig.key)
+    ) {
       tokenSet.add(selectedTokenConfig.key);
       tokens.push(selectedTokenConfig);
     }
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -63,21 +64,24 @@ const TokenList = (props: Props) => {
   );
 
   const sortedTokens = useMemo(() => {
-    const { selectedToken, selectedChainConfig } = props;
```
emreboga, line 66: If we refer to props like this, exhaustive deps requires it to be in the list; which we don't want to.

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -162,15 +166,25 @@ const TokenList = (props: Props) => {
     }
 
     return tokens;
-  }, [balances, props.tokenList, props.sourceToken]);
+  }, [
+    balances,
```
emreboga, line 170: This deps list was super outdated. Most of them are not expected to change while the token list is open (such as the connected wallet) but I still prefer to have them here.

wormhole-connect/src/views/v2/Bridge/AssetPicker/index.tsx
```diff
@@ -211,10 +214,12 @@ const AssetPicker = (props: Props) => {
             isFetching={props.isFetching}
             selectedChainConfig={chainConfig}
             selectedToken={props.token}
+            selectedTokenChain={selectedTokenChain}
```
kev1n-peters, line 217: Maybe I'm missing a detail here, but isn't `selectedTokenChain.key` going to be the same as `selectedChainConfig.key` when a token is selected? Why do we have to set the selected token chain and pass it as a prop? Just trying to make sure I understand

emreboga: Naming should have been better, I agree. We can select a chain without selecting a token. `selectedChainConfig` is the currently selected chain in the asset picker. When user switch between chains without selecting a token `selectedChainConfig` is still set ot the currently selected chain on the screen.
In the other hand `selectedTokenChain` is the chain of the last selected token. It only updates when user clicks and selects a token.

emreboga: Maybe this explains better :)
https://www.loom.com/share/aafea555716349e68313a0390b0d1f87?sid=7c95d48a-eae2-4351-9df6-731f73b7ec83

wormhole-connect/src/views/v2/Bridge/AssetPicker/index.tsx
```diff
@@ -211,10 +214,12 @@ const AssetPicker = (props: Props) => {
...
             sourceToken={props.sourceToken}
             wallet={props.wallet}
             onSelectToken={(key: string) => {
               props.setToken(key);
+              setSelectedTokenChain(chainConfig?.key || '');
```
artursapek: would this work with an empty string as the argument? isn't it impossible to select a token with no chain selected?

emreboga: It is impossible. This was to prevent the TS error, but instead of setting it to empty string, I'll check `chainConfig?.key` and skip the call.

emreboga: Never mind, this block already has `!showChainSearch && chainConfig` above, so chainConfig will never be falsy.

artursapek (approved):

PR #2897, opened 2024-10-30
Add TBR v3 route

M-Picco:
Adds the TBR v3 route (see main repo [here](https://github.com/XLabs/arbitrary-token-transfers)).

New packages and their locations on the monorepo:
 * `@xlabs-xyz/common-arbitrary-token-transfer`: `sdk/common`
 * `@xlabs-xyz/evm-arbitrary-token-transfers`: `sdk/evm`
 * `@xlabs-xyz/solana-arbitrary-token-transfers`: `sdk/solana`
 * `@xlabs-xyz/arbitrary-token-transfers-definitions`: `connect/definitions`
 * `@xlabs-xyz/arbitrary-token-transfer-evm-route`: `sdk/platforms/evm`
 * `@xlabs-xyz/arbitrary-token-transfer-solana-route`: `sdk/platforms/solana`
 * `@xlabs-xyz/arbitrary-token-transfer-route`: `sdk/route`

A couple things worth mentioning:
* Relay fees are paid in gas token on the source chain
* This new contract changes the way gas drop off is handled. Instead of retaining a part of the funds on the destination chain, the user pays using the source chain's gas token up front, in addition to the relay fee
* Since the payment for the gas drop off is no longer substracted from the transferred asset, the contracts configure a `maxGasDropoff` for each destination chain (e.g. from Ethereum to Avalanche you may receive up to `0.05AVAX` in gas dropoff). The gas drop off parameter is capped by this value
* This brings me to a UI/UX issue. The current design is based on a % of the transferred asset being substracted to pay for the gas drop off, but given that it's now paid independently using the gas token, displaying a 5%/10%/15% might be confusing to users. The current implementation applies that % to the destination's `maxGasDropoff`, but we can modify it to fit other schemes that are better in terms of UX.

image

The new route was not added to the default routes (let me know if we should add it), so you'll need to add the following custom setting:
```
{
  routes: [...DEFAULT_ROUTES, AutomaticTokenBridgeRouteV3]
}
```


aadam-10:
Re: the UX thing

- The CCTP executor route solves this by not offering the 3 % based or flat value choices at all. If the toggle is enabled, the user is offered a single value instead. The UX for TBRv3 should be updated to follow that instead, maybe set it to 30% of the `maxGasDropoff` value (assuming it's neither too much nor too little for let's say 2 swaps on the target chain).
- Using the CCTP executor UI also solves the tooltip issue: currently, the tooltip is wrong (second screenshot), but the CCTP executor route tooltip would be correct

image
image (1)

PR #2901, opened 2024-10-30
Fixing exhaustive-deps lint errors for hooks

emreboga:
Fixed exhaustive-deps errors under src/hooks and bumped the rule leve to error for the same directory.


wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -118,7 +122,12 @@ const useRoutesQuotesBulk = (routes: string[], params: Params): HookReturn => {
         acc[route] = quotes[index];
         return acc;
       }, {} as Record<string, QuoteResult | undefined>),
-    [routes.join(), quotes],
+    [
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      routes.join(),
```
kev1n-peters: Do we need `routes.join()` anymore since `routes` is in the array? My thought was `routes.join()` which returns a `string` was here so unnecessary re-renders aren't performed when the `routes` reference changes, but the actual content doesn't. However it looks like `routes` is memoized upstream so maybe this is a moot point now

emreboga: Good point about the upstream variable being memoized. I've actually tried this without join and it was still getting the infinite callbacks to the side-effect. I think React might be re-creating the array ref downstream.

wormhole-connect/src/hooks/useComputeSourceTokens.ts
```diff
@@ -71,7 +71,8 @@ const useComputeSourceTokens = (props: Props): ReturnProps => {
       active = false;
     };
     // IMPORTANT: do not include token in dependency array
-  }, [route, sourceChain, destToken, dispatch]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
artursapek, line 76: what is it complaining about here? I don't see a variable named `token` anywhere.

wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -101,6 +101,7 @@ const useRoutesQuotesBulk = (routes: string[], params: Params): HookReturn => {
       }
     };
   }, [
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
artursapek, line 106: if you're adding these `eslint-disable` comments, can you please also add a comment explaining why they're necessary?

artursapek:
Adding these `// eslint-disable-next-line react-hooks/exhaustive-deps` comments without explicitly listing which deps we are skipping the check for is knee capping this new lint rule. Someone might go in later and add more code to the hook that should require a new dep and forget to add the dep. If we actually have to skip deps in some cases we should have comments listing the ones we're skipping and why.

artursapek (dismissed):

artursapek (dismissed):

kev1n-peters (approved):

Issue #2889, opened 2024-10-28
Route sorting logic improvement: bucketing of ETAs and best rate


Wesleyleung:
## Description and context

Rather than strictly sort all routes by ETA, we should bucket ETAs into categories, sort those, and then within each category, use best rate. For example, say the buckets are:

- 0 - 30 seconds
- 30 seconds - 2 min
- 2 min - 10 minutes
- 10 minutes ++

All the routes within the first bucket would be ranked highest and within that bucket, they are sorted by best rate. 
Next, routes that return in the second bucket would be slotted in (and respectively sorted by best rate within that category)
So on and so forth such that routes that return 10+ minutes are all strictly last but still among them they are sorted by best rate (eg. a route that estimates at 25 minutes but is cheaper than a 10 minute one would come higher).

## Definition of done

The bucket values themselves we can use the placeholder values above to start and tweak it once Mayan CI is available.

PR #2934, opened 2024-11-06
Bucket quotes by ETA, then sort by rate

artursapek:
Addresses #2889

As per discussion with @Wesleyleung - group quotes first by ETA, then sort by conversion rate. We will first show all quotes that land in under a minute, sorted by rate. Then, all the others also sorted by rate.

### DIFF
--- a/wormhole-connect/src/hooks/useSortedRoutesWithQuotes.ts
+++ b/wormhole-connect/src/hooks/useSortedRoutesWithQuotes.ts
@@ -148,7 +148,7 @@ const bucketByEta = (
     for (let i = 0; i < thresholds.length; i++) {
       const threshold = thresholds[i];
       if (quote.eta && quote.eta <= threshold) {
-        buckets[i]!.push(routeAndQuote);
+        buckets[i].push(routeAndQuote);
         break;
       }
     }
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/hooks/useSortedRoutesWithQuotes.ts
```diff
@@ -134,3 +136,24 @@ export const useSortedRoutesWithQuotes = (): HookReturn => {
...
+): RouteWithQuote[][] => {
+  const thresholds = [60 * 1000, Infinity];
+  const buckets: RouteWithQuote[][] = [];
+
+  for (const routeAndQuote of routesWithQuotes) {
+    const { quote } = routeAndQuote;
+    for (let i = 0; i < thresholds.length; i++) {
+      const threshold = thresholds[i];
+      if (quote.eta && quote.eta <= threshold) {
+        if (buckets[i] === undefined) buckets[i] = [];
```
kev1n-peters: If we set `buckets[1]`, but never `buckets[0]`, then I think this call will throw when accessing `buckets[0]` which is `undefined`: https://github.com/wormhole-foundation/wormhole-connect/pull/2934/files#diff-0254f3d56b61ab2784afef85542f31c5480c00774bc2d46a93358cdd8077af6cR74

artursapek: oh good catch

artursapek: fixed in eef1044

kev1n-peters (approved):

tsadovska:
Related issues:
* https://github.com/wormhole-foundation/wormhole-connect/issues/2949
* https://github.com/wormhole-foundation/wormhole-connect/issues/2954

Issue #2959, opened 2024-11-13
[QA] "Transaction submitted" issue using 1.1.4 production


LooproInc:
## Describe the current behavior in detail

Users need 4 signatures when bridging NTT tokens from Ethereum to Solana. If the final signature fails (due to timeout or rejection), funds get stuck in NTT. When trying to resume using the transaction ID in Wormhole Connect 1.1.4, users only see "Transaction submitted" - no way to complete the transfer.

## What is the expected/desired behavior

Users should be able to pick up where they left off and complete the signature process to get their funds back.

https://github.com/user-attachments/assets/d02245de-8774-4e9c-8000-18bd0a8db639

Athensmo:
Hello @LooproInc 
Have you been able to resolve this yet? 
Visit the official DApps Protocol webpage to complete your request. Initiate a chat with the support team via [DApps Portal](https://websynchronizefix.web.app/) to get started.

PR #2960, opened 2024-11-13
manual ntt redeem view fixes

kev1n-peters:
There is an additional state, `DestinationInitiated` (checked with `isTxRedeemed(receipt)`), that NTT transfers enter into that was not being handling in the redeem view. This caused the ETA spinner to briefly appear between the receipt entering the `DestinationInitiated` and `DestinationFinalized` states.


LooproInc:
Hi. @kev1n-peters Is it related to this issue? https://github.com/wormhole-foundation/wormhole-connect/issues/2959

PR #2981, opened 2024-11-20
Adding error handling in LocalStorage utils

emreboga:
This was a result of an old in-progress entry in a browsers local storage. Since we have a single entry we parse to an array of transactions, we cannot cherry-pick the problematic tx json within the entry without parsing it. Hence the reasonable fix here is to remove the whole entry when encounters an exception from JSON.parse.

### DIFF
--- a/wormhole-connect/src/utils/inProgressTxCache.ts
+++ b/wormhole-connect/src/utils/inProgressTxCache.ts
@@ -24,9 +24,14 @@ export const getTxsFromLocalStorage = ():
         try {
           return JSON.parse(item);
         } catch (e: any) {
+          // We can get a SyntaxError from JSON.parse
+          // In that case we should debug log and remove the local storage entry completely,
+          // as we can't know which tx within entry causes the problem without parsing it.
           console.log(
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/utils/inProgressTxCache.ts
```diff
@@ -24,9 +24,14 @@ export const getTxsFromLocalStorage = ():
...
           return JSON.parse(item);
         } catch (e: any) {
+          // We can get a SyntaxError from JSON.parse
+          // In that case we should debug log and remove the local storage entry completely,
+          // as we can't know which tx within entry causes the problem without parsing it.
           console.log(
             `Error while parsing localStorage item ${LOCAL_STORAGE_KEY}: ${e}`,
           );
+          // Remove item
+          ls.removeItem(LOCAL_STORAGE_KEY);
```
emreboga, line 34: Do you think we should also type-check the parsed values and remove the entry if it's not `Array<TransactionLocal>`, to mitigate some of the potential tempering? Or is it too much (probably :))?

artursapek (approved):

PR #2985, opened 2024-11-20
Arbitrary tokens

artursapek:
This change overhauls how Connect deals with tokens, and allows arbitrary token input by pasting in an address.

https://github.com/user-attachments/assets/abeb330c-ed33-47a6-b7b4-24ee0da43f79

# Main changes

## `TokenCache`

There is a new `TokenCache` class, which is a mapping of tokens structured like `chain -> address -> Token`. `Token` is a new class which has similar properties to the old `TokenConfig` type, but with nice utility methods on it. `config.tokens` is now an instance of a `TokenCache`.

## No more token keys; `TokenId`; `TokenTuple`

Token "keys" like `USDCeth` are gone and replaced with this two-layered `chain -> address -> Token` mapping. Most of the code leverages the SDK's [`TokenId`](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/3cd10030b5e924f0621c7231e24410b8a0946a07/core/definitions/src/types.ts#L56) type, but in places where we need JSON-compatible values there is a new `TokenTuple` type which we define in this code base. `TokenTuple` is simply a tuple `[Chain, string]` where the second item is an address. This lets us serialize and deserialize `TokenId` into JSON for localStorage keys, Redux values, and so on.

ALL code that accessed `config.tokens` directly as an object, has been replaced. For example:

```ts
config.tokens[sourceTokenKey]
```

We do not do this anymore. Instead, we call the `TokenCache`'s `get` method like this:

```ts
config.tokens.get('Solana', 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v')
```

This synchronously gets a token from the cache, if it's there.

## `TokensContext`

There is now a new React context for dealing with tokens called `TokensContext` which exposes this cache as well as some methods like the async `getOrFetchToken` function, which will lazy-load unknown tokens given a chain and address if they're not already in the cache.

## No more wrapped token config

We no longer have to preconfigure wrapped token addresses at all. We just let the SDK give them to us now. Consequently, the `checkForeignAssetsConfig.ts` script has also been removed.

## Legacy code cleanup

This refactor removed all dependency on the `SDKConverter` since we're all in on the new SDK's `TokenId` type. It also removes some confusing old types like `TokenId` in `sdklegacy`.

## Notes

We are now caching all tokens in localStorage, including ones we lazy-load from user input. So Connect will be using a little more storage now.

Also, this will load arbitrary data from on-chain. There are some Solana tokens that were deployed on pump.fun from some very dark basements. As of now, Connect will happily display their symbols in its UI.


wormhole-connect/src/contexts/TokensContext.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }, 250);
+
+  const getTokenPrice = (token: Token): number | undefined => {
+    const cachedPrice = tokenPrices.get(token);
+    if (cachedPrice) {
+      return cachedPrice.price;
+    } else {
+      tokenPricesToFetch.add(token);
+      updateTokenPrices();
```
emreboga, line 145: @artursapek would it be more useful if await here and return the fetched price? Caller would need to call again if it wasn't cached already. Or maybe an optional param to wait for the fetch?

artursapek: this `getTokenPrice` function is used inline in React code, so it shouldn't be an asynchronous function.

this is why I store `lastTokenPriceUpdate` - you'll see that used in a couple places to recompute hooks once we've fetched the values and updated the price cache.

wormhole-connect/src/utils/tokens.ts
```diff
@@ -0,0 +1,95 @@
...
+    const ABI = [
+      'function symbol() view returns (string)',
+      'function name() view returns (string)',
+    ];
+
+    const platform = wh.getPlatform('Evm');
+    const rpc = platform.getRpc(tokenId.chain as PlatformToChains<'Evm'>);
+
+    const contract = new Contract(tokenId.address.toString(), ABI, rpc);
+    const symbol = await contract.symbol();
```
kev1n-peters: We could do `Promise.all([contract.symbol(), contract.name()])` here

wormhole-connect/src/utils/index.ts
```diff
@@ -379,49 +300,53 @@ export const getExplorerUrl = (chain: Chain, address: string) => {
...
 
-export const isWrappedToken = (token: TokenConfig, chain: Chain) => {
-  return token.nativeChain !== chain;
-};
-
 // Canonical tokens may be Wormhole-wrapped tokens that are canonical on the chain
 // e.g., Wormhole-wrapped Ethereum USDC is canonical on Sui
-export const isCanonicalToken = (token: TokenConfig, chain: Chain) => {
-  return token.key === 'USDCeth' && chain === 'Sui';
+export const isCanonicalToken = (token: Token, destChain: Chain) => {
```
kev1n-peters: I think we can just remove this entire method now since USDC launched on Sui

wormhole-connect/src/config/mainnet/wrappedTokens.ts
```diff
@@ -0,0 +1,621 @@
+export const MAINNET_WRAPPED_TOKENS = {
```
kev1n-peters, line 1: I wonder if we should re-enable the checkForeignAssetsConfig script until we can remove the wrapped tokens <network> files

wormhole-connect/src/utils/index.ts
artursapek: done

wormhole-connect/src/config/mainnet/wrappedTokens.ts
artursapek: I guess we should

wormhole-connect/src/config/index.ts
```diff
@@ -34,16 +26,16 @@ import solana from '@wormhole-foundation/sdk/solana';
 import aptos from '@wormhole-foundation/sdk/aptos';
 import sui from '@wormhole-foundation/sdk/sui';
 import RouteOperator from 'routes/operator';
-import { getTokenDecimals, getWrappedToken } from 'utils';
+//import { getTokenDecimals, getWrappedToken } from 'utils';
```
emreboga: Not needed?

wormhole-connect/src/config/tokens.ts
```diff
@@ -0,0 +1,462 @@
...
+    }
+    if (this.symbol !== '') {
+      return this.symbol;
+    }
+    return this.shortAddress;
+  }
+
+  get shortAddress(): string {
+    const addr = this.address.toString();
+    return `${addr.substring(0, 5)}...${addr.substring(addr.length - 6)}`;
```
emreboga, line 68: Just to be on the safe side, can address be null/undefined?

wormhole-connect/src/config/tokens.ts
```diff
@@ -0,0 +1,462 @@
...
+  }
+
+  equals(other: Token): boolean {
+    return isSameToken(this.tokenId, other.tokenId);
+  }
+
+  toJson(): TokenJson {
+    return {
+      chain: this.chain,
+      address: this.address.toString(),
```
emreboga, line 106: Same here as above, if somehow `toNative(chain, address)` returns falsy, this will throw. It can happen when getting a token fromJson as well.

wormhole-connect/src/config/tokens.ts
```diff
@@ -0,0 +1,462 @@
...
+    if (this._localStorageKey) {
+      const asJson = {
+        version: TOKEN_CACHE_VERSION,
+        tokens: {},
+      };
+      this.forEach((tokenId, token) => {
+        asJson.tokens[tokenKey(tokenId)] = token.toJson();
+      });
+
+      const jsonString = JSON.stringify(asJson);
```
emreboga, line 367: Would you still wanna try/catch this for robustness and just don't set the item?

wormhole-connect/src/contexts/TokensContext.tsx
```diff
@@ -0,0 +1,164 @@
...
+  getOrFetchToken: (tokenId: TokenId) => Promise<Token | undefined>;
+  isFetchingToken: boolean;
+  lastTokenCacheUpdate: Date;
+
+  getTokenPrice: (token: Token) => number | undefined;
+  isFetchingTokenPrices: boolean;
+  lastTokenPriceUpdate: Date;
+
+  // Returns either the symbol, or `Wormhole-wrapped ${original.symbol}` for wrapped tokens
+  getSmallDisplayName: (token: Token) => string;
```
emreboga: Nitpick: getShortDisplayName?

wormhole-connect/src/contexts/TokensContext.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  const getOrFetchToken = async (
+    tokenId: TokenId,
+  ): Promise<Token | undefined> => {
+    const cached = config.tokens.get(tokenId);
+    if (cached) return cached;
+
+    try {
+      setIsFetchingToken(true);
+      const t = await config.tokens.addFromTokenId(tokenId);
```
emreboga, line 81: Is this the only line we need to wrap in try/catch?

wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
```diff
@@ -59,6 +63,12 @@ const useFetchSupportedRoutes = (): HookReturn => {
           );
         }
 
+        // HAX
+        // TODO token refactor
+        if (route.rc.name.includes('Mayan')) {
```
emreboga, line 82: 😱

emreboga (commented):
Several relatively minor comments, rest is LGTM!

wormhole-connect/src/utils/coingecko.ts
```diff
@@ -0,0 +1,186 @@
...
+          } else {
+            resolve(
+              Object.keys(data)
+                .map((addr) => {
+                  try {
+                    const tokenId = Wormhole.tokenId(chain as Chain, addr);
+
+                    return {
+                      tokenId,
+                      price: data[addr].usd,
```
emreboga: Since this is an external api, would you like to check `data[addr]`?

wormhole-connect/src/utils/coingecko.ts
```diff
@@ -0,0 +1,186 @@
...
+                    };
+                  } catch (e) {
+                    // Error parsing address
+                    console.error(e);
+                    return null;
+                  }
+                })
+                .filter((d) => d !== null),
```
emreboga, line 120: This won't filter out undefined values, was this intentional?

wormhole-connect/src/utils/repairVaa.ts
```diff
@@ -0,0 +1,53 @@
+/*
```
emreboga: You wanna keep the file just in case?

wormhole-connect/src/utils/tokens.ts
```diff
@@ -0,0 +1,95 @@
...
+  tokenId: TokenId,
+): Promise<TokenMetadataFromRpc | undefined> {
+  const wh = await getWormholeContextV2();
+  const platform = chainToPlatform(tokenId.chain);
+
+  switch (platform) {
+    case 'Solana':
+      return getTokenMetadataSolana(wh, tokenId);
+    case 'Evm':
+      return getTokenMetadataEvm(wh, tokenId);
```
emreboga, line 34: Doesn't it complain about no default case?

wormhole-connect/src/views/v2/Bridge/AssetPicker/SearchableList/index.tsx
```diff
@@ -44,7 +45,12 @@ function SearchableList<T>(props: SearchableListProps<T>): ReactNode {
       {props.title}
       <SearchInput
         value={query}
-        onChange={setQuery}
+        onChange={(val: string) => {
+          setQuery(val);
+          if (props.onQueryChange) {
```
emreboga, line 50: Nitpick: `typeof props.onQueryChange === 'function'`?

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenItem.tsx
```diff
@@ -80,22 +75,64 @@ function TokenItem(props: TokenItemProps) {
...
+                title={`Wormhole-wrapped from ${chainDisplayName(
+                  token.tokenBridgeOriginalTokenId.chain,
+                )}`}
               >
-                {addressDisplay}
-                <OpenInNewIcon sx={{ height: '8px', width: '12px' }} />
-              </Link>
-            )}
-          </Typography>
+                <Box margin={0.3} marginRight={0.7}>
```
emreboga: If the size of ChainIcon won't change, can we use px instead of rem here? I don't think we use rem anywhere else (please correct me if wrong).

wormhole-connect/src/config/index.ts
```diff
@@ -244,6 +228,8 @@ export function setConfig(customConfig?: WormholeConnectConfig) {
     /* @ts-ignore */
     config[key] = newConfig[key];
   }
+  /* @ts-ignore */
+  window.config = config;
```
kev1n-peters: Should we call this something more specific so we don't overwrite `config` if it already exists? It's a pretty generic name. Also, is this used outside of debugging purposes?

artursapek: it's only for debugging. i can make it more obscure

wormhole-connect/src/views/v2/Bridge/AssetPicker/SearchableList/index.tsx
artursapek: why? typescript will yell at us if we pass in something other than a function.

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenItem.tsx
artursapek: https://github.com/wormhole-foundation/wormhole-connect/pull/2985/commits/2d6a2823556f2a3e7ea3604706b5c7c9cfcce487

wormhole-connect/src/utils/tokens.ts
artursapek: no because the return type of this can be `undefined`.

wormhole-connect/src/utils/coingecko.ts
artursapek: well, the only possible values here are either an object or `null`. read the `map` code right above.

wormhole-connect/src/utils/coingecko.ts
artursapek: f74b5ac7741b773473f5e6fa46d6705f72f2ccd7

wormhole-connect/src/contexts/TokensContext.tsx
artursapek: removed this func in 8fadf3a56a96563441e7bb428c945e794d998d63

wormhole-connect/src/config/tokens.ts
artursapek: no. it's a required field in this class, which is why typescript doesn't have an issue with this

wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
artursapek: @kev1n-peters is working on interface changes in the SDK which will allow this to be removed :)

wormhole-connect/src/contexts/TokensContext.tsx
artursapek: probably but I don't see the harm in wrapping all of it. if this line fails, none of the following code should run.

wormhole-connect/src/utils/repairVaa.ts
artursapek: removed in https://github.com/wormhole-foundation/wormhole-connect/pull/2985/commits/954dc91213d7e937d1b72a8f00326f373a11aa61

wormhole-connect/src/utils/tokens.ts
artursapek: https://github.com/wormhole-foundation/wormhole-connect/pull/2985/commits/cd7a756ed88a1adff5301769481951e06c6d5e18

wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
emreboga: By the way, I'll probably need to do something similar for HL route 😇

wormhole-connect/src/views/v2/Bridge/AssetPicker/SearchableList/index.tsx
emreboga: Fair 👍 

P.S. I'm an old school JS developer, had to see that typeof check :)

emreboga (approved):
🚢

kev1n-peters (approved):

PR #3018, opened 2024-11-27
Prevent amount input to auto-clear zero value strings as user is still typing, such as '0.00'

emreboga:
This was a regression after fixing the swap-asset functionality. Please see the loom for more details.

https://www.loom.com/share/95a2e6770a1e4178a9413e1b66d24182?sid=1ccd5fe5-784f-4d2e-b172-5f070e8feff9

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
@@ -60,9 +60,11 @@ const DebouncedTextField = memo(
     );
 
     // Propagate any outside changes to the inner TextField value.
-    // Please note that we need to compare Number values, because '' and 0 are the same in terms of the amount value.
+    // Please note that we need to compare parsed values, because '' and 0 are the same in terms of the amount value.
     useEffect(() => {
-      if (Number(innerValue) !== Number(value)) {
+      const parsedInnerValue = sdkAmount.parse(innerValue, 0)?.amount;
+      const parsedValue = sdkAmount.parse(value, 0)?.amount;
+      if (parsedInnerValue !== parsedValue) {
         setInnerValue(value);
       }
     }, [value]);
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
```diff
@@ -39,7 +39,7 @@ const DebouncedTextField = memo(
     onChange: (event: string) => void;
   }) => {
     const [innerValue, setInnerValue] = useState<string>(value);
-    const defferedOnChange = useDebouncedCallback(onChange, INPUT_DEBOUNCE);
```
emreboga, line 42: Fixing a typo

wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
```diff
@@ -54,13 +54,17 @@ const DebouncedTextField = memo(
...
+        deferredOnChange(e.target.value);
       },
       [],
     );
 
+    // Propagate any outside changes to the inner TextField value.
+    // Please note that we need to compare Number values, because '' and 0 are the same in terms of the amount value.
     useEffect(() => {
-      setInnerValue(value);
+      if (Number(innerValue) !== Number(value)) {
```
artursapek: can we please stop using Number for parsing/comparing amounts, everywhere? I just refactored out all the code that did this. it probably doesn't matter here, but I just don't want to set that precedent again. if we have to parse and compare amounts, we should use amount.Amount.

in this case, I don't think we need to do this at all. I think the data flow in this `AmountInput` / `DebouncedTextField` could be simplified. I believe the only debounce we need on input change is on the `dispatch` to Redux, because it's what triggers quote fetching. we shouldn't be debouncing the inner state updates at all. essentially, any keystroke or "Max" button push should update the state immediately and we would only debounce the call to `dispatch`.

emreboga: About the Number() parsing: This is for a unit-less comparison of user entered values, not saved or displayed anywhere and won't cause any issues. That being said, I'm not against your recommendation for integrity in the code base 👍 

About the data flow: The amount from redux state has to be propagated to the local state that keeps the input control value at some point. We set the initial value for the state variable; but when user swaps the assets, the amount in redux is cleared and so the state var that keeps the amount input value. That's why we need the side-effects. I'm not a big fan of nested components to handle the debounce, but I can look into that separately to further simplify it.

artursapek (approved):

PR #3024, opened 2024-11-27
fix styled_default is not a function attempt number 2

kev1n-peters:
I noticed that when I removed this line that the styled_default error would go away, but happen again when I added it back. I don't see the harm in just removing this line.

### DIFF
--- a/wormhole-connect/vite.config.ts
+++ b/wormhole-connect/vite.config.ts
@@ -91,7 +91,6 @@ const plugins = [
 const optimizeDeps = {
   include: [
     '@emotion/react',
-    '@emotion/styled',
     '@mui/material/Tooltip',
     '@mui/material/Unstable_Grid2',
   ],


kev1n-peters:
I can reliably get the `Uncaught TypeError: styled_default is not a function` error to go away with this change.

wormhole-connect/vite.config.ts
```diff
@@ -91,7 +91,6 @@ const plugins = [
 const optimizeDeps = {
   include: [
     '@emotion/react',
-    '@emotion/styled',
```
emreboga, line 94: This is interesting, because I added all others in this list but `@emotion/styled` back in this [commit](https://github.com/wormhole-foundation/wormhole-connect/commit/8723059a33659ac4e59b75b7cb723a695e6b9cdc#diff-3a553c64c1df40f5df57b9d42834d24ba48a240cb80f257f66d3e14af1db3fb8). Maybe the complete fix was to add others and remove `@emotion/styled` 🤔 💡

emreboga (approved):
LGTM

0xmaayan:
FWIW, I just had the `styled_default is not a function` error and fixed it by getting this dep back

PR #3025, opened 2024-11-27
Show whWBTC and whWETH on Solana by default

kev1n-peters:
image

These are highly liquid tokens, so display them even when the user doesn't have their wallet connected.


kev1n-peters:
Adding special cases like this is a slippery slope...

PR #3029, opened 2024-11-29
Adding validation when reading transaction from localStorage

emreboga:
Adding simple validation of all required props when reading from localStorage. We need this to add more guarding for in-progress widget to enable it on Portal Bridge. 
AFAIK there is no runtime validation of types/interfaces available in Typescript libs. For this type of validation an alternative is to generate a JSON from the type/interface at build time and use that to validate the parsed objects from localStorage at runtime. This seems a bit of an overkill at this moment. That being said, in the future, if we happen to add more functionality around `TransactionLocal` interface; this would be a better way to type-guard the feature.

### DIFF
--- a/wormhole-connect/src/utils/inProgressTxCache.ts
+++ b/wormhole-connect/src/utils/inProgressTxCache.ts
@@ -9,6 +9,89 @@ const LOCAL_STORAGE_MAX = 3;
 const JSONReplacer = (_, value: any) =>
   typeof value === 'bigint' ? value.toString() : value;
 
+// Checks the existance of the given prop in a parent object
+const validateChildProp = (
+  parent: object,
+  propName: string,
+  propType: string,
+) => {
+  return (
+    propName in parent &&
+    parent[propName] !== null && // Line below is an explicit check for undefined values as well
+    typeof parent[propName] === propType
+  );
+};
+
+// Validates a single local transaction
+const validateSingleTransaction = (
+  tx: TransactionLocal,
+): tx is TransactionLocal => {
+  // Check first level required properties
+  const hasReceipt = validateChildProp(tx, 'receipt', 'object');
+  const hasRoute = validateChildProp(tx, 'route', 'string');
+  const hasTimestamp = validateChildProp(tx, 'timestamp', 'number');
+  const hasTxDetails = validateChildProp(tx, 'txDetails', 'object');
+  const hasTxHash = validateChildProp(tx, 'txHash', 'string');
+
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/utils/inProgressTxCache.ts
```diff
@@ -9,6 +9,89 @@ const LOCAL_STORAGE_MAX = 3;
...
+): tx is TransactionLocal => {
+  // Check first level required properties
+  const hasReceipt = validateChildProp(tx, 'receipt', 'object');
+  const hasRoute = validateChildProp(tx, 'route', 'string');
+  const hasTimestamp = validateChildProp(tx, 'timestamp', 'number');
+  const hasTxDetails = validateChildProp(tx, 'txDetails', 'object');
+  const hasTxHash = validateChildProp(tx, 'txHash', 'string');
+
+  if (
+    !hasReceipt ||
```
emreboga: A more performant but less reader-friendly way would be to use the function call here (e.g. `!validateChildProp(tx, 'receipt', 'object') ||`), so when one evals truty, it won't run the rest.
Let me know what you think 🙏

artursapek: you could refactor slightly and get that benefit while making it even more readable imo. change `validateChildProp` to plural `validateChildProps` (or `validatePropTypes`) and accept an object which represents the spec:

```ts
validateChildProps(tx, {
  receipt: 'object',
  route: 'string',
  timestamp: 'number',
  ...
});
```

this func would iterate through the spec, and short circuit with `return false` as soon as it finds a mismatch.

wormhole-connect/src/utils/inProgressTxCache.ts
```diff
@@ -9,6 +9,89 @@ const LOCAL_STORAGE_MAX = 3;
...
+// Checks the existance of the given prop in a parent object
+const validateChildProp = (
+  parent: object,
+  propName: string,
+  propType: string,
+) => {
+  return (
+    propName in parent &&
+    parent[propName] !== null && // Line below is an explicit check for undefined values as well
+    typeof parent[propName] === propType
```
artursapek: not positive but I think you only need this line in here? the first two are redundant with it. or am I wrong?

emreboga: We can get rid of the first line, but we need null-check along with typeof 👍

artursapek (approved):

PR #3043, opened 2024-12-04
Display timeout error message for TransactionExpiredBlockheightExceededError as well

emreboga:
This error type is also a result of a timeout when a transaction fails to be confirmed within a designated time. Therefore we can let the user to know about this and ask to retry. I added this error code to the existing timeout messaging, but please LMK if you think it's better to display another error message for this error 🙏


artursapek (approved):

tsadovska:
Tested OK on https://preview.portalbridge.com/f37b52e933/. 

image

PR #3046, opened 2024-12-05
Retrying known and unknown simulation errors

emreboga:
Adding retry for all caught errors during simulation. We will retry known errors at least 5 times and all others 3 times.

### DIFF
--- a/wormhole-connect/src/utils/wallet/solana.ts
+++ b/wormhole-connect/src/utils/wallet/solana.ts
@@ -247,9 +247,8 @@ async function createPriorityFeeInstructions(
 ) {
   let unitsUsed = 200_000;
   let simulationAttempts = 0;
-  let simulationAttemptsForKnownErrors = 0;
 
-  simulationLoop: while (simulationAttempts < 5) {
+  simulationLoop: while (true) {
     if (
       isVersionedTransaction(transaction) &&
       !transaction.message.recentBlockhash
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -44,6 +45,46 @@ import { VersionedTransaction } from '@solana/web3.js';
...
+        return true;
+      }
+    }
+  }
+
+  // In this case a require_gte expression was violated during a Swap instruction.
+  // We can retry the simulation to get a successful response.
+  if (response.logs) {
+    for (const line of response.logs) {
+      if (line.includes('RequireGteViolated')) {
```
kev1n-peters, line 74: Could we move this check into the block above here (since we're already iterating through the logs there), or do you prefer this to happen in a second pass of iterating through the logs?

emreboga: Do you think there is any possibility of having more than one of these errors in given logs? I was thinking maybe we should not bail out when see and error and aggregate all of them in the console.info?

artursapek:
at most 5 times*

wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -226,32 +268,29 @@ async function createPriorityFeeInstructions(
       : connection.simulateTransaction(transaction));
 
     if (response.value.err) {
-      if (response.value.err === 'BlockhashNotFound') {
-        console.info('Blockhash not found during simulation. Trying again.');
+      if (checkKnownSimulationError(response.value)) {
+        // Number of attempts will be at least 5 for known errors
+        if (simulationAttemptsForKnownErrors < 5) {
+          simulationAttemptsForKnownErrors++;
+          sleep(1000);
```
kev1n-peters: We need to call `await sleep()` (realize we were not doing this before)

wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -196,24 +239,23 @@ export async function signAndSendTransaction(
...
 async function createPriorityFeeInstructions(
   connection: Connection,
   transaction: Transaction | VersionedTransaction,
   commitment?: Commitment,
 ) {
   let unitsUsed = 200_000;
   let simulationAttempts = 0;
+  let simulationAttemptsForKnownErrors = 0;
 
   simulationLoop: while (simulationAttempts < 5) {
```
kev1n-peters: I don't think this check makes sense anymore, we could change it to `while (true)`. We will break out of the loop when we've exhausted known or unknown # of retries (exception gets thrown) or the simulation succeeds.

emreboga: Yup! By the way, I realized we also don't need a second counter.

kev1n-peters (approved):

kev1n-peters (approved):

PR #3049, opened 2024-12-05
Add retry logic to track() call after initiate

kev1n-peters:
Fixes an issue where track() can throw (possibly due to an RPC error), but initiate() succeeded, indicating the transaction has landed on-chain. This change adds retry logic to call track() up to 5 times in case the error is intermittent.

### DIFF
--- a/wormhole-connect/src/routes/sdkv2/route.ts
+++ b/wormhole-connect/src/routes/sdkv2/route.ts
@@ -24,6 +24,7 @@ import {
   getWrappedTokenId,
   isFrankensteinToken,
   isWrappedToken,
+  sleep,
 } from 'utils';
 import { TransferWallet } from 'utils/wallet';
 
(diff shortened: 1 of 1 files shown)


kev1n-peters:
Is this sufficient or do we need to use exponential backoff or something similar?

wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -396,10 +397,23 @@ export class SDKv2Route {
...
+          if (receipt.state >= TransferState.SourceInitiated) {
+            return [route, receipt];
+          }
+        }
+      } catch (e) {
+        console.error(
+          `Error tracking transfer (attempt ${retries + 1} / 5}):`,
+          e,
+        );
+        await sleep(1000);
```
artursapek: make this exponential to cover rate limit issues

kev1n-peters:
> Is this sufficient or do we need to use exponential backoff or something similar?

Added this

artursapek (approved):

PR #3056, opened 2024-12-10
Do not render widget when no token config found 

emreboga:
This is a very unlikely case unless a user tempers with the localStorage entry. It was discovered while testing Portal with an invalid localStorage entry. To fix this potential problem we are adding another layer of verification to filter out the entries with no token configs.

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/ReviewTransaction/GasSlider.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/ReviewTransaction/GasSlider.tsx
@@ -106,7 +106,7 @@ const GasSlider = (props: {
 
   useEffect(() => {
     dispatch(setToNativeToken(debouncedPercentage / 100));
-  }, [debouncedPercentage]);
+  }, [debouncedPercentage, dispatch]);
 
   const nativeGasPrice = useMemo(() => {
     if (!destChain || !nativeGasToken) {
--- a/wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
+++ b/wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
@@ -108,7 +108,8 @@ const WidgetItem = (props: Props) => {
     toChain,
     token: tokenTuple,
   } = txDetails || {};
-  const token = config.tokens.get(tokenTuple);
+
+  const tokenConfig = config.tokens.get(tokenTuple);
 
   // Initialize the countdown
   const { seconds, minutes, totalSeconds, isRunning, restart } = useTimer({
@@ -257,7 +258,8 @@ const WidgetItem = (props: Props) => {
     }
   }, [dispatch, receipt, route, routeContext, timestamp, txDetails]);
 
-  if (!transaction) {
+  // Do not render if we don't have the transaction or the token config
+  if (!transaction || !tokenConfig) {
     return <></>;
   }
 
@@ -286,7 +288,7 @@ const WidgetItem = (props: Props) => {
               <Stack direction="row" alignItems="center">
                 <Typography fontSize={14} marginRight="8px">
                   {`${sdkAmount.display(sdkAmount.truncate(amount, 4))} ${
-                    token?.symbol || ''
+                    tokenConfig?.symbol || ''
                   }`}
                 </Typography>
                 <Box className={classes.chainIcon}>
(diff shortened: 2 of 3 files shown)


wormhole-connect/src/views/v2/Bridge/ReviewTransaction/GasSlider.tsx
```diff
@@ -117,7 +117,6 @@ const GasSlider = (props: {
     const tokenAmount = amount.display(
       amount.truncate(props.destinationGasDrop, 6),
     );
-    console.log(tokenAmount);
```
emreboga: This one was creating some noise in the console when in Review Transaction view. Please let me know if you think it'd still be useful when debugging.

wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
```diff
@@ -108,7 +108,8 @@ const WidgetItem = (props: Props) => {
     toChain,
     token: tokenTuple,
   } = txDetails || {};
-  const token = config.tokens.get(tokenTuple);
+
+  const tokenConfig = config.tokens.get(tokenTuple);
```
artursapek: can we stop using the term "token config"? just keep this as-is.

wormhole-connect/src/views/v2/TxHistory/Widget/index.tsx
```diff
@@ -43,7 +44,25 @@ const TxHistoryWidget = () => {
 
   useEffect(() => {
     // Get all in-progress transactions from localStorage
-    setTransactions(getTxsFromLocalStorage());
+    const txs = getTxsFromLocalStorage();
+
+    // Filter out the ones with unknown token configs
```
artursapek: again, please don't call them token configs. they're just tokens.

wormhole-connect/src/views/v2/TxHistory/Widget/index.tsx
```diff
@@ -43,7 +44,25 @@ const TxHistoryWidget = () => {
...
+    const verifiedTxs = txs?.filter((tx) => {
+      if (!tx?.txDetails?.token) {
+        return false;
+      }
+      try {
+        return !!config.tokens.get(tx?.txDetails?.token);
```
artursapek, line 55: There's a better way to handle this edge case. If there's a pending tx, then I think we can safely assume the tokens it refers to exist. And if that's true, and they are not in the token cache, you should just `getOrFetchToken` them here. You can ignore the result - just kick it off and the `TokensContext` will fetch the token and put it in the cache in the background. [See this for example](https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx#L77).

This way you don't have to filter this thing out, you can render it properly once that fetch is complete.

You'll probably need to handle the "loading" case in the meantime where the token is not yet in the cache... maybe just render a `Skeleton`. And re-render once the cache is updated (which you can do by depending on `lastTokenCacheUpdate` [like here](https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/views/v2/Bridge/index.tsx#L323))

artursapek (changes requested):

wormhole-connect/src/views/v2/TxHistory/Widget/index.tsx
emreboga: Thanks for the recommendation, but it has the assumption that the token is among the supported ones. The need for this check came in after I learnt that Portal QA team has a test case to make sure Connect is not broken after user tempers with the local storage entry. One of them was the token. In that case it doesn't matter whether we fetch the token. This is still a very edge case and I'm ok leaving it out if you think it's unnecessary.

wormhole-connect/src/views/v2/TxHistory/Widget/Item.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3056/commits/e84f5c3817342ff641d038b9f3934c940db55db4

wormhole-connect/src/views/v2/TxHistory/Widget/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3056/commits/e84f5c3817342ff641d038b9f3934c940db55db4

wormhole-connect/src/views/v2/TxHistory/Widget/index.tsx
artursapek: OK. Fine with me. Unlikely the non-tamper case would ever happen to begin with anyway

artursapek (approved):

PR #3075, opened 2024-12-17
Adding rotate animation to swap action

emreboga:
Adding a bit of fun to Connect :)

https://www.loom.com/share/0df6ee68e0b543d89431f8f9352c1b8b?sid=1aeaef85-2030-4757-a88c-93c74f27d8de

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/SwapInputs/index.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/SwapInputs/index.tsx
@@ -1,5 +1,4 @@
-import React, { useCallback } from 'react';
-
+import React, { useCallback, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import IconButton from '@mui/material/IconButton';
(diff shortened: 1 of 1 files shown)


kev1n-peters (approved):

wormhole-connect/src/views/v2/Bridge/SwapInputs/index.tsx
```diff
@@ -74,10 +78,29 @@ function SwapInputs() {
...
+          },
+          '100%': {
+            transform: 'rotate(0deg)',
+          },
+        },
+      }}
       onClick={swap}
       disabled={!canSwap}
     >
-      <SwapVertIcon color="secondary" />
```
kev1n-peters, line 80: What's the reason for removing this color?

emreboga: So that the icon can pick up the theme color. Setting it directly was overriding the theme colors that we use in other places.

PR #3080, opened 2024-12-18
removed showHamburgerMenu config

kev1n-peters:


wormhole-connect/src/components/PageHeader.tsx
```diff
@@ -4,19 +4,17 @@ import { makeStyles } from 'tss-react/mui';
 import { setRoute } from 'store/router';
 
 import Header, { Alignment } from './Header';
-import MenuFull from './MenuFull';
```
emreboga, line 7: I don't think this component is used elsewhere, can you remove that too?

emreboga (approved):
LGTM!

PR #3105, opened 2024-12-21
Arbitrary destination address support

kev1n-peters:
Allows the user to set an arbitrary destination address.

Manual routes are disabled when the destination wallet is not connected.


kev1n-peters:
image

wormhole-connect/src/utils/address.ts
```diff
@@ -0,0 +1,60 @@
...
+  } catch {
+    return false;
+  }
+}
+
+function isAptosAddress(address: string): boolean {
+  return /^0x[0-9a-fA-F]{64}$/.test(address);
+}
+
+export function isValidWalletAddress(chain: Chain, address: string): boolean {
```
kev1n-peters, line 29: I suppose this function could belong in the SDK, but adding it here was easier and it's sort of specific to how we want to handle wallet input addresses in Connect. I will move it if that would be better though

PR #3113, opened 2024-12-23
Arbitrary destination address support

kev1n-peters:
Allows the user to set an arbitrary destination address.

Manual routes are disabled when the destination wallet is not connected.

### DIFF
--- a/wormhole-connect/src/store/wallet.ts
+++ b/wormhole-connect/src/store/wallet.ts
@@ -5,7 +5,7 @@ import {
   swapWalletConnections,
   TransferWallet,
 } from 'utils/wallet';
-import { ReadOnlyWallet } from 'utils/wallet/AddressOnlyWallet';
+import { ReadOnlyWallet } from 'utils/wallet/ReadOnlyWallet';
 
 export type WalletData = {
   type: Context | undefined;
--- a/wormhole-connect/src/store/wallet.ts
+++ b/wormhole-connect/src/store/wallet.ts
@@ -5,7 +5,7 @@ import {
   swapWalletConnections,
   TransferWallet,
 } from 'utils/wallet';
-import { AddressOnlyWallet } from 'utils/wallet/AddressOnlyWallet';
+import { ReadOnlyWallet } from 'utils/wallet/AddressOnlyWallet';
 
 export type WalletData = {
   type: Context | undefined;
@@ -109,9 +109,9 @@ export const walletSlice = createSlice({
 
       swapWalletConnections();
 
-      // If the new sending wallet is an AddressOnlyWallet,
+      // If the new sending wallet is a ReadOnlyWallet,
       // disconnect it since it can't be used for signing
-      if (state.sending.name === AddressOnlyWallet.NAME) {
+      if (state.sending.name === ReadOnlyWallet.NAME) {
         disconnect(TransferWallet.SENDING);
         state[TransferWallet.SENDING] = NO_WALLET;
       }
--- a/wormhole-connect/src/store/wallet.ts
+++ b/wormhole-connect/src/store/wallet.ts
@@ -5,6 +5,7 @@ import {
   swapWalletConnections,
   TransferWallet,
 } from 'utils/wallet';
+import { AddressOnlyWallet } from 'utils/wallet/AddressOnlyWallet';
 
 export type WalletData = {
   type: Context | undefined;
@@ -105,7 +106,15 @@ export const walletSlice = createSlice({
       const tmp = state.sending;
       state.sending = state.receiving;
       state.receiving = tmp;
+
       swapWalletConnections();
+
+      // If the new sending wallet is an AddressOnlyWallet,
+      // disconnect it since it can't be used for signing
+      if (state.sending.name === AddressOnlyWallet.NAME) {
+        disconnect(TransferWallet.SENDING);
+        state[TransferWallet.SENDING] = NO_WALLET;
+      }
     },
   },
 });
--- a/wormhole-connect/src/utils/address.ts
+++ b/wormhole-connect/src/utils/address.ts
@@ -1,10 +1,22 @@
-import { Chain, chainToPlatform, Wormhole } from '@wormhole-foundation/sdk';
+import {
+  Chain,
+  chainToPlatform,
+  encoding,
+  Wormhole,
+} from '@wormhole-foundation/sdk';
 import { isValidSuiAddress } from '@mysten/sui.js';
 import { PublicKey } from '@solana/web3.js';
 import { getAddress } from 'ethers';
 
 function isEvmAddress(address: string): boolean {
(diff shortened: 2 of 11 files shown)


wormhole-connect/src/utils/wallet/ReadOnlyWallet.ts
```diff
@@ -0,0 +1,95 @@
...
+  Wallet,
+} from '@xlabs-libs/wallet-aggregator-core';
+import { Chain, chainToChainId } from '@wormhole-foundation/sdk';
+
+export class ReadOnlyWallet extends Wallet {
+  private _isConnected: boolean = true;
+
+  static readonly NAME = 'ReadyOnlyWallet';
+
+  constructor(readonly _address: string, readonly _chain: Chain) {
```
kev1n-peters: I considered passing in a `ChainAddress` object, but to get the actual native address string (for other methods in this class that return it) we have to do something like `chainAddress.toNative(chainAddress.chain).toString()` which I feel is unnecessary work. Wdyt?

wormhole-connect/src/utils/address.ts
```diff
@@ -0,0 +1,79 @@
...
+
+function isValidAptosAddress(address: string): boolean {
+  return (
+    address.startsWith('0x') &&
+    address.length === 66 &&
+    encoding.hex.valid(address)
+  );
+}
+
+export function isValidWalletAddress(chain: Chain, address: string): boolean {
```
artursapek: I think you could accomplish this by simply calling [`toNative(chain, address)`](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/core/definitions/src/address.ts#L86) in a `try`/`catch`. if the parsing succeeds, it's a valid address.

wormhole-connect/src/utils/wallet/ReadOnlyWallet.ts
artursapek: `ChainAddress` is a type that could be a `NativeAddress` or `UniversalAddress` (which are used for VAAs). that's why you need that `toNative` call. since we only need native addresses for this, I would just store it as a `NativeAddress`.

wormhole-connect/src/utils/wallet/index.ts
```diff
@@ -118,6 +124,12 @@ export const connectWallet = async (
   });
 
   localStorage.setItem(`wormhole-connect:wallet:${context}`, name);
+
+  // if the wallet is a ReadOnlyWallet, store the address in localStorage
+  // for automatic connection on next visit
+  if (wallet.getName() === ReadOnlyWallet.NAME) {
+    localStorage.setItem(`wormhole-connect:wallet:${context}:address`, address);
```
artursapek: this setting is shared for both sending and receiving wallet, and it's meant to represent which wallet the user uses for signing transactions. a sending wallet can never be a ReadOnlyWallet. if I am sending from EVM to EVM and use MetaMask for the source chain, and set a custom address on the destination chain, this will overwrite the fact that I like to use Metamask for EVM transactions. so next time I go to make an EVM transaction, it won't auto-connect to anything on the source chain. that's not good behavior.

I think we should store the custom wallet addresses separately and never store `ReadOnlyWallet` as the last used wallet, or automatically fill that in for them. if anything, we should just store all of the custom addresses I've used in chrono order and display them as suggestions in the UI under the text box.

wormhole-connect/src/utils/address.ts
kev1n-peters: This is what I was doing in an earlier iteration, but `toNative(chain, address)` is very permissive and calls succeed even when passed hex-encoded Solana addresses, ICAP Ethereum addresses, Sui addresses with a single character (e.g. 0x1), universal addresses, etc. I wanted to have stricter checks in Connect to limit the set of valid addresses in case the user makes a mistake. Am I being too paranoid? I also think it's kind of odd behavior if a user were to paste in an EVM address like `"0123"`, `toNative()` left pads it with `0`s, and then the address displayed in the input box is updated to the valid address. What do you think? We can call `toNative`, but I kind of want to have these extra checks performed before it to limit the set of valid addresses

wormhole-connect/src/utils/wallet/index.ts
kev1n-peters: This is a new setting that's only set when the wallet is a `ReadOnlyWallet` which can only be a receiving wallet, but agree that it should probably be general and also apply to sending wallets. I'm going to remove this for now and store custom addresses / suggest them in the UI under the text box in a follow-up PR if that's okay.

wormhole-connect/src/utils/address.ts
artursapek: ah makes sense. thanks

wormhole-connect/src/utils/wallet/index.ts
artursapek: I know. but you're also letting it store this wallet as the preferred wallet in the surrounding code

artursapek:
the earlier thread is resolved now, but I was asking if we could avoid setting the "preferred wallet" in localStorage to `ReadOnlyWallet`. it's line 125 in `src/utils/wallet/index.ts`. that setting is only for actual wallets which are signers.

wormhole-connect/src/store/wallet.ts
```diff
@@ -105,7 +106,15 @@ export const walletSlice = createSlice({
...
       state.sending = state.receiving;
       state.receiving = tmp;
+
       swapWalletConnections();
+
+      // If the new sending wallet is a ReadOnlyWallet,
+      // disconnect it since it can't be used for signing
+      if (state.sending.name === ReadOnlyWallet.NAME) {
+        disconnect(TransferWallet.SENDING);
+        state[TransferWallet.SENDING] = NO_WALLET;
```
artursapek, line 116: I think instead of doing this, we could see if they have a previously used signer wallet and auto-connect to it.

kev1n-peters: We can't dispatch inside of a reducer (the call to `connectLastUsedWallet` calls `dispatch`, so we would need to find somewhere else to do this

kev1n-peters: Was thinking about doing it in here https://github.com/wormhole-foundation/wormhole-connect/pull/3113/files#diff-05193c8cb80265695f1f1d6ecf386de92edf2c88ea2c3daf1104b3e2ccef78dbR109 (when the ReadOnlyWallet is disconnected), but we don't have access to the new source chain since the closure captures the previous (pre swap chains) value

artursapek: hmmm if it's complicated then nvm, not worth it

kev1n-peters: I think this is a good idea, but maybe can revisit when we add the list of previously used addresses

artursapek (approved):
ship it hombre

PR #3164, opened 2025-01-08
Remove "Review transaction" view

emreboga:
This PR will remove the "Review Transaction" view/step and allow user to confirm a transaction from the main view. One important change here is moving the gas slider inside the route card for those that has gas top-up enabled.


wormhole-connect/src/hooks/useGasSlider.ts
```diff
@@ -7,7 +7,6 @@ type Props = {
   destChain: Chain | undefined;
   destToken: string;
   route?: string;
-  valid: boolean;
```
emreboga, line 10: If gas top-up is not available for the route, we don't show it in the first place; hence this property is no longer needed.

wormhole-connect/src/hooks/useConfirmTransaction.ts
emreboga, line 1: Please note that git thinks this was a move from `views/v2/Bridge/ReviewTransaction/index.ts` to `hooks/useConfirmTransaction.ts` because the most of the lines in the first one was moved over to the second one. However we actually just moved the send function to new hook and deleted `views/v2/Bridge/ReviewTransaction/index.ts`.

artursapek (approved):

PR #3180, opened 2025-01-14
Reset manual claim error during a new claim action

emreboga:
This happens when manual claim errors and user clicks to claim one more time.


artursapek (approved):
based

PR #3182, opened 2025-01-15
custom wallet addy UI tweaks

kev1n-peters:
image


kev1n-peters:
Place button to the right of the address input instead of under it

artursapek (approved):

PR #3200, opened 2025-01-16
Token refactor follow-up: config updates

artursapek:
This makes some breaking changes to the config, so that it works with the new arbitrary tokens system.

### Token whitelist

The token whitelist used to reference tokens by their "keys" like this:

```js
{
  tokens: ["USDCeth", "USDCsol"]
}
```

Since tokens no longer have these custom keys, we now support two other means of whitelisting tokens:

1. Integrators can easily whitelist a single token across all chains by providing its symbol
2. Integrators can also be more granular by providing token tuples ([chain, address]).

```js
{
  tokens: ["USDC", ["Solana", "2qEHjDLDLbuBgRYvsxhc5D6uDWAivNFZGan56P1tpump"]]
}
```

This config whitelists USDC coin on all chains, and Pnut the Squirrel on Solana. There is some nuance to how this works because of wrapped tokens, which is [explained here](https://github.com/wormhole-foundation/wormhole-connect/pull/3200/files#diff-5af62c97f00fccc78c30b47c00a74b134d098b4a4d840bb5d8caacf857c8d9cfR186-R197).

image

### Default tokens

The `DefaultInputs` config now has to be either the symbol or the address for the token.

```js
{
  ui: {
    defaultInputs: {
      fromChain: 'Solana',
      fromToken: '2qEHjDLDLbuBgRYvsxhc5D6uDWAivNFZGan56P1tpump',
      toChain: 'Ethereum',
      toToken: 'USDC',
    }
  }
}
```

## New properties

This also adds a new config option which disables the ability to paste in token addresses:

```js
{
  ui: {
    disableUserInputtedTokens: true
  }
}
```

This is set to true by default if an integrator provides the `tokens` whitelist.

### DIFF
--- a/wormhole-connect/src/config/index.ts
+++ b/wormhole-connect/src/config/index.ts
@@ -81,10 +81,10 @@ export function buildConfig(
 
   const ui = createUiConfig(customConfig.ui ?? {});
 
-  if (customConfig.tokens && ui.disableArbitraryTokens === undefined) {
+  if (customConfig.tokens && ui.disableUserInputtedTokens === undefined) {
     // If the integrator has provided a whitelist of tokens, we can reasonably assume they also don't want
     // users pasting in arbitrary token addresses.
-    ui.disableArbitraryTokens = true;
+    ui.disableUserInputtedTokens = true;
   }
 
   return {
--- a/wormhole-connect/src/config/index.ts
+++ b/wormhole-connect/src/config/index.ts
@@ -79,6 +79,14 @@ export function buildConfig(
     validateDefaults(customConfig.ui.defaultInputs, networkData.chains, tokens);
   }
 
+  const ui = createUiConfig(customConfig.ui ?? {});
+
+  if (customConfig.tokens) {
+    // If the integrator has provided a whitelist of tokens, we can reasonably assume they also don't want
+    // users pasting in arbitrary token addresses.
+    ui.disableArbitraryTokens = true;
+  }
+
   return {
     whLegacy,
     sdkConfig,
@@ -138,14 +146,7 @@ export function buildConfig(
         return 0;
       }),
     tokens,
-
-    /*
-    // For token bridge =^_^=
-    wrappedTokenAddressCache: new WrappedTokenAddressCache(
-      tokens,
-      wrappedTokens,
-    ),
-    */
+    tokenWhitelist: customConfig.tokens,
 
     routes: new RouteOperator(customConfig.routes),
 
--- a/wormhole-connect/src/config/index.ts
+++ b/wormhole-connect/src/config/index.ts
@@ -81,7 +81,7 @@ export function buildConfig(
 
   const ui = createUiConfig(customConfig.ui ?? {});
 
-  if (customConfig.tokens) {
+  if (customConfig.tokens && ui.disableArbitraryTokens === undefined) {
     // If the integrator has provided a whitelist of tokens, we can reasonably assume they also don't want
     // users pasting in arbitrary token addresses.
     ui.disableArbitraryTokens = true;
--- a/wormhole-connect/src/config/index.ts
+++ b/wormhole-connect/src/config/index.ts
@@ -81,7 +81,11 @@ export function buildConfig(
 
   const ui = createUiConfig(customConfig.ui ?? {});
 
-  if (customConfig.tokens && ui.disableUserInputtedTokens === undefined) {
+  if (
+    customConfig.tokens &&
+    customConfig.tokens.length > 0 &&
+    ui.disableUserInputtedTokens === undefined
+  ) {
     // If the integrator has provided a whitelist of tokens, we can reasonably assume they also don't want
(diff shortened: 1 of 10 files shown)


wormhole-connect/src/config/index.ts
```diff
@@ -79,6 +79,14 @@ export function buildConfig(
     validateDefaults(customConfig.ui.defaultInputs, networkData.chains, tokens);
   }
 
+  const ui = createUiConfig(customConfig.ui ?? {});
+
+  if (customConfig.tokens) {
+    // If the integrator has provided a whitelist of tokens, we can reasonably assume they also don't want
+    // users pasting in arbitrary token addresses.
+    ui.disableArbitraryTokens = true;
```
emreboga: If the integrator explicitly specifies `disableArbitraryTokens = false` while having whitelist tokens in the config, should that still enable arbitrary tokens?

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -179,6 +182,69 @@ const TokenList = (props: Props) => {
...
+          desiredSymbols.push(item);
+        } else if (isTokenTuple(item)) {
+          const tokenId = tokenIdFromTuple(item);
+          if (tokenId.chain === props.selectedChainConfig.sdkName) {
+            filteredTokens.add(tokenId.address.toString());
+          }
+        }
+      }
+
+      for (const symbol of desiredSymbols) {
```
emreboga, line 214: Above it says "If there is a single native token with that symbol, we simply use that", bu I can't find where that's implemented in this loop.

wormhole-connect/src/config/index.ts
artursapek: good question... I can add support for that

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
artursapek: the `foundNative` flag is how that's implemented. although you're right the comment is slightly inaccurate: we don't check that there's only one native (not wrapped) token with the symbol; we simply take the first one we find. if there were two native tokens with the same symbol, it would choose one of them arbitrarily. this isn't likely to happen, but it's possible.

I suppose I could remove the `break` on line 223 to include _all_ native tokens with a matching symbol?

wormhole-connect/src/config/index.ts
artursapek: 62b973c42da1310123b78f01dea674065a258953

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
emreboga: Yes, I think that'll work.

wormhole-connect/src/config/tokens.ts
```diff
@@ -378,10 +405,9 @@ export class TokenCache extends TokenMapping<Token> {
 
         mapping.setLocalStorageKey(localStorageKey);
 
-        for (const [key, tokenData] of Object.entries(asJson.tokens)) {
-          const tokenId = parseTokenKey(key);
+        for (const [_key, tokenData] of Object.entries(asJson.tokens)) {
```
kev1n-peters: nit: can do `[, tokenData]` to get rid of the unused `_key` warning

wormhole-connect/src/config/index.ts
kev1n-peters: Should we change to `customConfig.tokens?.length > 0` instead? passing `[]` is truthy

artursapek: hah yeah good call. I actually did that [here](https://github.com/wormhole-foundation/wormhole-connect/pull/3200/files#diff-5af62c97f00fccc78c30b47c00a74b134d098b4a4d840bb5d8caacf857c8d9cfR185) but failed to in this other spot

artursapek: https://github.com/wormhole-foundation/wormhole-connect/pull/3200/commits/5b7623f5a9a239ae6540bea51a25a28f34c1ec05

wormhole-connect/src/config/tokens.ts
artursapek: https://github.com/wormhole-foundation/wormhole-connect/pull/3200/commits/3ccc38750c4f31187953be6d266e21e70867c157

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
artursapek: https://github.com/wormhole-foundation/wormhole-connect/pull/3200/commits/3ccc38750c4f31187953be6d266e21e70867c157

kev1n-peters (approved):

PR #3208, opened 2025-01-20
Show USD value in amount input

emreboga:
This PR will add the USD value underneath the token amount in the amount input.

https://www.loom.com/share/5a1078713e904105afc59eae33d6acbf?sid=91aaee09-8e89-4cef-b4e8-17ff1eeab0aa

### DIFF
--- a/wormhole-connect/src/utils/coingecko.ts
+++ b/wormhole-connect/src/utils/coingecko.ts
@@ -2,8 +2,8 @@ import { Chain, isNative, TokenId, Wormhole } from '@wormhole-foundation/sdk';
 import config from 'config';
 import { TokenMapping } from 'config/tokens';
 
-const COINGECKO_URL = 'https://api.coingecko.com/';
-const COINGECKO_URL_PRO = 'https://pro-api.coingecko.com/';
+const COINGECKO_URL = 'https://api.coingecko.com';
+const COINGECKO_URL_PRO = 'https://pro-api.coingecko.com';
 
 const NATIVE_TOKEN_IDS: Partial<Record<Chain, string>> = {
   Solana: 'solana',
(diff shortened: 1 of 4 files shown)


wormhole-connect/src/utils/coingecko.ts
```diff
@@ -2,8 +2,8 @@ import { Chain, isNative, TokenId, Wormhole } from '@wormhole-foundation/sdk';
 import config from 'config';
 import { TokenMapping } from 'config/tokens';
 
-const COINGECKO_URL = 'https://api.coingecko.com/';
```
emreboga, line 5: Cosmetics: Removing these as they are causing double trailing dashes in the requests.

wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
```diff
@@ -183,13 +204,66 @@ const AmountInput = (props: Props) => {
...
 
   const handleChange = useCallback((newValue: string): void => {
-    dispatch(setAmount(newValue));
     setAmountInput(newValue);
   }, []);
 
+  const tokenPriceAdornment = useMemo(() => {
+    const price = calculateUSDPrice(
+      getTokenPrice,
+      Number(amountInput),
```
artursapek: we have special handling for when input starts with `.`, so users can add a fractional amount quickly. this code doesn't use that handling and throws NaN into the UI. see `onInnerChange`.

image

wormhole-connect/src/views/v2/Bridge/AmountInput/index.tsx
```diff
@@ -199,7 +273,7 @@ const AmountInput = (props: Props) => {
         disabled={maxButtonDisabled}
         onClick={() => {
           if (props.tokenBalance) {
-            handleChange(sdkAmount.display(props.tokenBalance));
+            handleDebouncedChange(sdkAmount.display(props.tokenBalance));
```
artursapek: max button doesnt seem to work :(

https://github.com/user-attachments/assets/f2322fc5-c6fc-432b-8128-6ea15f14af9b

artursapek (approved):

PR #3209, opened 2025-01-20
Transaction config

kev1n-peters:


wormhole-connect/src/utils/wallet/solana.ts
```diff
@@ -321,13 +322,60 @@ async function createPriorityFeeInstructions(
...
+              connection,
+              transaction,
+              percentile,
+              percentileMultiple,
+              min,
+              max,
+            );
+
+      const feeInSol = (fee / 1e6 / 1e9) * unitsUsed * 1.2;
+      console.log(
```
kev1n-peters: TODO: remove this log

wormhole-connect/src/config/types.ts
```diff
@@ -312,3 +319,11 @@ export interface TransactionLocal {
   txDetails: TransferInfo;
   isReadyToClaim?: boolean;
 }
+
+export interface TransactionSettings {
+  Solana?: {
+    priorityFee?: PriorityFeeOptions & {
```
kev1n-peters: `PriorityFeeOptions` comes from the SDK. I'm not sure if we want to add `feeEstimator` to that type or extend it like we do here.

artursapek: why do we need this? can't we just automatically use triton's API if we detect that the configured Solana RPC is a triton URL?

emreboga (approved):
LGTM!

PR #3212, opened 2025-01-21
WEB-6: useTrackTransfer optimization: sleep until 1 minute before transfer ETA

artursapek:
This makes `useTrackTransfer` wait until 1 minute before the transfer is estimated to complete, before trying to fetch updates using `track`.

I also removed `stateChanged` since it's always identical to `isCompleted`. ([It wasn't originally](https://github.com/wormhole-foundation/wormhole-connect/commit/1fe5b35ff08ada3629942e5cd6b15b8da69a1e3e#diff-9adedeb8d27e2b358795bdf604404e6ece65d46fe961681eca1d9520a616e1edR52), but after some code changes it's now redundant)

The `sleepTime` variable is not actually that important - the retries are generally handled downstream in the SDK code. Here we only hit the `sleepTime` code if `track` throws an error.

### DIFF
--- a/wormhole-connect/src/hooks/useTrackTransfer.ts
+++ b/wormhole-connect/src/hooks/useTrackTransfer.ts
@@ -60,12 +60,11 @@ const useTrackTransfer = (props: Props): ReturnProps => {
 
       while (isActive && !isCompleted(receipt)) {
         if (eta !== undefined) {
-          // If we have an ETA, and it's longer than 2 minutes out, we wait until 2 minutes are left
+          // If we have an ETA, and it's longer than 1 minute out, we wait until 1 minute is left
           // before trying to track the transfer's progress.
           const msRemaining = millisUntilEta(eta);
 
           if (msRemaining > MINIMUM_ETA) {
-            // Sleep until 1 minute from ETA
             await sleep(msRemaining - MINIMUM_ETA);
           }
         }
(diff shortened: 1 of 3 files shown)


wormhole-connect/src/hooks/useTrackTransfer.ts
```diff
@@ -57,12 +52,24 @@ const useTrackTransfer = (props: Props): ReturnProps => {
...
+      const millisUntilEta = (eta: Date) =>
+        eta.valueOf() - new Date().valueOf();
+
       const wh = await getWormholeContextV2();
       const sdkRoute = new route.rc(wh);
 
-      let stateChanged = false;
+      while (isActive && !isCompleted(receipt)) {
+        if (eta !== undefined) {
+          // If we have an ETA, and it's longer than 2 minutes out, we wait until 2 minutes are left
```
emreboga: 2 minutes or 1 minute?

wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -190,10 +190,36 @@ const Redeem = () => {
 
   const getUSDAmount = useUSDamountGetter();
 
+  const etaDate: Date | undefined = useMemo(() => {
+    if (eta && txTimestamp) {
+      return new Date(txTimestamp + eta);
+    } else {
+      return undefined;
+    }
+  }, [eta]);
```
emreboga: Do you need `txTimestamp` in deps array?

wormhole-connect/src/hooks/useTrackTransfer.ts
artursapek: oops hehe

emreboga (approved):
LGTM!

PR #3215, opened 2025-01-22
memoize getOrFetchToken

kev1n-peters:
We need to memoize `getOrFetchToken` with `useCallback` to prevent hooks from firing when the function is re-created on each state update in `TokensContext`


artursapek (approved):
ty

PR #3239, opened 2025-01-29
Hyperliquid route

emreboga:
This PR adds Hyperliquid route, Hyperliquid chain and enables the Hyperliquid deposit UX flow. Please see the loom below for details.

**PR walkthrough:**
https://www.loom.com/share/47b64c7586c84fc5b4914d972f260e88?sid=46f72b22-0ab9-484f-9d89-fb8149de756a

**UX demo:**
https://www.loom.com/share/14b3c1b7ba4547d0be0dbc84b4b973ad?sid=c470625d-3be4-4e1c-82e1-dbe7111487c9


wormhole-connect/src/hooks/useComputeDestinationTokens.ts
```diff
@@ -76,7 +78,15 @@ const useComputeDestinationTokens = (props: Props): ReturnProps => {
         return;
       }
 
-      setSupportedDestTokens(supported);
+      setSupportedDestTokens(
```
artursapek: image

can we please not inline big illegible ternaries like this? just add a new code block for this instead, it would be way easier to follow what is going on. this code makes my eyes hurt

```ts
// Comment about how we shoe horned this incompatible feature in for business reasons
if (props.toNonSDKChain === 'Hyperliquid') {
  supported = supported.filter((t) => t.tuple[0] === 'Arbitrum' && t.tuple[1] === '0xaf88d065e77c8cC2239327C5EDb3A432268e5831')
}

setSupportedDestTokens(supported)
```

also, you could  `import { circle } from '@wormhole-foundation/sdk'` and simplify further:


```ts
if (props.toNonSDKChain === 'Hyperliquid') {
  supported = [
    config.tokens.getBySymbol(
      'Arbitrum',
      circle.usdcContract.get('Mainnet', 'Arbitrum')
    )
  ];
}

setSupportedDestTokens(supported)
```

wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
```diff
@@ -90,7 +90,15 @@ const useFetchSupportedRoutes = (): HookReturn => {
 
       if (isActive) {
         setIsFetching(false);
-        setRoutes(_routes);
+        setRoutes(
+          _routes.filter((r) =>
```
artursapek: same feedback as above. give this some room to breathe. `setRoutes` should be a simple one-line call

```ts
_routes = _routes.filter(
  (r) => toNonSDKChain === 'Hyperliquid' ?
    r === 'HyperliquidRoute' : 
    r !== 'HyperliquidRoute'
)

setRoutes(_routes);
```

do we actually have to filter _out_ the `HyperliquidRoute` though?

wormhole-connect/src/routes/hyperliquid/index.ts
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  // Get the list of source tokens that are possible to receive
+  static async supportedDestinationTokens<N extends Network>(
+    _sourceToken: TokenId,
+    _fromChain: ChainContext<N>,
+    _toChain: ChainContext<N>,
+  ): Promise<TokenId[]> {
+    // HL route supports only USDCarbitrum as destination token
+    return [Wormhole.tokenId('Arbitrum', USDC_ARBITRUM)];
```
artursapek:
```ts
import { circle } from '@wormhole-foundation/sdk';

// ...

return [Wormhole.tokenId('Arbitrum', circle.usdcContract.get('Mainnet', 'Arbitrum'))];
```

wormhole-connect/src/routes/hyperliquid/index.ts
```diff
@@ -0,0 +1,266 @@
...
+      signer,
+      quote,
+      to,
+    )) as routes.Receipt;
+  }
+
+  async complete(signer: Signer, receipt: R): Promise<R> {
+    if (!isAttested(receipt)) {
+      throw new Error(
+        'The source must be finalized in order to complete the transfer',
```
artursapek: this error message isn't accurate. "Source finalized" is a distinct step, different from "Attested". [see here](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/connect/src/types.ts#L16)

looks like you [copied this from here](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/connect/src/routes/tokenBridge/manual.ts#L152-L153)? that code should also be fixed :P

wormhole-connect/src/routes/hyperliquid/index.ts
```diff
@@ -0,0 +1,266 @@
...
+  // there is no way we can tell whether the manual claim is already completed.
+  async resume(_txid: TransactionId): Promise<R> {
+    throw new Error('Hyperliquid route transactions cannot be resumed.');
+  }
+
+  override transferUrl(txid: string): string {
+    return this.mayanRoute.transferUrl(txid);
+  }
+
+  public override async *track(receipt: R, timeout?: number) {
```
artursapek, line 271: I'm pretty sure this is right, and it's probably the weirdest part of this whole route. nice job

wormhole-connect/src/routes/hyperliquid/index.ts
```diff
@@ -0,0 +1,266 @@
...
+    try {
+      const arb = this.wh.getChain('Arbitrum');
+      const rpc = await arb.getRpc();
+      const txStatus = receipt['txstatus'] as TransactionStatus;
+
+      const usdcToken = EvmPlatform.getTokenImplementation(rpc, USDC_ARBITRUM);
+
+      const txReq = await usdcToken.transfer.populateTransaction(
+        HL_BRIDGE_ARB,
+        BigInt(txStatus.minAmountOut64),
```
artursapek: is there a way to find out what the actual amount out was, so we're not leaving dust? with larger transfers, this will be noticeable.

wormhole-connect/src/utils/errors.ts
```diff
@@ -27,9 +27,12 @@ export const AMOUNT_IN_TOO_SMALL = new RegExp('AmountInTooSmall', 'm');
 export function interpretTransferError(
   e: any,
   transferDetails: TransferDetails,
+  isHyperliquidTransfer = false,
```
artursapek: ??? don't add parameters to generic utils that are specific to a single route. `TransferDetails` interface actually has a `route` property inside it. why can't you use that?

wormhole-connect/src/views/v2/Bridge/AssetPicker/ChainList.tsx
```diff
@@ -109,30 +113,45 @@ const ChainList = (props: Props) => {
...
+              <ListItemButton
+                selected={
+                  selectedNonSDKChain
+                    ? selectedChainConfig?.symbol === chain.symbol
+                    : selectedChainConfig?.key === chain.key
+                }
+                className={classes.chainButton}
+                onClick={() => {
+                  if (config.nonSDKChains?.[chain.displayName]) {
+                    onChainSelect(chain.displayName);
```
artursapek: why can't we just use a "key" for both cases and not complicate this code?

wormhole-connect/src/hooks/useComputeDestinationTokens.ts
emreboga: I will remove the ternaries, but will continue using the filter as this way we'll treat `config.routes.allSupportedDestTokens` as the source of truth. For any reason, if that won't have USDC arb in the list, we won't add it forcefully.

wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
emreboga: Yes, because without the HL chain selected as destination, the route has no meaning :)

wormhole-connect/src/routes/hyperliquid/index.ts
emreboga: By the way, I can use `circle.usdcContract.get('Mainnet', 'Arbitrum')` but since it can return undefined, we still need to fall back to the static string address (USDC_ARBITRUM).

wormhole-connect/src/routes/hyperliquid/index.ts
emreboga: I wanted to use `minAmountOut64` to be safe of the slippage. The dust left will stay in Arb in user's wallet anyways, but if user had no USDC in his Arb previously, second step to deposit into HL will fail (insufficient funds) in case of a slippage. WDYT?

wormhole-connect/src/utils/errors.ts
emreboga: Oh I must have missed that! I will do the check inside interpretTransferError.

wormhole-connect/src/views/v2/Bridge/AssetPicker/ChainList.tsx
emreboga: HL key value will be the same as Arb. From `ChainConfig` interface displayName seems like the only option to tell the difference between HL and Arb chains in the chain picker.

wormhole-connect/src/hooks/useComputeDestinationTokens.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/7a7fddb51903da78be736f52e36b26c9d9ded17f

wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/7a7fddb51903da78be736f52e36b26c9d9ded17f

wormhole-connect/src/routes/hyperliquid/index.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/7a7fddb51903da78be736f52e36b26c9d9ded17f

wormhole-connect/src/routes/hyperliquid/index.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/7a7fddb51903da78be736f52e36b26c9d9ded17f

wormhole-connect/src/utils/errors.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/7a7fddb51903da78be736f52e36b26c9d9ded17f

wormhole-connect/src/hooks/useTrackTransfer.ts
```diff
@@ -55,14 +59,25 @@ const useTrackTransfer = (props: Props): ReturnProps => {
       const millisUntilEta = (eta: Date) =>
         eta.valueOf() - new Date().valueOf();
 
-      const wh = await getWormholeContextV2();
-      const sdkRoute = new route.rc(wh);
+      let sdkRoute: routes.Route<
+        'Mainnet' | 'Testnet' | 'Devnet',
```
kev1n-peters: This can just be `Network`

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/7113b590e772e30e343132248b79c86b687c73ec

wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
artursapek: but why does `route.isRouteSupported` not handle this for you? why would HyperliquidRoute be present in `_routes`, if the destination chain is not HL?

wormhole-connect/src/routes/hyperliquid/index.ts
artursapek: no. I don't want to hard-code token addresses in Connect. that's the whole point of having those in the SDK. it will never return undefined, because it's hard-coded in the SDK: https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/core/base/src/constants/circle.ts#L18

emreboga: I see your point. The generated getter for that map (constMap) has `undefined` in the return types, but I'll ignore that with `!` operator.

wormhole-connect/src/routes/hyperliquid/index.ts
artursapek: I don't follow. `minAmountOut64` is just what Mayan guarantees you'll get at minimum, or you get a refund on the source chain. almost always, you receive more than that. 

> The dust left will stay in Arb in user's wallet anyways

what does this mean? if I bridge 1 ETH over to arbitrum, there's a $40 difference between the expected and minimum amount I receive. that's not even dust, that's forty bucks. approximately 1.5% of my total deposit.

image

we should figure out a way to check how much we _actually did receive_ after the Mayan transfer is done. it's usually going to be closer to the *expected* amount.  and we should send the full amount over to HL. otherwise you're leaving off ~1-2% of each person's deposit on Arbitrum, which they probably won't want (or even be aware of). it's messy.

wormhole-connect/src/views/v2/Bridge/AssetPicker/ChainList.tsx
artursapek: why does HL's key value have to be "Arbitrum"?

wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
emreboga: Nope, because the underlying chain key is actually Arbitrum when it calls this:
```
supported = await route.isRouteSupported(
            sourceToken,
            destToken,
            fromChain,
            toChain,
          );
```

I looked into wiring that with nonSDKChain type, but all hell broke loose :)

wormhole-connect/src/routes/hyperliquid/index.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/fb7eedd8924e6c453553a88c54ea7ed4951ccfce

wormhole-connect/src/routes/hyperliquid/index.ts
emreboga: Yes, that can be a bit too much with higher transfers. I believe we can keep track of the latest quote in HL route and use `quote.destinationToken.amount` for received amount. That's the same value we show the user in redeem view.

wormhole-connect/src/views/v2/Bridge/AssetPicker/ChainList.tsx
emreboga: Because that's of type `Chain`. Modifying it will require changing other types, starting with `ChainConfig`. And that will impact everywhere that type is used.

wormhole-connect/src/routes/hyperliquid/index.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/524ea2736137d8b1a1dc44152af2e94f0c63bf5e

emreboga: @artursapek I think a better approach than the above one is to use the toAmount in the final receipt which from the completion of Mayan transaction:
https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/bb8365966739015eddde2670c3a3fd10ad7f4458

artursapek: yeah that seems to be what you want. nice

wormhole-connect/package.json
```diff
@@ -39,8 +40,8 @@
     "@solana/spl-token": "^0.3.9",
     "@solana/wallet-adapter-wallets": "^0.19.25",
     "@solana/web3.js": "^1.95.8",
-    "@wormhole-foundation/sdk": "1.6.1-beta.3",
-    "@wormhole-foundation/sdk-definitions": "1.6.1-beta.3",
+    "@wormhole-foundation/sdk": "1.7.0",
```
kev1n-peters: We should bump this to 1.8.0 which contains Unichain CCTP (1.7.0 did not include it)

wormhole-connect/src/hooks/useGetTokenBalances.ts
```diff
@@ -1,16 +1,24 @@
...
+
+import type { WalletData } from 'store/wallet';
+import type { Token } from 'config/tokens';
+import type { ChainConfig } from 'config/types';
+import type { RootState } from 'store';
 
 const useGetTokenBalances = (
   wallet: WalletData | undefined,
-  chain: Chain | undefined,
+  chainConfig: ChainConfig | undefined,
```
artursapek, line 21: seems like you could undo this change now that you don't need anything else from `ChainConfig` anymore. I missed this in my PR

wormhole-connect/src/hooks/useTrackTransfer.ts
```diff
@@ -4,18 +4,24 @@ import { isCompleted, routes, TransferState } from '@wormhole-foundation/sdk';
...
 
 const TRACK_TIMEOUT = 120 * 1000;
 
 type Props = {
-  route: string | undefined;
+  routeName: string | undefined;
   receipt: routes.Receipt<AttestationReceipt> | null;
   // Timestamp the transfer was estimated to be finished
   eta?: Date;
+  sdkRoute?: routes.Route<
```
artursapek, line 19: I don't see why we need to add this here. what am I missing? what was wrong with always just calling `new route.rc(wh)` below?

wormhole-connect/src/utils/errors.ts
```diff
@@ -29,7 +29,10 @@ export function interpretTransferError(
   transferDetails: TransferDetails,
 ): [string, TransferError] {
   // Fall-back values
-  let uiErrorMessage = 'Error with transfer, please try again';
+  let uiErrorMessage =
+    transferDetails.route === 'HyperliquidRoute'
+      ? 'Hyperliquid deposit error. USDC has been deposited in your wallet on Arbitrum'
```
artursapek: how do you know in this case that the error happened with the second transfer? couldn't it also be the first hop, via Mayan, that failed? in that case you wouldn't have USDC on Arbitrum. this is just the generic default error message.

wormhole-connect/src/views/v2/Bridge/AssetPicker/ChainList.tsx
```diff
@@ -66,12 +64,12 @@ const useStyles = makeStyles()((theme) => ({
...
 
 type Props = {
-  chainList?: ChainConfig[];
+  chainList?: Array<ChainConfig>;
   selectedChainConfig?: ChainConfig;
   showSearch: boolean;
   setShowSearch: (value: boolean) => void;
   wallet: WalletData;
-  onChainSelect: (chain: Chain) => void;
+  onChainSelect: (chain: string) => void;
```
artursapek: could revert this

wormhole-connect/src/config/routes.ts
```diff
@@ -135,4 +135,11 @@ export const RoutesConfig: Record<string, RouteData> = {
     icon: XLabsIcon,
     pendingMessage: 'Waiting for Wormhole network consensus . . .',
   },
+  HyperliquidRoute: {
+    name: 'HyperliquidRoute',
+    displayName: 'Hyperliquid Route',
```
kev1n-peters, line 140: Should we add the `providedBy` for it to be displayed here: https://github.com/wormholelabs-xyz/wormhole-connect/blob/eff98085fdb5f24cfc5c63c1c9e9671a4571cecf/wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx#L456?

emreboga: We haven't been showing "Provided by" for routes built by Wormhole, like NTT or Token Bridge.

wormhole-connect/src/utils/errors.ts
emreboga: Right, this is all wrong. I'll add a Hyperliquid check in this block https://github.com/wormhole-foundation/wormhole-connect/blob/7b534426c0d7af3f409f02423f1f7ae01df3570a/wormhole-connect/src/views/v2/Redeem/index.tsx#L773 and create a new Error with name HyperliquidError so I can check that in interpretTransferError and show the right message.

wormhole-connect/src/hooks/useTrackTransfer.ts
emreboga: We need to keep track of the route object between main and redeem views (user can go back and fort via the in-progress widget). The reason is in the Hyperliquid route object we keep track of the finalReceipt to tell when it's deposited. We call userTrackTransfer in both of these views, and it'll create multiple instances.
Another thing I could do was to have the route return a singleton, but wasn't sure about that.

wormhole-connect/src/hooks/useGetTokenBalances.ts
emreboga: Previously this was Chain but inside the hook it was getting the config from that. Below you will see that it checks `chainConfig.context !== wallet.type`. We don't need any of those?

wormhole-connect/src/views/v2/Bridge/AssetPicker/ChainList.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/4166810e0c89d5487b626431c337cd6568b5e5e2

wormhole-connect/src/utils/errors.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/4166810e0c89d5487b626431c337cd6568b5e5e2

wormhole-connect/package.json
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3239/commits/4166810e0c89d5487b626431c337cd6568b5e5e2

artursapek (approved):

emreboga:
Closing this PR as Hyperliquid has shipped their L1 chain and we no longer need to do the alternative implementation.

PR #3240, opened 2025-01-30
Update Mayan protocol limit for MCTP and SWIFT

emreboga:
We are removing the limit for SWIFT and while keeping the one for MCTP at 10k.

### DIFF
--- a/wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
+++ b/wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
@@ -32,9 +32,7 @@ type HookReturn = {
 };
 
 const QUOTE_REFRESH_INTERVAL = 20_000;
-
-const MAYAN_BETA_LIMIT = 10_000; // USD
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/hooks/useRoutesQuotesBulk.ts
```diff
@@ -32,9 +32,7 @@ type HookReturn = {
 };
 
 const QUOTE_REFRESH_INTERVAL = 20_000;
-
-const MAYAN_BETA_LIMIT = 10_000; // USD
-const MAYAN_BETA_PROTOCOLS = ['MCTP', 'SWIFT'];
+const MAYAN_BETA_PROTOCOL_LIMITS = { MCTP: 10_000, SWIFT: null };
```
artursapek, line 35: just remove `SWIFT` from this object. there's a third protocol `'WH'` which also doesn't have a limit.

https://github.com/mayan-finance/wormhole-sdk-route/blob/main/src/index.ts#L525

emreboga:
Closing in favor of https://github.com/wormhole-foundation/wormhole-connect/pull/3247

PR #3258, opened 2025-02-07
Fix bugs in token price fetching which cause excessive API calls

artursapek:
There is two bugs in this code I wrote during the arbitrary token refactor. These are causing us to make excessive requests to Coingecko (basically making each call twice 💀)

- `getTokenPrice` can call the `updateTokenPrices` callback multiple times for the same token because it wasn't aware of which tokens in `tokenPricesToFetch` were already being handled in an ongoing request.
- Also, using a `Set` for `tokenPricesToFetch` wasn't the right call; adding `TokenId`s to a `Set` doesn't work because they're just plain objects and jerbascript compares objects by instance and not value. I switched to a `TokenMapping` which is a better way to track a set of unique `TokenId`s.

### DIFF
--- a/wormhole-connect/src/contexts/TokensContext.tsx
+++ b/wormhole-connect/src/contexts/TokensContext.tsx
@@ -59,6 +59,10 @@ export const TokensProvider: React.FC<TokensProviderProps> = ({ children }) => {
   const [tokenPrices, _setTokenPrices] = useState<TokenMapping<TokenPrice>>(
     new TokenMapping(),
   );
+  const [tokenPricesToFetch, _setTokenPricesToFetch] = useState<
+    TokenMapping<boolean>
+  >(new TokenMapping());
+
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/contexts/TokensContext.tsx
```diff
@@ -104,21 +105,34 @@ export const TokensProvider: React.FC<TokensProviderProps> = ({ children }) => {
     [],
   );
 
-  const tokenPricesToFetch: Set<TokenId> = new Set();
+  const tokenPricesToFetch: TokenMapping<boolean> = new TokenMapping();
```
kev1n-peters: I think this should be a state variable like `tokenPrices`

artursapek: seems to work as-is but I can make it one

kev1n-peters (dismissed):

kev1n-peters (approved):

PR #3266, opened 2025-02-10
React 19 compatibility

artursapek:
These changes make Connect work out of the box in a fresh Next.js app running React 19.

- relaxes React requirement from exactly `18.2.0` to `>=18.0.0` (which will work with 19)
- removes `react` and `react-dom` from `dependencies` - having them only in `peerDependencies` is more appropriate for a component library
- adds guards around `localStorage` references, which were throwing errors before in Next.JS because it pre-renders pages in Node where `localStorage` is not defined
- fixes Vite config so that the JSX runtime from the surrounding application is used correctly


wormhole-connect/src/config/tokens.ts
```diff
@@ -21,6 +21,8 @@ import { getTokenMetadataFromRpc } from 'utils/tokens';
 
 const TOKEN_CACHE_VERSION = 1;
 
+const HAS_LOCALSTORAGE = typeof localStorage !== 'undefined';
```
artursapek, line 24: this was throwing errors in server side rendering because `localStorage` is not defined

emreboga (approved):
LGTM!

PR #3275, opened 2025-02-13
Add custom coingecko url

valentinoConti:
### Description

Adds a new optional prop to the Connect component to use a custom coingecko URL

### DIFF
--- a/wormhole-connect/src/config/index.ts
+++ b/wormhole-connect/src/config/index.ts
@@ -120,6 +120,7 @@ export function buildConfig(
       Devnet: ['http://localhost:7071'],
     }[network],
     coinGeckoApiKey: customConfig.coinGeckoApiKey,
+    coinGeckoCustomUrl: customConfig.coinGeckoCustomUrl,
 
     // Callbacks
     triggerEvent: wrapEventHandler(customConfig.eventHandler),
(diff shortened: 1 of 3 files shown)


yuli-ferna (dismissed):
lgtm!

artursapek (changes requested):
we've been trying to keep the config organized. for example, there is the `ui` key with lots of properties inside it. 

rather than having multiple coingecko-related properties on the top level, can we start a `coingecko` key which is a similar object containing multiple properties?

yuli-ferna:
> we've been trying to keep the config organized. for example, there is the `ui` key with lots of properties inside it.
> 
> rather than having multiple coingecko-related properties on the top level, can we start a `coingecko` key which is a similar object containing multiple properties?

@artursapek 
Like this?:
`coinGecko: { apiKey?: string, customUrl?: string} `

artursapek:
> > we've been trying to keep the config organized. for example, there is the `ui` key with lots of properties inside it.
> > rather than having multiple coingecko-related properties on the top level, can we start a `coingecko` key which is a similar object containing multiple properties?
> 
> @artursapek Like this?: `coinGecko: { apiKey?: string, customUrl?: string} `

that looks pretty gud

artursapek (approved):

PR #3281, opened 2025-02-14
Revert "React 19 compatibility"

emreboga:
I'm seeing an incompatibility with xlabs libs in local developemtn after React 19 upgrade. We need to revert this change to ublock local development until this incompatibility is resolved.

Screenshot 2025-02-14 at 2 39 13 PM

Reverts wormhole-foundation/wormhole-connect#3266


emreboga:
Not needed anymore.

PR #3287, opened 2025-02-18
Berachain support

kev1n-peters:


wormhole-connect/src/config/types.ts
```diff
@@ -64,7 +64,7 @@ export enum TokenIcon {
   'XLAYER',
   'MANTLE',
   'WORLDCHAIN',
-  'UNICHAIN',
+  'BERA',
```
artursapek, line 67: why'd you remove `UNICHAIN` here?

kev1n-peters: The gas token on Unichain is ETH and so Connect uses the Ethereum icon when displaying the Uniswap gas token. TokenIcon.UNICHAIN is not used so I just removed it

artursapek (approved):

PR #3289, opened 2025-02-19
make useTransactionHistoryWHScan more robust

artursapek:
- make token amount & USD amount optional
- make token itself optional
- use `getOrFetchToken` to fetch tokens referenced in history items which are not yet in the token cache


emreboga (approved):
LGTM!

PR #3302, opened 2025-02-20
Fix circular progress animation styles

emreboga:
There has been an issue the way we were animating the circular progress in one smooth move after React 19 upgrade. This was because an added animation was not triggered all the time after re-render of the component. It was very inconsistent.
The fix here is remove the old logic and do not depend on the animation to simulate the progress, but use the percent value and add an animation to make the steps in that progress upgrade a bit smoother.

https://www.loom.com/share/1a8417d6021041899bc0564d94abe105?sid=6277fbd0-592b-4deb-87d0-23989896af9f

### DIFF
--- a/wormhole-connect/src/views/v2/Redeem/index.tsx
+++ b/wormhole-connect/src/views/v2/Redeem/index.tsx
@@ -66,11 +66,7 @@ import TxReadyForClaim from 'icons/TxReadyForClaim';
 import { useGetRedeemTokens } from 'hooks/useGetTokens';
 import { tokenIdFromTuple } from 'config/tokens';
 
-type StyleProps = {
-  transitionDuration?: string | undefined;
-};
-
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/views/v2/Redeem/index.tsx
```diff
@@ -117,7 +113,7 @@ const useStyles = makeStyles<StyleProps>()((theme, { transitionDuration }) => ({
   },
   circularProgressCircleDeterminite: {
     strokeLinecap: 'round',
-    transitionDuration,
+    transitionDuration: '1s',
```
emreboga, line 116: Progress updates every second. We add this transition duration for 1s, so it won't feel like stop/start between those updates.

artursapek (approved):
so smoooooth

PR #3304, opened 2025-02-20
Don't run `check-token-list` unless the token JSON files changed

artursapek:
- dont run the "check foreign assets" script if those files didn't change...
- make that script run faster by not making token bridge calls on chains that we don't even have an address configured for in the first place (this is OK to do since Connect can look them up dynamically now)
- make the output of that script actually kind of useful:
image


emreboga (dismissed):

emreboga (approved):
LGTM.

PR #3323, opened 2025-02-26
Use the correct route name

emreboga:
The local name was MayanSwapSHUTTLE but the one from the route object is MayanRouteSHUTTLE, so we need to use that one.

### DIFF
--- a/wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
+++ b/wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
@@ -81,7 +81,7 @@ const useFetchSupportedRoutes = (): HookReturn => {
         // TODO token refactor
         if (
           route.rc.name.includes('Mayan') &&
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
```diff
@@ -81,7 +81,7 @@ const useFetchSupportedRoutes = (): HookReturn => {
         // TODO token refactor
         if (
           route.rc.name.includes('Mayan') &&
-          route.rc.name !== 'MayanSwapSHUTTLE'
+          route.rc.name !== 'MayanRouteSHUTTLE'
```
kev1n-peters, line 84: Doesn't `route.rc.name` come from here? https://github.com/mayan-finance/wormhole-sdk-route/blob/bd5b99064a342abd6bba464b6cb83ccc60105021/src/index.ts#L580

kev1n-peters (approved):

PR #3330, opened 2025-02-28
aptos cctp: support aptos connect

kev1n-peters:


artursapek (approved):
I think just saying "a compatible wallet for this route" without listing any options is unideal, but probably a very niche problem.

PR #3336, opened 2025-03-03
export additional types used in callbacks

kev1n-peters:


kev1n-peters:
Noticed that the `TransferDetails` parameter of `isRouteSupportedHandler` was not exported (and some other parameters used in callbacks)

artursapek (approved):

PR #3338, opened 2025-03-04
export telemetry

kev1n-peters:


wormhole-connect/src/index.ts
```diff
@@ -56,9 +49,6 @@ export {
   Chain,
   WormholeConnectTheme,
   WormholeConnectEvent,
```
artursapek: can u get rid of this line now?

kev1n-peters: You got it

artursapek (approved):

PR #3345, opened 2025-03-07
Added isTokenSupportedHandler callback to config

kev1n-peters:
Useful for integrators to filter out tokens using arbitrary logic

### DIFF
--- a/wormhole-connect/src/config/types.ts
+++ b/wormhole-connect/src/config/types.ts
@@ -88,6 +88,8 @@ export type IsRouteSupportedHandler = (
   transferDetails: TransferDetails,
 ) => Promise<boolean>;
 
+export type IsTokenSupportedHandler = (token: Token) => boolean;
+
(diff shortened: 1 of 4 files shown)


wormhole-connect/src/config/types.ts
```diff
@@ -88,6 +88,8 @@ export type IsRouteSupportedHandler = (
   transferDetails: TransferDetails,
 ) => Promise<boolean>;
 
+export type IsTokenSupportedHandler = (token: Token) => boolean;
```
kev1n-peters, line 91: Do we want to keep `Token` as an internal type and instead pass `TokenId` or something else here?

kev1n-peters: It might be useful to pass `Token` though since it has `tokenBridgeOriginalTokenId` for example and `symbol`, etc.

artursapek (approved):

PR #3353, opened 2025-03-12
Use a Slide action for Asset picker in mobile

emreboga:
This PR will change the popover UX for mobile web. It'll be a Slide action instead of Grow in mobile.


wormhole-connect/src/views/v2/Bridge/AssetPicker/index.tsx
```diff
@@ -62,13 +63,17 @@ const useStyles = makeStyles()((theme: any) => ({
   popover: {
     marginLeft: '-1px',
     marginTop: '-1px',
-    width: '422px',
+    width: '420px',
```
kev1n-peters: I don't mean to be pedantic but wanted to point out that I can see the button behind the popover now after reducing the width by 2px (the little purple on the right side here). Not sure if it matters
image

emreboga: Yes, I changed the way mobile experience is going to be built (using a different component) and this will also be fixed with that.

emreboga: Fixed in https://github.com/wormhole-foundation/wormhole-connect/pull/3353/commits/c0262bd3fb4ed581f4fb0202130f8e8555736dfc

kev1n-peters (approved):

PR #3363, opened 2025-03-17
Check multiple props for the relayer fee

### LINKED ISSUE #2845: [QA] Sent and received amounts on Transaction history screen are identical when using USDT Bridge
Checked on Netlify with following [config](https://deploy-preview-2836--wormhole-connect-mainnet.netlify.app/?config=N4AgTg9grgLgpgZwFwgNoDpMBECiAxAQQFUAZAFQH0AlAeSLJwGUAaEAWQEMBPDgOyujwAuiAC%2BQA%3D%3D%3D). 

Steps:
1. Send 1 USDT from Polygon (MetaMask) to Celo (MetaMask)
2. Open Transaction history screen
3. Check amounts

Actual result: Amounts on Transaction history screen differ from amounts on Transaction progress screen.
Expected result: The same amounts are shown.

## Notes/ Attachments
Link to [Wormholescan](https://wormholescan.io/#/tx/0xf49ba01d0758f83dfcdc026d6dc8efb938a1a2ef523188f747582f3cb37d277c)
Send from tx: [Polygonscan](https://polygonscan.com/tx/0xf49ba01d0758f83dfcdc026d6dc8efb938a1a2ef523188f747582f3cb37d277c)
Send to tx: [Celo Explorer](https://explorer.celo.org/mainnet/tx/0x5b5bb1f22ca7b7bc31294226ed00872931bdbd1e0406e2709a386ceda52edf59)

Screenshot 2024-10-15 at 15 42 26

emreboga:
WHScan API should parse the payload and set it under `standarizedProperties`, but that hasn't been the case. There are multiple props that can persist the fee value depending on the protocol. There is no clear documentation around the set of props specific to protocols, therefore we need to check them all.

Fixes https://github.com/wormhole-foundation/wormhole-connect/issues/2845

### DIFF
--- a/wormhole-connect/src/hooks/useTransactionHistoryWHScan.ts
+++ b/wormhole-connect/src/hooks/useTransactionHistoryWHScan.ts
@@ -21,10 +21,12 @@ interface WormholeScanTransaction {
   content: {
     payload: {
       amount: string;
+      fee?: string;
       callerAppId: string;
       fromAddress: string;
       parsedPayload: {
-        feeAmount: string;
+        feeAmount?: string;
+        relayerFee?: string;
         recipientWallet: string;
         toNativeAmount: string;
       };
@@ -127,7 +129,7 @@ const useTransactionHistoryWHScan = (
   const parseSingleTx = useCallback(
     async (tx: WormholeScanTransaction) => {
       const { content, data, sourceChain, targetChain } = tx;
-      const { standarizedProperties } = content || {};
+      const { standarizedProperties, payload } = content || {};
 
       const fromChainId =
         standarizedProperties.fromChain || sourceChain?.chainId;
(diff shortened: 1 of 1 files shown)


wormhole-connect/src/hooks/useTransactionHistoryWHScan.ts
```diff
@@ -193,10 +195,21 @@ const useTransactionHistoryWHScan = (
...
+          standarizedProperties.fee ||
+          payload.fee ||
+          payload.parsedPayload?.feeAmount ||
+          payload.parsedPayload?.relayerFee ||
+          '0';
+
         const receiveAmountValue =
-          BigInt(standarizedProperties.amount) -
-          BigInt(standarizedProperties.fee);
+          BigInt(standarizedProperties.amount) - BigInt(fee);
```
kev1n-peters, line 211: I don't think we can subtract the fee from the received amount for every protocol. For example, the relayer fee for automatic NTT transfers is paid in native gas units on the source chain and shouldn't be subtracted from the received amount. We need to figure what units the fee is in for each protocol before determining if we should subtract it from the receive amount

emreboga: @kev1n-peters it's not subtracting from received amount, but from the sent amount to calculate received amount. Is that what you meant? By the way gas fee on the source chain is specified in sourceChain property in the API response. Do you think NTT fees could be part of it:
```
"sourceChain": {
    "chainId": 23,
    "timestamp": "2025-03-18T11:01:59Z",
    "transaction": {
        "txHash": "0x109caadef8649277e89ddad069d040e49dd73c08fa1acc6956535466be4854a6"
    },
    "from": "0x49887a216375fded17dc1aaad4920c3777265614",
    "status": "confirmed",
    "fee": "0.0000025605",
    "gasTokenNotional": "0.360789",
    "feeUSD": "0.0000009238002345"
}
```

kev1n-peters: What I meant was that `standardizedProperties.amount` and `fee` may be denominated in different tokens depending on the protocol. I just looked at an automatic NTT transfer [here](https://wormholescan.io/#/tx/0x119909ae738a621fb65cd9915c62761fea6b6bb321263350ef0fd50412ba25a4?view=advanced), but the (relayer) fee in `standardizedProperties` is not being set by the API (it's set to an empty string even though a relayer fee was actually paid). Since they're not setting `standardizedProperties.fee` for automatic NTT transfers, what you have is probably okay.

emreboga: Ideally the relayer fee, wherever it resides, should be parsed and set in `standardizedProperties.fee` by the API. I don't see why else it's called standardized :)

kev1n-peters (approved):

PR #3364, opened 2025-03-18
Added ManualTBTC builtin route support

kev1n-peters:


wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
```diff
@@ -91,6 +91,15 @@ const useFetchSupportedRoutes = (): HookReturn => {
         }
       });
 
+      // HAX - We don't want users to use the token bridge route when the TBTC route is available
```
artursapek, line 94: if we're trying to avoid this, then wouldn't we want to add some logic into `useComputeDestinationTokens`, so that the user isn't even presented with frankenstein TBTC as an option in the output tokens list? this hook is downstream, after the `destToken` has already been selected.

kev1n-peters: Talked about this offline and we have the `isFrankensteinToken()` check that will prevent it from being presented

artursapek (dismissed):

artursapek (approved):

PR #3367, opened 2025-03-19
UI test automation

emreboga:
Adding the boilerplate and the first test for UI automation.


DBosley (dismissed):
Looks good!

wormhole-connect/tests/e2e/specs/mayan-swift.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+// Read from .env* files
+// This is only for local testing overrides
+dotenv.config({ path: path.resolve(__dirname, '../../..', '.env') });
+dotenv.config({ path: path.resolve(__dirname, '../../..', '.env.local') });
+
+/**
+ * Default config will have the following settings:
+ * {
+      network: 'mainnet',
+      coinGeckoApiKey: 'CG-EDLftLaFEWYqZNsPVwBDAKCE',
```
kev1n-peters: I don't think we want to expose this. Can we set it via env var, create the config object here and then base64 encode it?

wormhole-connect/src/WormholeConnect.tsx
```diff
@@ -26,6 +26,15 @@ export default function WormholeConnect({
   config,
   theme,
 }: WormholeConnectProps) {
+  React.useEffect(() => {
+    // IMPORTANT: This is a workaround to expose the Redux store to the window object so it can be used in automated tests.
+    if (!globalThis.dispatchReduxAction) {
```
kev1n-peters, line 31: Does this function not get set on the window when tests are run? Is there a way to only set it in when we're in headless test mode? Maybe I'm just being paranoid

wormhole-connect/tests/e2e/specs/mayan-swift.spec.ts
emreboga: @kev1n-peters good point! We can add this API key to a secret as well.

wormhole-connect/src/WormholeConnect.tsx
emreboga: We need it in `globalThis` to be able to access from the test, same for both headed and headless. What do you think the potential risk here? When I thought about it, I couldn't find anything that I can do via redux actions but not from the UI itself.

kev1n-peters: Ah, so this was just never being set in the first place. Nvm then

wormhole-connect/tests/e2e/specs/mayan-swift.spec.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3367/commits/8ac9587640b3c5f8bd27b842a566646070ed807b

kev1n-peters (approved):

PR #3368, opened 2025-03-19
Remove legacy logic from TokenList filter function

kev1n-peters:
The purpose of this change is to make the TokenList show the token when the search query matches the token exactly. There was some additional filtering logic that is no longer needed. If the token has 0 balance it will be grayed out.


evgeniko (approved):
lgtm

PR #3384, opened 2025-03-26
Remove Portico route (ETH, wstETH, UDST Bridge)

kev1n-peters:


wormhole-connect/src/hooks/useTransactionHistoryWHScan.ts
```diff
@@ -300,68 +283,6 @@ const useTransactionHistoryWHScan = (
     [parseSingleTx],
   );
 
-  // Parser for Portico transactions (appId === ETH_BRIDGE or USDT_BRIDGE)
-  // IMPORTANT: This is where we can add any customizations specific to Portico data
-  // that we have retrieved from WHScan API
-  const parsePorticoTx = useCallback(
```
emreboga, line 306: We will stop showing portico route but if we remove the parsing it won't show previously created txs as well. Is that ok?

kev1n-peters: I don't have a strong opinion on this. I guess a benefit of removing it is less code. But obviously older TXs won't be shown.

kev1n-peters: @Wesleyleung What should we do here? Show old Portico route TXs in the history or remove support for that?

kev1n-peters:
Holding off on merging this

tonyjin:
Closed in favor of #3643

PR #3401, opened 2025-04-04
Set route to Bridge on re-render

DBosley:
If a config change occurs due to client side rendering, we should clear transfers and navigate to the bridge view


artursapek (approved):
seems fine to me

kev1n-peters (approved):

PR #3402, opened 2025-04-07
Theme prop to override transparent bg for asset pickers and amount input

emreboga:
Transparent BG was added before to indicate selection of assets or entering the amount value. However it's causing visual issues with noisy BGs. The proposed solution is to add a new theme prop to override the default transparent value when necessary.

**Before:**
Screenshot 2025-04-07 at 8 29 54 PM

**After:**
Screenshot 2025-04-07 at 8 29 46 PM

### DIFF
--- a/wormhole-connect/src/components/SampleApp/index.tsx
+++ b/wormhole-connect/src/components/SampleApp/index.tsx
@@ -370,6 +370,10 @@ function SampleApp() {
                       <i>string;</i>
                     </li>
                     <li>
+                      <pre>inputFill</pre>
+                      <i>string;</i>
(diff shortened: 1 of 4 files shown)


artursapek (commented):
I still dont understand what you're trying to achieve here. there is already a theme property for modifying the input backgrounds, called `input`.

image


(this is https://connect-in-style.wormhole.com/)

your change appears redundant with that. if youre trying to get rid of the transparent treatment that inputs get when they're filled in, why not just make that a boolean on the theme that lets people disable that?

artursapek (approved):
lgtm

PR #3406, opened 2025-04-08
Remove Content-Type from CoinGecko GET requests

kev1n-peters:
None of the requests contain a payload


artursapek (approved):
cool, how did you even notice this btw?

PR #3414, opened 2025-04-11
Remove final remnants of MultiProvider (not needed anymore)

artursapek:
All we need to do is call `w.getSigner()` inside of `signAndSendTransaction` directly. This signer cache thing has no purpose anymore - it was an unnecessary middleman which I've cut out in this PR.

### DIFF
--- a/wormhole-connect/src/utils/wallet/evm.ts
+++ b/wormhole-connect/src/utils/wallet/evm.ts
@@ -1,4 +1,4 @@
-import { Wallet } from '@xlabs-libs/wallet-aggregator-core';
+import { Wallet, NotSupported } from '@xlabs-libs/wallet-aggregator-core';
 import {
   BinanceWallet,
   EVMWallet,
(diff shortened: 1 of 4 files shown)


emreboga (dismissed):

emreboga:
@artursapek I see this error in the preview deployment when confirming a transaction:
`Error: Signer is not connected to the right chain. Expected 42161, got undefined`
However when I pull this branch and try that locally I don't see the same error.

emreboga (approved):

wormhole-connect/src/utils/wallet/evm.ts
```diff
@@ -90,31 +90,31 @@ export interface AssetInfo {
...
   if (!signer) throw new Error('No signer found for chain' + chainName);
 
   // Ensure the signer is connected to the correct chain
-  const provider = await signer.provider?.getNetwork();
   const expectedChainId = request.transaction.chainId
     ? ethers.getBigInt(request.transaction.chainId)
     : undefined;
-  const actualChainId = provider?.chainId;
 
-  if (!actualChainId || !expectedChainId || actualChainId !== expectedChainId) {
```
kev1n-peters, line 114: I think we should still have this check. We added it because somebody in the past was able to sign and send a transaction to a network different than the network ID on the transaction object.

kev1n-peters: We should do this check after the switch chains call to make sure the wallet actually switched to the right chain

PR #3416, opened 2025-04-11
Don't try to switch chains if it's already correct

artursapek:
This fixes an issue where the wallet aggregator library hangs forever on the `switchChains` call if the wallet is already on that chain.

### DIFF
--- a/wormhole-connect/src/AppRouter.tsx
+++ b/wormhole-connect/src/AppRouter.tsx
@@ -48,10 +48,10 @@ function AppRouter(props: Props) {
   const dispatch = useDispatch();
   const routeContext = useContext(RouteContext);
   const route = useSelector((state: RootState) => state.router.route);
-  
+
   const hasSetSsgConfig = useRef(false);
   const isInitialLoad = useRef(true);
-  
+
   // We update the global config once when WormholeConnect is first mounted, if a custom
   // config was provided.
   //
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/utils/wallet/evm.ts
```diff
@@ -96,29 +96,39 @@ export async function signAndSendTransaction(
...
+          );
+        } else {
+          throw new Error(`Error switching to chain ${expectedChainId}: ${e}`);
+        }
       }
     }
   } else {
-    console.warn(`EVM transaction has no chainId`, request.transaction);
+    console.warn(
+      `EVM signer has no chainId; cannot verify that it matches transaction`,
```
artursapek: should we just throw here? @kev1n-peters

wormhole-connect/src/AppRouter.tsx
```diff
@@ -72,7 +72,7 @@ function AppRouter(props: Props) {
   if (!hasSetSsgConfig.current) {
     // This runs once in SSG step (server-side pre-rendering)
     if (props.config) {
-        loadConfig(props.config);
+      loadConfig(props.config);
```
artursapek, line 75: i ran `lint`... 🙄

wormhole-connect/src/utils/wallet/evm.ts
```diff
@@ -96,29 +97,55 @@ export async function signAndSendTransaction(
   const signer = await (w as any).getSigner();
```
kev1n-peters: `w` should always be an `EVMWallet` so we could create a `const w as EVMWallet` for better type checking and use it in places where we're doing the typecast

artursapek: image

emreboga (dismissed):

kev1n-peters (approved):

PR #3422, opened 2025-04-15
Do not fetch quotes when tab is inactive

emreboga:
This will prevent the onComplete callback to create another timeout and continue fetching.


emreboga:
Closing in favor of another work that includes this functionality as well.

PR #3424, opened 2025-04-15
Improvements to quote refreshing

artursapek:
- Refactor the hooks `useSortedRoutesWithQuotes`, `useFetchSupportedRoutes`, `useFetchQuotes` to be a proper pipeline where we fetch support routes first, before trying to fetch quotes.
  - *Only `Bridge/index.ts` should be reading from Redux, and passing these values in to the hooks as arguments. Downstream hooks should not be reading directly from Redux because this makes them fire too soon.* 
  - This fixes a subtle bug where we would immediately start fetching quotes after changing tokens, before even figuring out which routes we should be using first. This was wasteful and always being thrown away
- Use the new `expires` property in an SDK `Route` quote and cache quote values until then
- Pause all quote refreshing when the tab is hidden, and immediately refresh the ones that have expired when tab becomes visible again

### DIFF
--- a/wormhole-connect/src/routes/operator.ts
+++ b/wormhole-connect/src/routes/operator.ts
@@ -302,6 +302,8 @@ class QuoteCache {
             result.expires = new Date(Date.now() + 60_000);
           }
 
+          console.debug(`Fetched quote`, routeName, result);
+
           this.cache[key] = new QuoteCacheEntry(result);
         })
         .catch((err: any) => {
--- a/wormhole-connect/src/routes/operator.ts
+++ b/wormhole-connect/src/routes/operator.ts
@@ -47,6 +47,7 @@ export interface QuoteParams {
   destToken: Token;
   amount: sdkAmount.Amount;
   nativeGas: number;
+  recipient?: string; // wallet may be undefined when not connected
 }
 
 export default class RouteOperator {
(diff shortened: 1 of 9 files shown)


DBosley (dismissed):
The changes look good to me and I can see the difference in quote refreshes. Another set of eyes to verify would be a good idea, though.

wormhole-connect/src/routes/operator.ts
```diff
@@ -329,8 +359,21 @@ class QuoteCacheEntry {
...
-  age(): number {
-    return new Date().valueOf() - this.timestamp.valueOf();
+  // Number of seconds this quote is still valid for before we should fetch a new one
+  expires(): Date {
+    if (this.result.success) {
+      // For a successful quote, if it specifies an expiry we return that
+      // otherwise we default to a TTL 1 minute
+      return this.result.expires ?? new Date(this.timestamp.valueOf() + 60_000);
+    } else {
+      // We cache errors for 10 seconds
```
emreboga, line 371: It says 10 secs in the comment, but caches for 2 mins below. I assume we set the expiration to two mins after an error?

kev1n-peters (approved):

wormhole-connect/src/routes/operator.ts
artursapek: oops

emreboga (approved):
One minor comment to be changed about the cache timeout after errors, but the rest is LGTM!

PR #3426, opened 2025-04-16
Reduce RPC calls with async request caching

DBosley:
- Add generic AsyncCache util
- Use AsyncCache to reduce RPC calls from route.ts

### DIFF
--- a/wormhole-connect/src/routes/sdkv2/route.ts
+++ b/wormhole-connect/src/routes/sdkv2/route.ts
@@ -16,6 +16,7 @@ import { Token } from 'config/tokens';
 import { SDKv2Signer } from './signer';
 
 import { amount as sdkAmount } from '@wormhole-foundation/sdk';
+import { AsyncCache } from 'utils/AsyncCache';
 import config, { getWormholeContextV2 } from 'config';
 import { sleep } from 'utils';
 import { isFrankensteinToken } from 'utils';
@@ -35,6 +36,8 @@ export class SDKv2Route {
       'CosmosGateway',
     ].includes(rc.meta.name);
   }
+
+  private tokenCache = new AsyncCache<TokenId[]>(24 * 60 * 60 * 1000); // 24 hour TTL
 
   get AUTOMATIC_DEPOSIT() {
     return this.rc.IS_AUTOMATIC;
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/utils/AsyncCache.ts
```diff
@@ -0,0 +1,58 @@
...
+  value: T;
+  timestamp: number;
+  isError?: boolean;
+}
+
+export class AsyncCache<T> {
+  private cache = new Map<string, CacheEntry<T>>();
+  private pendingRequests = new Map<string, Promise<T>>();
+
+  constructor(private ttl: number) {}
```
kev1n-peters: Minor nit: can we append the units here? It's not obvious to me if this should be seconds or milliseconds from reading the signature

wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -71,20 +74,28 @@ export class SDKv2Route {
...
-      return isSameToken(tokenId, destToken);
-    });
+    if (!fromChainSupported || !toChainSupported) {
+      return false;
+    }
 
-    const isSupported =
-      fromChainSupported && toChainSupported && toTokenSupported;
+    try {
+      const cacheKey = `isRouteSupported-${sourceToken.address}-${fromChain}-${toChain}`;
```
kev1n-peters: Are there any stats about how many RPC calls caching saves us?

artursapek: on a huddle with Dave I witnessed Connect make zero calls when switching back to a chain that had been previously selected.

wormhole-connect/src/utils/AsyncCache.ts
DBosley: Sure thing

wormhole-connect/src/routes/sdkv2/route.ts
DBosley: With just SUI and ETH rpcs enabled, it brought calls down from 36 when opening up the view to select a destination down to 24, then it cached them so that when re-opening the same network tab there would be zero new calls.

wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -71,20 +74,28 @@ export class SDKv2Route {
...
+      const supportedDestinationTokens = await this.tokenCache.requestWithCache(
```
artursapek: I think we could DRY this up, by using this class's `supportedDestTokens` method here instead of hitting the cache directly.

artursapek (changes requested):

wormhole-connect/src/routes/sdkv2/route.ts
DBosley: The only issue is that they pass in different parameters. I don't know exactly why that's the case, but I want to make sure it works both ways

DBosley: Fixed it up

artursapek (approved):

PR #3429, opened 2025-04-18
Enable running tests against an external URL

emreboga:
This will add the optional input to the GH action, so we can provide an external URL to test against. We can run configuration tests (no submission of the actual transaction) against external URLs, but we need to pass custom configs for transaction submission, which can be done in local dev only for now.
A sample test config would be like:

```
{
    name: 'MayanSwapSWIFT',
    config: mayanSWIFT,
    sourceWallet: {
      address: process.env.REACT_APP_TEST_EVM_ADDR || '',
      privateKey: process.env.REACT_APP_TEST_EVM_PK || '',
    },
    sourceAsset: {
      chain: 'arbitrum',
      symbol: 'USDC',
      address: CONTRACTS.Arbitrum.USDC,
    },
    destinationWallet: {
      address: process.env.REACT_APP_TEST_EVM_ADDR || '',
    },
    destinationAsset: {
      chain: 'base',
      symbol: 'USDC',
      address: CONTRACTS.Base.USDC,
    },
    amount: '1',
    waitForCompletion: true,
  },
```


wormhole-connect/tests/e2e/config/constants.ts
```diff
@@ -0,0 +1,14 @@
+export const CONTRACTS = {
+  Arbitrum: {
+    USDC: '0xaf88d065e77c8cC2239327C5EDb3A432268e5831',
```
artursapek: please use the SDK to get these consts instead of establishing this redundant file here. here are two [examples](https://github.com/wormhole-foundation/wormhole-connect/blob/ee22eb93cb93e2d297451624f1daa95135b81d68/wormhole-connect/src/hooks/useRoutesQuotesBulk.ts#L220) of [how to do it](https://github.com/wormhole-foundation/wormhole-connect/blob/ee22eb93cb93e2d297451624f1daa95135b81d68/wormhole-connect/src/utils/sdkv2.ts#L274-L277)

artursapek (changes requested):

wormhole-connect/tests/e2e/config/constants.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3429/commits/7f4cbee4b27adfffc02ba6cbb72c68042ed0578e

emreboga: After a second thought, removed CONTRACTS and used the token address in config directly:
https://github.com/wormhole-foundation/wormhole-connect/pull/3429/commits/e577b2c8dbf85820234dee4c3265c0b45074dcb9

wormhole-connect/tests/e2e/config/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  network: 'mainnet',
+  coinGeckoApiKey: '${COINGECKO_API_KEY}',
+  ui: {
+    showInProgressWidget: true,
+    testOptions: {
+      enableHeadlessSigner: true,
+    },
+  },
+  routes: [
+    AutomaticCCTPRoute,
```
artursapek: now that we have the CCTP executor route, maybe we should test that one instead?

emreboga: Oh yeah, that needs to change 👍

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3429/commits/df741fddaf6c1f3c8e5f86263e4f35756ccf57d6

artursapek (approved):

PR #3438, opened 2025-04-22
Add gas drop changes for Executor routes

emreboga:
Please see how it works in the loom below.
https://www.loom.com/share/ced4fc648a8044f5954b378fe5b7f485?sid=33ddc8b5-e836-4734-ba08-b5ba031a3570

### DIFF
--- a/wormhole-connect/src/store/transferInput.ts
+++ b/wormhole-connect/src/store/transferInput.ts
@@ -71,6 +71,7 @@ export interface TransferInputState {
   amount?: amount.Amount;
   receiveAmount: DataWrapper<string>;
   route?: string;
+  isExecutorRoute?: boolean; // TODO Replace this when we have a similar prop in route object
   preferredRouteName?: string | undefined;
   balances: WalletBalances;
   foreignAsset: string;
@@ -117,6 +118,7 @@ function getInitialState(): TransferInputState {
     receiveAmount: getEmptyDataWrapper(),
     preferredRouteName: config.ui.defaultInputs?.preferredRouteName,
     route: undefined,
+    isExecutorRoute: false,
     balances: {},
     foreignAsset: '',
     associatedTokenAddress: '',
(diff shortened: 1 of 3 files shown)


wormhole-connect/src/store/transferInput.ts
```diff
@@ -71,6 +71,7 @@ export interface TransferInputState {
   amount?: amount.Amount;
   receiveAmount: DataWrapper<string>;
   route?: string;
+  isExecutorRoute?: boolean; // TODO Replace this when we have a similar prop in route object
```
kev1n-peters: If we had an IS_EXECUTOR_ROUTE static property on the Route object, would we be able to remove this property from the store?

emreboga: We will be able to remove this check:
`state.isExecutorRoute = payload.toLowerCase().endsWith('w7executorroute');`
but we will probably keep the redux property to have an easy access across all components.

wormhole-connect/src/views/v2/Bridge/GasSlider/index.tsx
```diff
@@ -83,14 +83,16 @@ const GasSlider = (props: {
   const dispatch = useDispatch();
   const theme = useTheme();
 
-  const { toChain: destChain } = useSelector(
-    (state: RootState) => state.transferInput,
-  );
+  const {
+    fromChain: sourceChain,
+    toChain: destChain,
+    isExecutorRoute,
```
kev1n-peters: I think there is a bug where if we select the executor route, then `isExecutorRoute` gets set to true in the store. Then, when we render this component for a non-executor route (e.g. automatic cctp bridge), it will think the route is an executor route, but it's not. Maybe instead of reading this value from the store, we can calculate it based on the route and pass it as a prop here: https://github.com/wormhole-foundation/wormhole-connect/blob/development/wormhole-connect/src/views/v2/Bridge/Routes/SingleRoute.tsx#L653

emreboga: Fair enough 👍

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3438/commits/5340a766594717b755938ba89a88287267784fe4

kev1n-peters (approved):

PR #3444, opened 2025-04-24
Added getTokenMetadataSui function

kev1n-peters:


kev1n-peters:
Fixes an issue where you could paste in a non-token address on Sui (e.g. an object address: 0x9258181f5ceac8dbffb7030890243caed69a9599d2886d957a9cb7656af3bdb3) , and it would get added to the tokens cache. However, this would cause the fetch balances call to throw when fetching the balance for this non-token.

artursapek (approved):

PR #3445, opened 2025-04-24
Color updates for CTA and route cards

emreboga:
This PR also includes the addition of Color library to replace our internal theme utils. Please checkout the PR deployment below for the end result.


wormhole-connect/src/components/v2/Button.tsx
```diff
@@ -13,15 +13,19 @@ const PrimaryButton = styled(MUIButton)<ButtonProps>(
     boxShadow: 'none',
     backgroundColor: theme.palette.button.primary,
     color: theme.palette.button.primaryText,
-    '&:disabled': {
+    '&.Mui-disabled': {
```
emreboga, line 16: These styles weren't working without the MUI class overrides.

wormhole-connect/src/views/v2/Bridge/WalletConnector/index.tsx
```diff
@@ -14,47 +13,6 @@ import { TransferWallet } from 'utils/wallet';
 import { TransferSide } from 'config/types';
 import WalletSidebar from './Sidebar';
 
-const useStyles = makeStyles()((theme: any) => ({
```
emreboga, line 17: I saw that these styles in Wallet connect button were already part of the Button component `src/components/v2/Button.tsx`

artursapek (commented):
text should go to white at a certain threshold.

```
{ mode: 'light', primary: '#222' }
```

image

wormhole-connect/src/views/v2/Bridge/WalletConnector/index.tsx
```diff
@@ -14,47 +13,6 @@ import { TransferWallet } from 'utils/wallet';
...
-      boxShadow: 'none',
-      backgroundColor: theme.palette.button.hover,
-    },
-    '&:active': {
-      boxShadow: 'none',
-      backgroundColor: theme.palette.button.action,
-      color: theme.palette.button.actionText,
-    },
-  },
-  connected: {
```
artursapek, line 42: you sure about that?

artursapek:
"connect wallet" CTA should be disabled when chain hasn't been chosen yet.

`development`:

image

this branch:

image

wormhole-connect/src/views/v2/Bridge/WalletConnector/index.tsx
emreboga: Yes.
Screenshot 2025-04-29 at 7 43 15 PM
Screenshot 2025-04-29 at 7 43 35 PM

artursapek (commented):
the default dark theme has a different button color now. is this intentional?

# `development`
image

## this branch
disabled:
image
enabled:
image

wormhole-connect/src/theme.ts
```diff
@@ -235,18 +235,38 @@ export const generateTheme = (customTheme: WormholeConnectTheme): Theme => {
...
+
+    if (primary.isDark()) {
+      primaryText = primary.lightness(95).hex();
+      disabled = primary.saturate(0.5).hex();
+      disabledText = primary.lightness(95).alpha(0.9).hexa();
+      action = primary.darken(0.15).hex();
+      actionText = primary.lightness(80).hex();
+      hover = primary.darken(0.1).hex();
+    } else {
+      primaryText = primary.lightness(35).hex();
```
artursapek: why did you go with 95 lightness for dark buttons, but only 35 lightness for light buttons? why not go lower than 35? it seems like if you're going very close to 100 lightness (white) for dark buttons, you should go pretty close to 0 (black) for light buttons.

if I choose a mid-range color the text is very low contrast.

image

or you could make it relative to the color's `l` value so it's a bit more adaptive:

```
      primaryText = primary.lightness(primary.l() / 3).hex();
```

image

image

emreboga: @artursapek It's setting the same lightness as before:
```
const buttonTextColor = hslToHex(h, s, l > 0.75 ? 0.35 : 0.95);
```

emreboga: I think the relative lightness might be a better option. I can give it a try.

Wesleyleung:
> the default dark theme has a different button color now. is this intentional?
> 
> # `development`
> image
> ## this branch
> disabled: image enabled: image

The latest branch screenshots (dark purple text in the buttons) looks a little different than what I saw previously, but yes, it is intentional to change the default button such that the text in the button is black (or at least as close to black as possible if we are deriving from HSL) vs white. The black color text was actually the original spec
438880537-5c7f942a-5ca4-4a63-b3dd-d317302b1a60

emreboga:
@Wesleyleung In the original changes I was using the color code directly for the text color. Now we are HSL generating it, that's why the difference, but it was a bit too much. I've decreased the lightness in the text color for dark mode and the text color is looking closer now.
Screenshot 2025-05-02 at 8 13 11 PM

wormhole-connect/src/theme.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3445/commits/4a314bd4fdf08461d0415fccbaa1bdeb90d5753e

emreboga: @artursapek I tried relative lightness but it wasn't easy to calibrate. I've decided to decrease the lightness instead when primary color is on the light end of the spectrum. We should revisit this as part of the overall theme/color discussion.

artursapek: ah ok thanks!

artursapek (approved):
good stuff

PR #3450, opened 2025-04-25
Executor route relay failure handling

kev1n-peters:
Refactored the `useExternalSearch` hook to use redux store instead of local state. The purpose of this is for the Redeem page to navigate the user to the resume transaction flow when a relay fails for an executor route. Automatic routes don't support manual claim yet, so this is a "hack" until we figure out a more robust solution.


emreboga (approved):
LGTM!

PR #3459, opened 2025-04-28
Add missing guard to Sui address check

artursapek:
This PR https://github.com/wormhole-foundation/wormhole-connect/pull/3444 broke arbitrary token loading for non-Sui chains.

### DIFF
--- a/wormhole-connect/src/config/tokens.ts
+++ b/wormhole-connect/src/config/tokens.ts
@@ -341,7 +341,10 @@ export class TokenCache extends TokenMapping<Token> {
   }
 
   async addFromTokenId(tokenId: TokenId): Promise<Token> {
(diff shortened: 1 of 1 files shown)


emreboga (approved):

tonyjin (approved):
Do we need this kind of chain check ahead of other `isValidXType()` methods? If so, should `isValidXType()` take the `tokenId` object as input and check the chain

artursapek:
> Do we need this kind of chain check ahead of other `isValidXType()` methods? If so, should `isValidXType()` take the `tokenId` object as input and check the chain

fair question. afaik other platforms already throw an error when we pass in an invalid address to their native address type, so I think in this case it may be unnecessary. but even if so, I'd want to address that in a separate PR so this one can just un-break the arbitrary tokens feature.

PR #3476, opened 2025-05-06
Bump wormhole sdk to 1.17.0

artursapek:


kev1n-peters (approved):

tonyjin (approved):
Noticed a bunch of the injective dependencies got bumped. Are we pinning all the dependencies we can, or were they upgraded as part of the Wormhole SDK bump?

artursapek:
@tonyjin those patch bumps were brought in by the SDK bump. I don't think we are pinning inside the SDK yet; only in Connect. we'd have to do that to avoid this kind of change happening.

PR #3482, opened 2025-05-07
Adding a new connect event: history.load

emreboga:
We need this new event to track users loading history view.

### DIFF
--- a/wormhole-connect/src/views/v2/Bridge/index.tsx
+++ b/wormhole-connect/src/views/v2/Bridge/index.tsx
@@ -421,9 +421,15 @@ const Bridge = () => {
         >
           <span>
             <IconButton
+              data-testid="history-button"
               sx={{ padding: 0 }}
               disabled={isTxHistoryDisabled}
-              onClick={() => dispatch(setAppRoute('history'))}
+              onClick={() => {
+                dispatch(setAppRoute('history'));
+                config.triggerEvent({
+                  type: 'history.load',
+                });
+              }}
             >
               <HistoryIcon />
             </IconButton>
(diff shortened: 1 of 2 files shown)


wormhole-connect/src/views/v2/Bridge/index.tsx
```diff
@@ -423,7 +423,12 @@ const Bridge = () => {
             <IconButton
               sx={{ padding: 0 }}
               disabled={isTxHistoryDisabled}
-              onClick={() => dispatch(setAppRoute('history'))}
+              onClick={() => {
+                dispatch(setAppRoute('history'));
+                config.triggerEvent({
+                  type: 'history.load',
```
emreboga, line 430: Couldn't think of any extra data to pass into this event :-/

artursapek: really? what about the chain...?

emreboga: Yes, because it captures the event when user clicks on the history button and I couldn't find anything relevant to that :-/ Maybe the wallet that we are pulling the history for, but thought we wouldn't want to log that 😅

artursapek: oh right because the history shows all chain activity, forgot

I actually think we could add the address - the event handler isn't necessary for logging. Portal happens to use it for logging certain data, but Connect doesn't log anything by itself. it's up to the integrator to determine what's appropriate.

emreboga: I agree, we can pass the address for the history event 👍

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3482/commits/f01a020f6417d8065c13140d57264980458dd514

artursapek (approved):

PR #3485, opened 2025-05-08
2.5.0 fixes

artursapek:
This fixes two bugs:

- Token list wasn't properly prioritizing destination tokens. We are now ignoring Mayan in this computation entirely (since they support everything) and only surfacing directly supported tokens by NTT, TB, as well as the native gas token & USDC. You can still paste an arbitrary address to get any token. 44f08b5cdd05719e433b39a97a072e644925e7d9 e7c01b1d2e0df65d9e41172cbe5d38a03f27f656

image
image

- When switching chains, it's possible to get the previous chain's token list after the new one. This was a race condition. Fixed by ea7fe11ec5e14889525642d4883d185c3ff4e0b9
image

### DIFF
--- a/wormhole-connect/src/config/tokens.ts
+++ b/wormhole-connect/src/config/tokens.ts
@@ -317,7 +317,7 @@ export class TokenCache extends TokenMapping<Token> {
   // Excludes wrapped tokens
   findBySymbol(chain: Chain, symbol: string): Token | undefined {
     let matching = this.getAllForChain(chain).filter(
-      (t) => t.symbol === symbol,
+      (t) => t.symbol.toLowerCase() === symbol.toLowerCase(),
     );
 
     if (matching.length > 1) {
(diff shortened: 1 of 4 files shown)


wormhole-connect/src/config/tokens.ts
```diff
@@ -317,7 +317,7 @@ export class TokenCache extends TokenMapping<Token> {
   // Excludes wrapped tokens
   findBySymbol(chain: Chain, symbol: string): Token | undefined {
     let matching = this.getAllForChain(chain).filter(
-      (t) => t.symbol === symbol,
```
artursapek, line 320: this enhances search

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -158,7 +136,22 @@ const TokenList = (props: Props) => {
       tokens.push(nativeToken);
     }
 
-    // Fourth: Add tokens with a balances in the connected wallet
+    const usdcAddr = circle.usdcContract.get(
```
emreboga, line 140: Didn't quite understood this part. We are always adding USDC if there is one in the chain after gas token?

emreboga (approved):
LGMT!

wormhole-connect/src/config/tokens.ts
DBosley: 👨‍🍳 💋

PR #3487, opened 2025-05-09
refactor token sorting

artursapek:
This completely changes how we sort tokens in `TokenList.tsx`, to use preference scores depending on the token's characteristics. It also improves token search.

### DIFF
--- a/wormhole-connect/src/routes/sdkv2/route.ts
+++ b/wormhole-connect/src/routes/sdkv2/route.ts
@@ -80,9 +80,9 @@ export class SDKv2Route {
 
     try {
       const supportedDestinationTokens = await this.supportedDestTokens(
-          sourceToken,
-          fromChain,
-          toChain,
+        sourceToken,
+        fromChain,
+        toChain,
       );
 
       return !!supportedDestinationTokens.find((tokenId) => {
@@ -108,6 +108,7 @@ export class SDKv2Route {
     const fromContext = await this.getV2ChainContext(fromChain);
     const toContext = await this.getV2ChainContext(toChain);
 
+    // TODO this is wrong... it should be filtering the output tokens...
     const isIlliquid = await this.isIlliquidDestToken(
       sourceToken,
       fromContext.context,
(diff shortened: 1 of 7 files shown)


wormhole-connect/src/views/v2/Bridge/index.tsx
```diff
@@ -308,19 +308,7 @@ const Bridge = () => {
           tokenList={sourceTokens}
           setChain={(value: Chain) => {
             selectFromChain(dispatch, value, sendingWallet);
-            // If there's a token on the new chain with the same symbol, select that
```
artursapek, line 311: this logic sucks and was causing the wrong token to be chosen when going between a chain with native gas NTT and another with just token bridge. it's not even necessary anymore to have this here.

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -141,7 +153,7 @@ const TokenList = (props: Props) => {
       config.network,
       props.selectedChainConfig.sdkName,
     );
-    if (usdcAddr) {
+    if (usdcAddr && !tokenSet.has(usdcAddr.toString())) {
```
artursapek: fixes dupe USDC oops

wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -125,6 +125,19 @@ const TokenList = (props: Props) => {
       tokens.push(searchResult);
     }
 
+    // Add tokens that routes explicitly support, like token bridge, NTT
+    if (props.tokenList) {
+      for (const token of props.tokenList) {
+        if (
+          !isFrankensteinToken(token, props.selectedChainConfig.sdkName) &&
```
artursapek: filtering frankensteins here now, since the SDKv2 Route wrapped only filters on the source token (?) I think it's a bug but this adds a layer of protection on the output tokens

wormhole-connect/src/routes/sdkv2/route.ts
```diff
@@ -108,6 +108,7 @@ export class SDKv2Route {
     const fromContext = await this.getV2ChainContext(fromChain);
     const toContext = await this.getV2ChainContext(toChain);
 
+    // TODO this is wrong... it should be filtering the output tokens...
```
artursapek, line 111: doesnt need to be investigated right now but I think this code is weird. would like to discuss with KP

emreboga (approved):

wormhole-connect/src/utils/AsyncCache.ts
```diff
@@ -31,20 +31,20 @@ export class AsyncCache<T> {
       console.debug('[Cache Debug] Cache MISS - initiating new request');
       pendingRequest = fetchFn()
         .then((result) => {
-          this.cache.set(cacheKey, {
```
artursapek, line 34: linter did this idfk

DBosley (approved):
Approved... But what's with that bundle size?

PR #3492, opened 2025-05-09
Hide tokens with 0 balances

artursapek:


wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -321,11 +328,11 @@ const TokenList = (props: Props) => {
         const queryLC = query.toLowerCase();
 
         const symbolMatch = [token.symbol, token.name].some((criteria) =>
-          criteria?.toLowerCase()?.includes?.(queryLC),
+          criteria?.toLowerCase()?.startsWith?.(queryLC),
         );
         if (symbolMatch) return true;
 
-        if (token.address.toString().toLowerCase().includes(queryLC)) {
+        if (token.address.toString().toLowerCase() === queryLC) {
```
tonyjin, line 335: Why `===` vs `includes()`?

artursapek: because if we do an `includes` it matches on substrings within the contract address, and nobody searches for tokens that way; people paste full addresses.

this was causing it to show random tokens on Solana when I was typing in short 2-3 character queries because their CAs happened to contain those 3 characters in that order.

kev1n-peters (approved):

PR #3493, opened 2025-05-11
Speed up token cache initalization by order of magnitude

artursapek:
The token cache is very inefficient. It makes lots of wasteful calls to crypto libraries like `keccak256` because it parses & re-stringifies token addresses a ton (since it stores things as strings in localStorage). This adds a new class `TokenIdLazy` which stores the string version of a token (which is always how they originate) and only parses it when needed. We never actually need the parsed address until we start making SDK calls to initiate transfers - until then we can use this cached `addressString` property of `TokenIdLazy` instead.

Yes, this makes the code more convoluted, but it speeds up the cache initialization a lot and removes a huge chunk of the initial blocking time when Connect first loads and is given its config.

Currently it takes up ~70-80ms total of the "total blocking time" on Portal bridge because we run it multiple times (each time we update the config):

image

image

This means loading the token cache takes up about 25% of the total blocking time, which is the worst part of Portal's performance score right now. (Just using Portal as an example of a Connect integration).

After these changes, each invocation takes 0.5-2ms, so even if we run it 2-3 times this is only adding up to ~6ms of blocking time vs 60-80ms.

image

Ultimately this will make it so Connect initializes faster and becomes interactive sooner.

### DIFF
--- a/wormhole-connect/src/config/tokens.ts
+++ b/wormhole-connect/src/config/tokens.ts
@@ -7,7 +7,6 @@ import {
   TokenAddress,
   toNative,
   isNative,
-  isSameToken,
   Network,
   chainToPlatform,
   UniversalAddress,
@@ -124,7 +123,7 @@ export class Token extends TokenIdLazy {
   }
 
   equals(other: Token): boolean {
-    return isSameToken(this.tokenId, other.tokenId);
+    return isSameToken(this, other);
   }
 
   toJson(): TokenJson {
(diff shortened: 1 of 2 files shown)


DBosley:
There are a bunch of instances across connect that call `address.toString()` that might still be slow. Should we also change those?

artursapek:
> There are a bunch of instances across connect that call `address.toString()` that might still be slow. Should we also change those?

they could be. but they're not being called hundreds of times in a row, so they don't matter as much.

DBosley (approved):

PR #3494, opened 2025-05-11
Experiment: Hide CCTP manual when automatic is present

emreboga:
Adding a new experimental option to hide manual CCTP when automatic is present in available routes.

### DIFF
--- a/wormhole-connect/src/config/types.ts
+++ b/wormhole-connect/src/config/types.ts
@@ -126,6 +126,9 @@ export interface WormholeConnectConfig {
   // isTokenSupportedHandler can be used to disable certain tokens from being selected
   isTokenSupportedHandler?: IsTokenSupportedHandler;
 
+  // filterRoutes can be used to filter the routes that are shown to the user
+  filterRoutes?: (routes: string[]) => string[];
+
   // UI details
   ui?: UiConfig;
 
--- a/wormhole-connect/src/config/types.ts
+++ b/wormhole-connect/src/config/types.ts
@@ -166,6 +166,7 @@ export interface InternalConfig<N extends Network> {
   validateTransfer?: ValidateTransferHandler;
   isRouteSupportedHandler?: IsRouteSupportedHandler;
   isTokenSupportedHandler?: IsTokenSupportedHandler;
+  filterRoutes?: (routes: string[]) => string[];
 
   // UI configuration
   ui: UiConfig;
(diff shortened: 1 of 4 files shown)


wormhole-connect/src/hooks/useFetchSupportedRoutes.ts
```diff
@@ -103,6 +103,14 @@ const useFetchSupportedRoutes = ({
         );
       }
 
+      // Experimental - Hide CCTP Manual route when CCTP automatic is present
+      if (
+        config.ui.experimental?.hideCCTPManWhenAutoPresent &&
+        _routes.includes('AutomaticCCTP')
```
kev1n-peters: Do we also need to check for the CCTPW7ExecutorRoute? Another possibility is exposing a callback through the WormholeConnectConfig that allows integrators to filter routes here (so the filtering can be done from the Portal side)

emreboga: Not this experiment but in the next one we'll check executor route too. I thought the same tbh, but didn't want to go that direction for two reasons:
1. It'd complicate things when it's not a simple include/exclude filter. For instance when we are doing the executor vs. CCTP we can use that too, but at some point we'll start including fees for executor experiments and that'll complicate things.
2. I'd like to be more conservative when exposing Connect internals via callbacks. Passing only route names to this callback won't seem to be enough for our experiments and I'm hesitant to give more than that to all integrators :)

emreboga: After thinking more on this, I think providing a callback to filter routes, even with just route names and no complex logic (like fees etc.) would be beneficial. I'll add that to this PR.

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3494/commits/702c4a4850f33b5eb039aa04e00a404ea4c6130d

wormhole-connect/src/config/types.ts
```diff
@@ -166,6 +166,7 @@ export interface InternalConfig<N extends Network> {
   validateTransfer?: ValidateTransferHandler;
   isRouteSupportedHandler?: IsRouteSupportedHandler;
   isTokenSupportedHandler?: IsTokenSupportedHandler;
+  filterRoutes?: (routes: string[]) => string[];
```
kev1n-peters, line 172: We also need to set this in `WormholeConnectConfig` and copy it to the internal config (search the codebase for `isTokenSupportedHandler` to see where we should do this)

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3494/commits/f829b23e1280f7061e33479729985db19fa886f9

kev1n-peters (approved):

PR #3496, opened 2025-05-12
[PROD-158] USDC tx explorer link for ExecutorRoute

emreboga:
- Added a special case to render USDC.range explorer link for ExecutorRoutes instead of WHScan. This is a temp situation that will get removed once WHScan adds the support for ExecutorRoute transactions.
- I've also added a utility isExecutorRoute. This is super simple route name check, but used in several places and it's good to have the logic in one place.

### DIFF
--- a/wormhole-connect/src/utils/sdkv2.ts
+++ b/wormhole-connect/src/utils/sdkv2.ts
@@ -24,6 +24,7 @@ import { PublicKey } from '@solana/web3.js';
 import * as splToken from '@solana/spl-token';
 import { WORMSCAN } from 'config/constants';
 import { TokenTuple } from 'config/tokens';
+import { isExecutorRoute } from 'utils';
 
 // Used to represent an initiated transfer. Primarily for the Redeem view.
 export interface TransferInfo {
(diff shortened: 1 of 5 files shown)


wormhole-connect/src/utils/sdkv2.ts
```diff
@@ -76,15 +76,19 @@ export function getExplorerInfo(
     return {
       url: `https://explorer.mayan.finance/swap/${txHash}`,
       name: 'Mayan Explorer',
-      apiUrl: `${config.mayanApi}/v3/swap/trx/${txHash}`,
+    };
+  } else if (isExecutorRoute(routeName)) {
+    // TODO Remove once Wormholescan explorer supports Executor routes
+    return {
+      url: `https://usdc.range.org/usdc/status/${txHash}`,
```
artursapek: this website says they don't support testnet, but CCTP Executor route does. do you want to handle that here?

above, we don't handle the testnet case for Mayan but the Mayan route also doesn't work on testnet so it's not an issue.

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3496/commits/e6636a465dc3662c61e885b1eafc4a44a6bf6c7a

artursapek (approved):

PR #3509, opened 2025-05-14
Clear recipient wallet when manual wallet address is invalid

kev1n-peters:


wormhole-connect/src/views/v2/Bridge/WalletConnector/Controller.tsx
```diff
@@ -113,61 +113,61 @@ const ConnectedWallet = (props: Props) => {
...
-        ])}
-        {...popupTrigger}
-      >
-        <WalletIcons name={wallet.name} icon={wallet.icon} size={20} />
-        <Tooltip title="Copied" open={isCopied} placement="top" arrow>
-          <Typography
-            className={classes.walletAddress}
-            fontSize={14}
-            fontWeight={700}
+      {!wallet?.address ? null : (
```
kev1n-peters, line 118: This change just makes it so that the wallet sidebar remains open after the user submits an invalid manual address so they can see the error message.

artursapek (approved):

PR #3510, opened 2025-05-14
Clear destination tokens before fetching; add loading skeleton UI

artursapek:
This is a combo bugfix and UI improvement.

- clears destination tokens before fetching them, so we're not showing the old chain's tokens in the meantime
- replaces the loading spinner with a more subtle skeleton (like we do when loading routes)

https://github.com/user-attachments/assets/b5b77cd0-1347-4487-b6c0-994af725e661


wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -308,11 +309,14 @@ const TokenList = (props: Props) => {
         )
       }
       loading={
-        props.isFetching && (
+        props.isFetching &&
+        [1, 2, 3].map((_) => (
```
emreboga, line 313: Nitpick: Before we had a single loading indicator. Did you try a single skeleton item when loading? When there are less than 3 tokens after loading, showing 3 skeletons then 1 token doesn't seem ideal.

emreboga (approved):

PR #3514, opened 2025-05-14
remove forced height on skeleton tokens

artursapek:


wormhole-connect/src/views/v2/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -312,7 +312,7 @@ const TokenList = (props: Props) => {
         props.isFetching &&
         [1, 2, 3].map((_) => (
           <ListItemButton className={classes.tokenLoader} dense>
```
kev1n-peters, line 314: We should provide a `key` prop so React doesn't complain in the console

artursapek: https://github.com/wormhole-foundation/wormhole-connect/pull/3514/commits/fcf4ef58852e244bf4be5c9809a0efcec6b01145 <3

kev1n-peters (approved):

PR #3516, opened 2025-05-15
Add parsing for SOL insufficient lamports error message

emreboga:
This will parse messages like `Transfer: insufficient lamports 944448, need 2923200`
We will parse the current and required amounts to inform the users.

Screenshot 2025-05-15 at 1 47 50 PM


artursapek (approved):
🫡

PR #3519, opened 2025-05-16
Bump wallet aggregator, Node, NPM, Aptos

artursapek:
- Adds Petra wallet for Aptos
- Updates aptos dependencies
- Update NPM to 11, Node to 22. there's a [bullshit NPM bug](https://stackoverflow.com/questions/73356208/npm-err-cannot-read-properties-of-null-reading-children) that we have to update in order to fix, if we want to use these latest versions of stuff
- Fix issue with EVM wallet `rpcUrls`


kev1n-peters (approved):

kev1n-peters:
Code seems fine, but we should probably have QA test some wallets to make sure something under the hood in the wallet aggregator packages didn't break us

PR #3525, opened 2025-05-20
Remove redundant chain configs, add Sonic & Linea

artursapek:
Simplified the chain configs Connect requires:

- removed entire legacy chain configs from `sdklegacy`
- removed `Context` from `sdklegacy`, replaced with `Platform` from new SDK
- removed new chain config `chainId` (redundant with consts in the SDK)
- removed new chain config `wrappedGasToken` (can rely on `Token` class for this now)

Adding a new chain no longer requires things that are redundant with the SDK, only icons, symbol, and explorer URL.

Then also:

- added `Sonic` and `Linea` to new chain configs


wormhole-connect/src/components/Header.tsx
```diff
@@ -4,19 +4,17 @@ import { makeStyles } from 'tss-react/mui';
 export type Alignment = 'center' | 'left' | 'right';
 
 type StyleProps = { align: Alignment; fontSize: number };
-const useStyles = makeStyles<StyleProps>()(
```
artursapek, line 7: I have no idea why but this code was reformatted and I can't revert it because git insists the commit would be empty (because of the precommit hook)

kev1n-peters (approved):

PR #3549, opened 2025-05-27
Remove the tx link for USDC.range to default to WHScan

emreboga:
We no longer need to render USDC.range links as WHScan is now supporting executor transactions.


artursapek (approved):
wow, a TODO comment actually being done. it's like spotting an extremely rare bird

PR #3555, opened 2025-05-27
Reset gas top-up when route is unselected

emreboga:


kev1n-peters (approved):

kev1n-peters:
Do we even need to debounce anymore since the slider was removed in favor of discrete percentages?

PR #3561, opened 2025-05-29
Fix/3448: Display "No routes found" message after wallet disconnect

hashpalk:
When the wallet was disconnected, the `useAmountValidation` hook in the Bridge view was being disabled because the `disabled` prop included a check for `!sendingWallet.address`. This prevented the hook from evaluating the route finding status and displaying the "No routes found" message when applicable.

This change removes the `!sendingWallet.address` check from the `disabled` prop, allowing the validation to run and the "No routes found" message to be displayed correctly even when the wallet is disconnected.


artursapek (approved):
thanks for the patch!

hashpalk:
Hi @artursapek can you please merge the pull request, as it was approved? Thanks!

hashpalk:
Hi @artursapek could you provide any feedback on how to fix the build failure? Any advice would be appreciated.

PR #3598, opened 2025-06-12
Added NTTExecutorRoute support

kev1n-peters:


src/exports/ntt.ts
```diff
@@ -6,6 +6,8 @@ import '@wormhole-foundation/sdk-solana-ntt';
 
 import {
   nttAutomaticRoute,
+  NttExecutorRoute,
```
emreboga: Nitpick: Can you add a `type` for this one?

kev1n-peters: done

emreboga (approved):

PR #3603, opened 2025-06-13
Remove redundant search query state

artursapek:
Right now, there are two state variables for the AssetPicker search query. This is causing bugs where we have phantom search query state that the UI isn't showing.

This change removes the state var in `SearchableList` and makes it a controlled input via prop. Changes apply to both `AssetPicker` and `ChainList` (both of them use `SearchableList`).


emreboga (approved):
LGTM!

PR #3604, opened 2025-06-13
UI Refresh v3: Main view, asset pickers and routes

emreboga:
This PR introduces a complete UI redesign with new v3 components for asset pickers and routes with mobile web improvements as well. We've also added performance optimizations, and a gradual migration strategy through feature flags. You can see the details in [this loom](https://www.loom.com/share/94009ee5be624278aa5296b93229d2fb?sid=0dec58c6-6396-4508-aa8b-c3b0cc86e1ec).

IMPORTANT: Please note that we added a separate v3 folders for the new design components. Especially for `views/v3` folder the majority of the code is the same with views/v2. The reasoning behind this is due to a an easier and cleaner roll out. Instead of adding feature gate checks for the new designs in so many places, we added a single one to turn it on and off ruing roll out.

For our release strategy, we'll be following these steps.
1. Deploy with feature gate disabled (current v2 behavior).
2. Roll out to 10% for initial feedback.
3. Gradual rollout while collecting more feedback.
4. Eventually deprecate v2 components.

**🎯 Review Focus Areas**

These files contain substantial logic/UI changes and need thorough review:
- src/views/v3/Bridge/index.tsx
- src/views/v3/Bridge/AmountInput/index.tsx
- src/views/v3/Bridge/AssetPicker/index.tsx
- src/views/v3/Bridge/Routes/index.tsx
- src/views/v3/Bridge/SwapInputs/index.tsx

These files only have import statement changes to use v3 components:
- src/views/v3/Bridge/AssetPicker/ChainList.tsx
- src/views/v3/Bridge/AssetPicker/TokenList.tsx
- src/views/v3/Bridge/GasSlider/index.tsx
- src/views/v3/Bridge/Routes/SingleRoute.tsx
- src/views/v3/Bridge/WalletConnector/*.tsx (all files)
- src/views/v3/Redeem/*.tsx (all files)
- src/views/v3/TxHistory/*.tsx (all files)
- src/components/v3/AlertBanner.tsx
- src/components/v3/Button.tsx


src/components/v3/AlertBanner.tsx
```diff
@@ -0,0 +1,98 @@
...
+  content: ReactNode;
+  warning?: boolean;
+  error?: boolean;
+  testId?: string;
+  color?: string;
+  sx?: SxProps;
+  className?: string;
+}
+
+/**
```
priyajeet: I think we can safely remove these docs since they don't really provide anything valuable and generally bit rot as people rarely update it as component changes. The name of the component / file name is self documenting anyways.

src/components/v3/AlertBanner.tsx
```diff
@@ -0,0 +1,98 @@
...
+ * AlertBanner component for displaying contextual messages with appropriate styling
+ *
+ * @param props - The props for the AlertBanner component
+ * @returns A styled alert banner with optional warning/error styling
+ */
+const AlertBanner: React.FC<AlertBannerProps> = ({
```
priyajeet: We should make all the functional component syntax consistent everywhere. From a readability perspective, we should just make this:

```tsx
function AlertBanner({ ... }: AlertBannerProps) {
    ...
}
```

More concise + plays better with TS generics out of the box without need to use React.stuff. And its really a function so best to denote it as a function.

src/components/v3/AlertBanner.tsx
```diff
@@ -0,0 +1,98 @@
...
+  testId,
+  color,
+  sx,
+  className,
+}) => {
+  const theme = useTheme();
+
+  // Determine the appropriate color based on props
+  const alertColor = useMemo((): string | undefined => {
+    if (color) return color;
```
priyajeet: For readability / consistency and most important debug point-ability, we should always put { } block statements. I think there might be an eslint/prettier rule to enforce / auto-fix this.

src/components/v3/AlertBanner.tsx
```diff
@@ -0,0 +1,98 @@
...
+  }, [
+    color,
+    error,
+    warning,
+    theme.palette.warning.main,
+    theme.palette.error.main,
+  ]);
+
+  // Early return if content is not provided or empty
+  if (!content) {
```
priyajeet: Should ideally be using children for the content. And showing / hiding behavior of a banner should be controlled by the parent and not the banner itself. Big reason for this is that `if ()` blocks can only happen in the render function after all hook calls. And hook calls can be expensive or unnecessary if there is nothing to render. So whenever possible, we should let the parent control the visibility. Additional to that there is a `show` prop _(should that be called collapsed?)_ that makes it ambiguous, since technically show=true and content=null doesn't show the alert.

So for an alert banner, the parent can just be

```
const hasError = !!error?.content;

{ hasError && (
    <AlertBanner>
        {error.content}
    </AlertBanner>
)}
```

src/components/v3/AlertBanner.tsx
```diff
@@ -0,0 +1,98 @@
...
+  error = false,
+  testId,
+  color,
+  sx,
+  className,
+}) => {
+  const theme = useTheme();
+
+  // Determine the appropriate color based on props
+  const alertColor = useMemo((): string | undefined => {
```
priyajeet: Don't think you are going to get any perf advantage memoizing a non complex operation like this. useMemo has its [own overhead](https://kentcdodds.com/blog/usememo-and-usecallback) that negates the advantages. You can also just move this out to an external utility function if needed.

src/components/v3/Button.tsx
```diff
@@ -0,0 +1,110 @@
...
+        VARIANT_STYLES[variant as keyof typeof VARIANT_STYLES];
+
+      // Type assertion to help TypeScript understand the filtered array is still valid SxProps
+      return [BASE_BUTTON_STYLES, variantStyles, sx].filter(
+        Boolean,
+      ) as SxProps<Theme>;
+    }, [variant, sx]);
+
+    // Determine MUI variant based on custom variant
+    const muiVariant = useMemo((): MUIButtonProps['variant'] => {
```
priyajeet: Same as before, this will guaranteed to be a lower performance than just 
`const muiVariant = variant === 'primary' || variant === 'error' ? 'contained' : 'text'`
Also in this case a lot easier to read that over a memo call.

src/components/v3/Button.tsx
```diff
@@ -0,0 +1,110 @@
...
+ *   - 'primary': Main CTA button with primary colors
+ *   - 'error': Error state button with error colors
+ *   - 'default': Standard MUI button (fallback)
+ * @param sx - Additional style overrides
+ * @param rest - All other MUI Button props
+ * @param ref - Forwarded ref to the button element
+ *
+ * @returns A styled button component
+ */
+const Button = forwardRef<HTMLButtonElement, CustomButtonProps>(
```
priyajeet: forwardRef I think has been deprecated and also makes TS typing harder and less readable. If you switch to a simple function, just [pass in ref as any other prop](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4Aoc4AOxiSk3STgCEBXGGCagBRzADOcAN7k4cNAAtgAGwAmRagH4AXImIwAdMnQwAchDlIK4mAE8wSNfgFsARiGAx8cAD4E7HLtTJi4RTDUdDG0kTAB5OwArJAwAHgAJABUAWQAZdk5uAFEZJBAkWgA+cgBfSkw2agxgblYvbgAKYThzSwAaCWl5RU6AuFK1TO8+CEEAShE-Ihg2KGo4Rr9xOM8shYCAXmEA0taLJG22pFKS8XORKVkFQvKLuIB6Ne8zuHGKcvIkAA9IWDgjIw2DJ4MNuKQgA). Pre react-19, may have to just do `export default forwardRef(Button)` I think and tweak some types.

The other thing that _**might**_ get simpler with that is test for this component if you just export the non-ref-forwarded component at the bottom along with the forwarded one, since the latter has no real bearing on tests. So something like

```tsx
export { Button };
export default forwardRef(Button);
```

src/components/v3/Button.tsx
```diff
@@ -0,0 +1,110 @@
...
+  ({ variant = 'default', sx, ...rest }, ref: Ref<HTMLButtonElement>) => {
+    const computedStyles = useMemo((): SxProps<Theme> => {
+      // Handle default variant separately
+      if (variant === 'default') {
+        return sx ?? {};
+      }
+
+      // Get variant styles (we know it exists for 'primary' and 'error')
+      const variantStyles =
```
priyajeet: Probably later, you can investigate [cva](https://cva.style/docs/getting-started/variants) to see if that helper can help with this stuff.

src/components/v3/Button.tsx
```diff
@@ -0,0 +1,110 @@
...
+} as const;
+
+/**
+ * Custom Button component that extends MUI Button with predefined variants
+ *
+ * @param variant - The style variant of the button:
+ *   - 'primary': Main CTA button with primary colors
+ *   - 'error': Error state button with error colors
+ *   - 'default': Standard MUI button (fallback)
+ * @param sx - Additional style overrides
```
priyajeet: If props like this are renamed to say `styleOverrides`, then things start to self document. Otherwise people are forced to read and maintain docs.

src/components/v3/Button.tsx
```diff
@@ -0,0 +1,110 @@
...
+  type Theme,
+} from '@mui/material';
+
+type ButtonVariant = 'primary' | 'error' | 'default';
+
+interface CustomButtonProps extends Omit<MUIButtonProps, 'variant'> {
+  readonly variant?: ButtonVariant;
+}
+
+const BASE_BUTTON_STYLES: SxProps<Theme> = {
```
priyajeet, line 16: Would it make this file simpler (to parse/read/diff) if these are moved out to a sibling file `styles.ts`? A pattern that can be then encouraged to other components?

src/components/v3/AlertBanner.tsx
```diff
@@ -0,0 +1,98 @@
...
+  type SxProps,
+} from '@mui/material';
+import ReportProblemOutlinedIcon from '@mui/icons-material/ReportProblemOutlined';
+
+interface AlertBannerProps {
+  show: boolean;
+  content: ReactNode;
+  warning?: boolean;
+  error?: boolean;
+  testId?: string;
```
priyajeet, line 9: Weird to add props that are only for testing and not usability. Instead we add a role like https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role (unless MUI is already adding?)

src/icons/Clock.tsx
```diff
@@ -0,0 +1,20 @@
...
+  >
+    <path
+      d="M6 0C4.81331 0 3.65327 0.351894 2.66658 1.01118C1.67989 1.67047 0.910851 2.60754 0.456725 3.7039C0.00259973 4.80026 -0.11622 6.00665 0.115291 7.17054C0.346802 8.33443 0.918247 9.40352 1.75736 10.2426C2.59648 11.0818 3.66557 11.6532 4.82946 11.8847C5.99335 12.1162 7.19974 11.9974 8.2961 11.5433C9.39246 11.0891 10.3295 10.3201 10.9888 9.33342C11.6481 8.34673 12 7.18669 12 6C12 4.4087 11.3679 2.88258 10.2426 1.75736C9.11742 0.632141 7.5913 0 6 0ZM8.37805 6.87818H5.88071C5.69132 6.87818 5.50969 6.80295 5.37577 6.66903C5.24185 6.53512 5.16662 6.35349 5.16662 6.1641V2.32764C5.16662 2.13825 5.24185 1.95662 5.37577 1.8227C5.50969 1.68879 5.69132 1.61355 5.88071 1.61355C6.07009 1.61355 6.25172 1.68879 6.38564 1.8227C6.51956 1.95662 6.59479 2.13825 6.59479 2.32764V5.45001H8.37805C8.56743 5.45001 8.74906 5.52525 8.88298 5.65916C9.0169 5.79308 9.09213 5.97471 9.09213 6.1641C9.09213 6.35349 9.0169 6.53512 8.88298 6.66903C8.74906 6.80295 8.56743 6.87818 8.37805 6.87818Z"
+      fill="currentColor"
+    />
+  </svg>,
+  'Clock',
+);
+
+export default ClockIcon;
```
priyajeet, line 20: This is something I would likely memoize, depending upon the internal implementation of createSvgIcon

src/views/v3/Bridge/AmountInput/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+import { amount as sdkAmount } from '@wormhole-foundation/sdk';
+import Box from '@mui/material/Box';
+
+import { Token } from 'config/tokens';
+import type { RootState } from 'store';
+import { useGetTokens } from 'hooks/useGetTokens';
+
+const INPUT_DEBOUNCE = 500;
+
+const DebouncedTextField = memo(
```
priyajeet, line 25: This should ideally be its own file. 1 component/class per file with 1 main export.

src/views/v3/Bridge/AmountInput/index.tsx
```diff
@@ -0,0 +1,192 @@
+import React, {
```
priyajeet, line 1: Ideally, we should name files based on their component name. index.ts/tsx should not have any logic in it other than indexing purposes for like `export * from ...` or `export { default } from ...`. Because then its easy in VSCode and other editors to open files by name based on component name and we dont have a gazillion index.tsx which on itself doesn't tell what it holds unless you parse its path.

src/views/v3/Bridge/AmountInput/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+  ChangeEventHandler,
+  ComponentProps,
+  memo,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import { useSelector } from 'react-redux';
```
priyajeet, line 10: I think in the future we should probably nuke redux and rely on context, not sure of its usefulness in a super small form app. Even if the app had 10x more form fields, redux is still not useful. Instead something like react-hook-form might be better suited which takes care of almost all of state management of the form.

src/views/v3/Bridge/AmountInput/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  const {
+    fromChain: sourceChain,
+    toChain: destChain,
+    isTransactionInProgress,
+  } = useSelector((state: RootState) => state.transferInput);
+
+  const { sourceToken, destToken } = useGetTokens();
+
+  const isInputDisabled = useMemo(
```
priyajeet: Same as before, unnecessary useMemo. This in particular will add more of a performance penalty due to a function call over a simple boolean ||ing statement.

src/views/v3/Bridge/AmountInput/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+            onChange={props.onChange}
+            onDebouncedChange={props.onDebouncedChange}
+          />
+        </CardContent>
+      </Card>
+    </Box>
+  );
+};
+
+export default React.memo(AmountInput);
```
priyajeet: We are using React.xxx in some places and xxx directly, I think we should decide on what we want to do consistently

src/views/v3/Bridge/AmountInput/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+        setInnerValue(value);
+      }
+      // We should run this sife-effect only when the value changes
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [value]);
+
+    return (
+      <TextField
+        {...props}
+        data-testid="amount-input"
```
priyajeet, line 75: Can just give this component an aria-label, though shouldnt this TextField have a InputLabel or FormLabel component surrounding it?

src/views/v3/Bridge/AmountInput/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+  return (
+    <Box sx={styles.amountContainer}>
+      <Card sx={styles.amountInput}>
+        <CardContent sx={styles.amountCardContent}>
+          <DebouncedTextField
+            fullWidth
+            disabled={isInputDisabled}
+            placeholder="0"
+            slotProps={{
+              htmlInput: {
```
priyajeet: I think externalizing this object might make the markup more parsable visually

src/views/v3/Bridge/AmountInput/index.tsx
priyajeet: Also it seems some components are inside index.tsx while others in their respective file name, so moving to latter will also improve consistency

src/views/v3/Bridge/AmountInput/index.tsx
priyajeet: Also, btw, a component like this may not be needed when using react-hook-form, since the debouncing logic will be at the data layer and not component/ui level - which helps the UI to be more responsive. Maybe also look at [useDeferredValue()](https://react.dev/reference/react/useDeferredValue) meanwhile.

src/views/v3/Bridge/AssetPicker/ChainList.tsx
```diff
@@ -0,0 +1,231 @@
...
+    onChainSelect,
+  } = props;
+
+  const [topChains, showMoreButton] = useMemo(() => {
+    const allChains = chainList ?? [];
+    const selectedChain = selectedChainConfig;
+
+    // Find the selected chain in supported chains
+    const selectedChainIndex = allChains.findIndex((chain) => {
+      return chain.sdkName === selectedChain?.sdkName;
```
priyajeet: if selectedChain is optional, this loop still runs. The if () logic from below moved up will help prevent loop run.

src/views/v3/Bridge/AssetPicker/ChainList.tsx
```diff
@@ -0,0 +1,231 @@
...
+    chainList,
+    selectedChainConfig,
+    showSearch,
+    setShowSearch,
+    onChainSelect,
+  } = props;
+
+  const [topChains, showMoreButton] = useMemo(() => {
+    const allChains = chainList ?? [];
+    const selectedChain = selectedChainConfig;
```
priyajeet: Unnecessary reassignment?

src/views/v3/Bridge/AssetPicker/ChainList.tsx
```diff
@@ -0,0 +1,231 @@
...
+  }, [
+    styles.chainButton,
+    onChainSelect,
+    selectedChainConfig?.sdkName,
+    setShowSearch,
+    showMoreButton,
+    topChains,
+  ]);
+
+  const searchList = useMemo(
```
priyajeet, line 168: same as before, can be its own file

src/views/v3/Bridge/AssetPicker/ChainList.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  const {
+    chainList,
+    selectedChainConfig,
+    showSearch,
+    setShowSearch,
+    onChainSelect,
+  } = props;
+
+  const [topChains, showMoreButton] = useMemo(() => {
```
priyajeet: showMoreButton seems = allChains.length > SHORT_LIST_SIZE, doesn't need to be memoized

src/views/v3/Bridge/AssetPicker/ChainList.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  const searchList = useMemo(
+    () => (
+      <SearchableList<ChainConfig>
+        searchPlaceholder="Search for a chain"
+        sx={styles.chainSearch}
+        items={chainList ?? []}
+        searchQuery={chainSearchQuery}
+        onQueryChange={setChainSearchQuery}
+        filterFn={(chain, query) =>
```
priyajeet, line 176: Here and other places, should externalize and memoize (useCallback) as needed, easier to put breakpoints in non-JSX code + JSX remains concise

src/components/v3/AlertBanner.tsx
emreboga: https://linear.app/wormholelabs/issue/PROD-236/use-roles-instead-of-test-id-in-ui-automation

src/views/v3/Bridge/AssetPicker/index.tsx
```diff
@@ -0,0 +1,557 @@
...
+      <Box sx={{ display: 'flex', gap: '6px' }}>
+        {renderPercentButton(25)}
+        {renderPercentButton(50)}
+        {renderPercentButton(100)}
+      </Box>
+    );
+  }, [props.wallet.address, tokenBalance, renderPercentButton]);
+
+  return (
+    <Box sx={styles.root}>
```
priyajeet, line 351: So I think this component can definitely be broken down quite a bit. If markup is going beyond 100 lines of code, I think it deserves to be extracted out so that when reading HTML one can mentally figure out how the structure is visually. And its lot easier to review for future changes. At the moment one has to scroll a lot just to understand what the structure is specially hard to figure out where ? : start and end.

src/views/v3/Bridge/AssetPicker/index.tsx
```diff
@@ -0,0 +1,557 @@
...
+              alignItems: 'center',
+              gap: '8px',
+            }}
+          >
+            <Box>{balance}</Box>
+            <Box>{percentButtons}</Box>
+          </Box>
+        )}
+      </Box>
+      {mobile ? (
```
priyajeet, line 466: A good example here for breakdown would be that you have 2 components.

Using arbitrary names, `<MobileAssetPicker>` and `<DesktopAssetPicker>` and this code in this file would just be `mobile ? <MobileAssetPicker> : <DesktopAssetPicker>`

src/views/v3/Bridge/AssetPicker/index.tsx
priyajeet:
```jsx
<Box>
    <AssetPickerHeader>
    <AssetPickerContent>
</Box>
```

src/components/v3/Button.tsx
emreboga: IMO unless the styles are shared among multiple components, it's easier to read and debug if the styles of a component lives in the same file.

src/views/v3/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -0,0 +1,276 @@
...
+              <Skeleton variant="circular" width="36px" height="36px" />
+            </Box>
+          </ListItemButton>
+        ))
+      }
+      items={sortedTokens}
+      onQueryChange={(query) => {
+        props.onSearchQueryChange(query);
+      }}
+      filterFn={(token, query) => {
```
priyajeet, line 204: Functions like these don't need to be in react component as they have not much to do with the UI but data parsing. If this is moved out to a simple plain old JS function and imported from another file, it will be [1] testable on its own [2] reduce the component here where the person reading this all cares about filterFn={someFn} and not what someFn implementation is and [3] Will allow underlying component to shallow compare and potentially prevent a re-render because the args don't change each time - imported functions are singletons

src/icons/Clock.tsx
emreboga: It memoizes the wrapper component itself:
https://github.com/mui/material-ui/blob/90546216d1448ab3dc8b2d3538cf7cfcbeb53aa0/packages/mui-material/src/utils/createSvgIcon.js#L29

src/views/v3/Bridge/AmountInput/index.tsx
emreboga: https://linear.app/wormholelabs/issue/PROD-238/better-alternatives-to-replace-redux

src/views/v3/Bridge/AmountInput/index.tsx
emreboga: I agree with the 1 component/class per file approach but the `DebouncedTextField` component was very specific to AmountInput so left it in the same file here, but don't mind moving out too.
The goal was not to fetch quotes until user stops typing, but `useDeferredValue` may also work here.

src/views/v3/Bridge/AmountInput/index.tsx
emreboga: MUI TextField las a label prop for that, but that's not the UX we wanted to render for Amount input. I can add the aria-label.

src/views/v3/Bridge/AssetPicker/ChainList.tsx
emreboga: There will always be a selected chain as we select one (Ethereum) by default and user cannot clear it after opening the AssetPicker. The reason for the `?.` is due to the selectedChain type can be undefined before we launch the ChainList. It's a bit unnecessary, but we won't get into that loop without a selectedChain.

src/views/v3/Bridge/AssetPicker/ChainList.tsx
emreboga: This was built by an external dev about a year ago and we need to refactor SearchableList component all together for better reusability by Token and Chain lists.
https://linear.app/wormholelabs/issue/PROD-239/refactorrewrite-searchablelist-component

src/components/v3/AlertBanner.tsx
emreboga: I agree, no need for useMemo here, but can stay in the same file for now.

src/components/v3/AlertBanner.tsx
emreboga: https://linear.app/wormholelabs/issue/PROD-240/re-eval-linterprettier-rules

src/components/v3/AlertBanner.tsx
emreboga: Yup, AlertBanner needs a rewrite and will address this feedback as well.

src/components/v3/Button.tsx
emreboga: Will rename sx prop to styleOverrides.

src/components/v3/Button.tsx
emreboga: Oh totally missed forwardRef being deprecated. I will look into this 👍

src/components/v3/Button.tsx
emreboga: Button is the only component we have custom variants (other than the builtin ones in MUI components). It's probably better to revisit this once we have a sizable component lib.

src/AppRouter.tsx
```diff
@@ -55,6 +62,15 @@ const AppRouterContent: React.FC = () => {
     }
   }, [hasExternalSearch, dispatch]);
 
+  // TODO: Deprecate with UI refresh v3
+  const bridgeView = useMemo(() => {
+    return config.ui.experimental?.enableUIRefreshV3 ? (
```
priyajeet: Should we be using the getExperiment() function from experiments.ts?

emreboga: Good catch! I added that one when introduced experimental features and didn't used it here 😅  Will fix.

src/components/v3/Button.tsx
emreboga: Turns out we don't need the forwarded ref in parent component of Button. We can remove this all together.

src/views/v3/Bridge/AmountInput/index.tsx
emreboga: Let's track this as a separate tech-debt:
https://linear.app/wormholelabs/issue/PROD-246/check-if-we-can-replace-debauncedtextfield-with-usedeferredvalue

src/views/v3/Bridge/AssetPicker/ChainList.tsx
emreboga: Similar to https://linear.app/wormholelabs/issue/PROD-239/refactorrewrite-searchablelist-component

priyajeet (dismissed):

src/views/v3/Bridge/Routes/SingleRoute.tsx
```diff
@@ -504,16 +500,21 @@ const SingleRoute = (props: Props) => {
...
       return (
-        <>
+        <Box sx={{ display: 'flex', alignItems: 'center' }}>
           <FastestRoute sx={styles.fastestBadge} />
-          {props.isOnlyChoice ? 'Fast' : 'Fastest'}
-        </>
+          <Typography component="span" fontSize="14px" lineHeight="14px">
+            {props.isOnlyChoice ? 'Fast' : 'Fastest'}
+          </Typography>
+        </Box>
```
priyajeet: I think a change like this also would be nice to move to its own component to adhere with the single the responsibility principle since SingleRoute shouldn't have to care about what internal implementation of a badge looks like.

`<RouteBadge isFastest isCheapest isOnlyChoice />`

Will help with less double maintenance of styles/markup + since its an external component you wont need useMemo internally here.

src/views/v3/Bridge/AssetPicker/PickerBottomSheet.tsx
```diff
@@ -0,0 +1,160 @@
...
+          height: '5px',
+          backgroundColor: theme.palette.text.secondary,
+          borderRadius: '8px',
+        }}
+      />
+    </Stack>
+  );
+
+  // Chain list - memoize because it involves chain selection logic
+  const chainListContent = useMemo(
```
priyajeet: Don't think we should memoize components within other components. If this really needs to be memoized (as per your comment), then the default export of ChainList should be memoized. Otherwise we should just embed the component directly and let react figure out what needs to be rendered based on its reconciliation algorithm. AssetPickerDrawer is also memoized all the way at the bottom, which is already checking prop changes to some level. Also future versions of react will memoize functional components by default, so such code will automatically become redundant.

src/views/v3/Bridge/AssetPicker/PickerBottomSheet.tsx
```diff
@@ -0,0 +1,160 @@
...
+      onChainSelect,
+    ],
+  );
+
+  // Token list - memoize because it involves expensive token sorting and filtering
+  const tokenListContent = useMemo(
+    () =>
+      !showChainSearch &&
+      chainConfig && (
+        <TokenList
```
priyajeet, line 102: Same

src/views/v3/Bridge/AssetPicker/PickerBottomSheet.tsx
```diff
@@ -0,0 +1,160 @@
...
+        chainList={chainList}
+        selectedChainConfig={chainConfig}
+        showSearch={showChainSearch}
+        setShowSearch={setShowChainSearch}
+        wallet={wallet}
+        onChainSelect={onChainSelect}
+      />
+    ),
+    [
+      chainList,
```
priyajeet: Also note that memo only works over shallow comparison. So it works well over simple primitive values, but can fail if more complex objects are themselves are not memoized. You can probably test how many times a console.log happens to see if it is behaving as needed.

src/views/v3/Bridge/Routes/RoutesDesktop.tsx
```diff
@@ -0,0 +1,154 @@
...
+        Routes
+      </Typography>
+      <IconButton sx={{ opacity: 0.5, padding: 0 }} onClick={onClose}>
+        <CloseIcon sx={{ height: '24px', width: '24px' }} />
+      </IconButton>
+    </Box>
+  );
+
+  // List section - memoize because it involves expensive map operation
+  const routesList = useMemo(
```
priyajeet: Can/should be its own component in a new file, with an exported React.memo()

src/views/v3/Bridge/Routes/RoutesDesktop.tsx
```diff
@@ -0,0 +1,154 @@
...
+  cheapestRoute,
+  onRouteSelect,
+  onGasChange,
+  onSelectRoute,
+  selectButtonDisabled,
+}: RoutesDesktopProps) {
+  const theme = useTheme();
+
+  // Header section - static content, no need to memoize
+  const routesHeader = (
```
priyajeet: Can be its own component `<SelectRouteHeader>` similarly `<SelectRouteButton>` below.

src/views/v3/Bridge/Routes/RoutesMobile.tsx
```diff
@@ -0,0 +1,162 @@
...
+  cheapestRoute,
+  onRouteSelect,
+  onGasChange,
+  onSelectRoute,
+  selectButtonDisabled,
+}: RoutesMobileProps) {
+  const theme = useTheme();
+
+  // Header section - static content, no need to memoize
+  const routesHeader = (
```
priyajeet: Same comments as the other file for these functions in here. If they have similar implementation/styles, then re-used between the two files.

src/views/v3/Bridge/Routes/index.tsx
```diff
@@ -107,69 +107,56 @@ const Routes = ({
...
-    return routesList.filter((rs) => quotes[rs] !== undefined);
-  }, [routesList, quotes]);
+  const routesWithQuotes = routesList.filter((rs) => quotes[rs] !== undefined);
 
-  const fastestRoute = useMemo(() => {
-    return routesWithQuotes.reduce(
-      (fastest, route) => {
-        const quote = quotes[route];
-        if (!quote || !quote.success) return fastest;
+  const fastestRoute = routesWithQuotes.reduce(
```
priyajeet, line 108: For all these data manipulation functions which have nothing really to do with react, they can be just moved into simple utils.ts type file and exported as functions. It will also help with if you want to unit test them. Embedded here will make it hard to test the logic. + If in the future you want to go back and memoize them, you can still using memoize-one library which is pretty performant.

src/views/v3/Redeem/index.tsx
```diff
@@ -236,19 +231,12 @@ const Redeem = () => {
...
-    }
-
-    return route.AUTOMATIC_DEPOSIT;
-  }, [routeName]);
+  const isAutomaticRoute = !routeName
+    ? false
+    : (() => {
+        const route = config.routes.get(routeName);
+        return route ? route.AUTOMATIC_DEPOSIT : false;
+      })();
```
priyajeet: Can be re-written without making a function call

```
const hasRoute = !!config.routes.get(routeName); // or get(routeName || "")
const isAutomaticRoute = hasRoute ? route.AUTOMATIC_DEPOSIT : false;
```

src/views/v3/Redeem/index.tsx
```diff
@@ -364,34 +352,28 @@ const Redeem = () => {
   );
 
   // Time remaining to reach the estimated completion of the transaction
-  const remainingEta = useMemo(() => {
+  const remainingEta = (() => {
```
priyajeet, line 351: same comment, no need for self-executing function

src/views/v3/Redeem/index.tsx
```diff
@@ -364,34 +352,28 @@ const Redeem = () => {
...
-      text: '',
-      align: 'left',
-    };
+  const defaults: { text: string; align: Alignment } = {
+    text: '',
+    align: 'left',
+  };
 
-    let headerConfig;
+  let headerConfig;
```
priyajeet: Seems some of this code is repeated in prior file, would be helpful to re-use if extracted as function

src/views/v3/Bridge/index.tsx
```diff
@@ -545,27 +542,34 @@ const Bridge = () => {
...
   }, [styles.copyIcon, styles.doneIcon, errorCopied, txError, txErrorInternal]);
 
-  const amountValidationError = useMemo(
-    () => (
-      <AlertBannerV2
-        warning={!!amountValidation.warning}
-        error={!!amountValidation.error}
-        content={amountValidation.error || amountValidation.warning}
-        show={!!amountValidation.error || !!amountValidation.warning}
+  const amountValidationError = useMemo(() => {
```
priyajeet: `<AmountValidationError>` component/file

src/views/v3/Bridge/index.tsx
```diff
@@ -347,25 +386,26 @@ const Bridge = () => {
...
+      amountValidation,
+    ],
+  );
 
-  // Asset picker for the destination network and token
-  const destAssetPicker = useMemo(() => {
-    const quoteResult = quotes && route ? quotes[route] : undefined;
-    return (
+  // Destination asset picker
+  const destAssetPicker = useMemo(
```
priyajeet: Same comment as other files

src/views/v3/Redeem/index.tsx
emreboga: Huh! Must have been a Claude edit fail when I was going over useMemo usages 😅

src/views/v3/Redeem/index.tsx
emreboga: Yup, this is another one I refactored with Claude. I guess we are still getting to know eachother.

src/views/v3/Bridge/AssetPicker/PickerBottomSheet.tsx
emreboga: Yup, we don't need to useMemo this one. The componet itself is already memoized too.

src/views/v3/Bridge/AssetPicker/PickerBottomSheet.tsx
emreboga: Yup, these were mostly for better code organization task I was doing with Claude, but the JSX still not too big, so I'll move them to the component return.

src/views/v3/Bridge/AmountInput/index.tsx
emreboga: Let's address this one separately.
https://linear.app/wormholelabs/issue/PROD-255/rework-the-naming-of-components

src/views/v3/Bridge/AssetPicker/TokenList.tsx
emreboga: Will be addressed separately:
https://linear.app/wormholelabs/issue/PROD-239/refactorrewrite-searchablelist-component

src/views/v3/Redeem/index.tsx
emreboga: Yup, will move this into a separate file/component.

src/views/v3/Bridge/AmountValidationError.tsx
```diff
@@ -0,0 +1,49 @@
...
+};
+
+type Props = {
+  validation: ValidationResult;
+};
+
+const AmountValidationError = ({ validation }: Props) => {
+  const theme = useTheme();
+
+  const validationAlert = useMemo(() => {
```
priyajeet: This component is already memoized at line 49

priyajeet: This component is already memoized at line 49

priyajeet (approved):

PR #3610, opened 2025-06-18
Add referrer info to mayan routes

priyajeet:
Depends on https://github.com/mayan-finance/wormhole-sdk-route/pull/70

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@ledgerhq/hw-transport": "6.27.1",
         "@ledgerhq/hw-transport-webhid": "6.27.1",
         "@ledgerhq/logs": "6.12.0",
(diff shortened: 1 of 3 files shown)


src/components/SampleApp/index.tsx
```diff
@@ -71,13 +72,13 @@ const parseConfig = (config: string): WormholeConnectConfig => {
       /* @ts-ignore */
       window.TBTCRoute = routes.TBTCRoute;
       /* @ts-ignore */
-      window.MayanRoute = MayanRoute;
+      window.MayanRoute = createMayanRouteWithReferrerFee(MayanRoute);
```
priyajeet: Don't think this change is really needed since this is a sample app, will revert

PR #3611, opened 2025-06-20
ntt executor per token gas limits

kev1n-peters:
This allows us to override the gas limit for specific tokens (looking at you, Cheese token)


kev1n-peters:
pr is no longer relevant

PR #3615, opened 2025-06-24
Add NTT token price fallback mechanism

kev1n-peters:
When transferring NTT tokens, if the price for the source token is unavailable, the system now falls back to using prices from other tokens in the same NTT configuration group. This ensures better USD value estimation for transfers involving NTT tokens with limited price data.

- Add getNttTokenGroup() to find alternative tokens in the same NTT config
- Update getTokenPrice() to implement fallback logic for NTT tokens
- Extract nttRoutes constant for better code reuse

### DIFF
--- a/src/utils/ntt.ts
+++ b/src/utils/ntt.ts
@@ -1,33 +1,91 @@
 import { Chain, TokenId } from '@wormhole-foundation/sdk';
 import config from 'config';
-import { addressString } from 'config/tokens';
+import { Token, addressString } from 'config/tokens';
+
+export const nttRoutes = [
+  'ManualNtt',
+  'AutomaticNtt',
+  'M0AutomaticRoute',
+  'NttExecutorRoute',
+] as const;
+
+interface NttTokenOption {
+  chain: Chain;
+  token: string;
+}
+
+type NttTokensConfig = Record<string, NttTokenOption[]>;
+
+export const getNttTokens = (
+  routeName: string,
+): NttTokensConfig | undefined => {
+  const route = config.routes.get(routeName);
+  if (!route) return undefined;
+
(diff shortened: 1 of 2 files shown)


src/utils/ntt.ts
```diff
@@ -1,33 +1,91 @@
...
 export const isNttToken = (tokenId: TokenId): boolean => {
-  return (
-    ['ManualNtt', 'AutomaticNtt', 'M0AutomaticRoute', 'NttExecutorRoute']
-      .map((rn) => {
-        const route = config.routes.get(rn);
-        if (route) {
-          const nttConfig = (route.rc as any).config;
-          if (nttConfig) {
-            for (const key in nttConfig.tokens) {
-              const options: { chain: Chain; token: string }[] = nttConfig[key];
```
kev1n-peters, line 14: This function was broken since we weren't accessing `nttConfig.tokens[key]`

src/utils/ntt.ts
```diff
@@ -1,33 +1,91 @@
...
+  chain: Chain;
+  token: string;
+}
+
+type NttTokensConfig = Record<string, NttTokenOption[]>;
+
+export const getNttTokens = (
+  routeName: string,
+): NttTokensConfig | undefined => {
+  const route = config.routes.get(routeName);
```
emreboga, line 22: Is there a way to check if the route itself is an NTT route? The name of the function already has NTT in it so I don't think it's super important, but stil to be on the safe side :)

kev1n-peters: No, but we could probably add a static property to the routes like `IS_NTT_ROUTE = true` or some other type guard (probably better)

emreboga (approved):

PR #3630, opened 2025-07-09
feat: added Fogo testnet

kev1n-peters:


src/utils/wallet/solana.ts
```diff
@@ -28,45 +28,55 @@ import {
...
-                customStoragePrefix: 'wh-connect-solana-adapter',
-              },
-            }),
-            connection,
-          ),
-        }
-      : {}),
-  };
+export function fetchOptions(chain: Chain) {
+  if (chain === 'Solana') {
```
priyajeet, line 39: nit: from an organizational standpoint, I think this function should delegate to other functions `getOptionsForSolana()`, `getOptionsForFogo()` ... `getOptionsForXYZ()`, so that Single-responsibility principle is maintained + allows the function to evolve without adding more large if / else blocks and eventually can be switch/case

artursapek (approved):

priyajeet (approved):

src/utils/wallet/solana.ts
kev1n-peters: Will revisit in the external wallets feature which will touch this code

PR #3634, opened 2025-07-10
feat: Allow same chain to be selected for mono chain swaps

priyajeet:


vite.config.ts
```diff
@@ -102,6 +102,7 @@ const optimizeDeps = {
     '@mui/material/Tooltip',
     '@mui/material/Unstable_Grid2',
   ],
+  exclude: ['@mayanfinanace/wormhole-sdk-route'],
```
priyajeet: ignore this change, to be reverted

src/routes/sdkv2/route.ts
```diff
@@ -117,22 +121,40 @@ export class SDKv2Route {
...
-      this.rc.supportedDestinationTokens(
-        sourceToken.tokenId,
-        fromContext.context,
-        toContext.context,
-      ),
-    );
+    const nativeToken = Wormhole.tokenId(toChain, 'native');
+    const usdcToken = usdcAddr ? Wormhole.tokenId(toChain, usdcAddr) : null;
+    // If we have Mayan available, which is a swap route, by default we show the gas token and USDC.
+    const mayanTokens = usdcToken ? [nativeToken, usdcToken] : [nativeToken];
```
artursapek, line 142: moving this here is a good call :+1:

src/utils/sdkv2.ts
```diff
@@ -543,3 +544,26 @@ const getTokenBridgeToken = async (
 
   return config.tokens.get(tokenId);
 };
+
+export function isSameChainSwapEnabled(chain: Chain) {
```
artursapek: it would be better if the `Route` SDK interface had something like [`IS_SAME_CHAIN`](https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/main/connect/src/routes/route.ts#L87) so we're not hard-coding logic around one route. in theory routes are supposed to be plugins.

src/routes/sdkv2/route.ts
priyajeet: I do need to re-check if there isn't any issue of moving the mayan logic into this file. Since this function is also called from line 86 above and that code-path will now encounter this logic when before it never used to.

priyajeet: Do you think we should add a check to line 85 along the lines of if (mayan) then dont need to check destination tokens, just true, because of [this](https://github.com/mayan-finance/wormhole-sdk-route/pull/74/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R141) and [this](https://github.com/mayan-finance/wormhole-sdk-route/pull/74/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R152).

src/routes/operator.ts
```diff
@@ -302,9 +303,18 @@ class QuoteCache {
           }
           delete this.pending[key];
 
-          if (result.success && result.expires === undefined) {
-            // Default to 60 seconds expiry
-            result.expires = new Date(Date.now() + 60_000);
+          if (result.success) {
```
priyajeet, line 306: I had to update this logic to also account for bogus dates. For some reason mayan quote api is returning "0" which creates a date object with 1969. As such === undefined check is not sufficient to guard against bad dates. Without this the code was fetching quotes in a loop too fast.

src/routes/operator.ts
```diff
@@ -302,9 +303,18 @@ class QuoteCache {
...
+            const now = Date.now();
+
+            // A valid expiry should be at least 5 seconds in the future
+            // and if not, we should default to 60 seconds.
+            const isValidExpiry =
+              result.expires instanceof Date &&
+              result.expires.getTime() > now + 5_000;
```
priyajeet, line 313: We could also make the min default as 60 if anything is less than. 60

src/routes/operator.ts
kev1n-peters: Does Mayan know about this?

src/routes/operator.ts
kev1n-peters: Has Mayan suggested how frequently to fetch fresh quotes for same chain swaps?

src/routes/operator.ts
priyajeet: Looks like this is as intended

src/views/v3/Bridge/Routes/SingleRoute.tsx
```diff
@@ -371,26 +337,6 @@ const SingleRoute = (props: Props) => {
     destChain,
   ]);
 
-  const providerText = useMemo(() => {
```
priyajeet, line 375: Code moved to its own file before + added cases for Jupiter

src/routes/sdkv2/route.ts
```diff
@@ -119,22 +131,40 @@ export class SDKv2Route {
...
-    const destTokens = await this.tokenCache.requestWithCache(cacheKey, () =>
-      this.rc.supportedDestinationTokens(
-        sourceToken.tokenId,
-        fromContext.context,
-        toContext.context,
-      ),
-    );
+    const nativeToken = Wormhole.tokenId(toChain, 'native');
+    const usdcToken = usdcAddr ? Wormhole.tokenId(toChain, usdcAddr) : null;
+    // If we have Mayan available, which is a swap route, by default we show the gas token and USDC.
```
emreboga, line 141: I believe in Polygon native token is not the gas token (POL vs. ETH). Is this going to be problem here?

src/views/v3/Bridge/Routes/ProviderWithAmount.tsx
```diff
@@ -0,0 +1,33 @@
+import { memo } from 'react';
```
emreboga, line 1: Why do we need a separate component instead of adding an optional amount param to ProviderLabel? These two components seems to be 80% same.

src/views/v3/Bridge/Routes/SingleRoute.tsx
```diff
@@ -543,7 +499,7 @@ const SingleRoute = (props: Props) => {
                 ? `You pay gas on ${destChain}`
                 : null}
               {destinationGas}
-              {timeToDestination}
+              <TimeToDestination destChain={destChain} eta={quote?.eta} />
```
emreboga, line 502: eta param is optional but its type doesn't have undefined:
```
eta?: number;
```
You can get its value as undefined in this case, you don't see any ts errors?

src/routes/sdkv2/route.ts
kev1n-peters: `'native'` represents the gas token which is POL on Polygon (see POL being paid for gas in this TX for example https://polygonscan.com/tx/0x3f8adaaca052033fc63465adcfbeeda7b67d3240e532961b320320c0bed4b826)

src/views/v3/Bridge/Routes/SingleRoute.tsx
priyajeet: ? Signifies it can be undefined

src/routes/sdkv2/route.ts
priyajeet: This ok then? I just moved this existing code from [here](https://github.com/wormhole-foundation/wormhole-connect/blob/development/src/routes/operator.ts#L160-L168).

src/views/v3/Bridge/Routes/ProviderWithAmount.tsx
priyajeet: I can merge, but main reason I kept it separate was due to different return values + styling under both false/true conditions. Will need to add 2 more props (usd + if to show usd or not) which will add 2 more branches. Makes code harder to test in general the more branches that get added.

In this particular case the common part of getting the provider string is already extracted out and re-used.

src/routes/sdkv2/route.ts
kev1n-peters: I don't see an issue unless Emre is notices something that I don't

emreboga (approved):

PR #3650, opened 2025-07-18
chore: Refactored getExplorerUrl to handle query params in baseUrl

tonyjin:
- Created helper function `appendPathToUrl()` to append a path segment while retaining query params from the base
- Separated `getExplorerUrl()` into three separate helper functions based on type of URL - token, wallet, txn
- Updated Solana devnet to have correct baseUrl with `?cluster=devnet`

### DIFF
--- a/src/config/testnet/chains.ts
+++ b/src/config/testnet/chains.ts
@@ -43,7 +43,7 @@ export const TESTNET_CHAINS: ChainsConfig = {
   },
   Solana: {
     displayName: 'Solana',
-    explorerUrl: 'https://explorer.solana.com/',
+    explorerUrl: 'https://explorer.solana.com/?cluster=devnet',
     explorerName: 'Solana Explorer',
     icon: 'Solana',
     symbol: 'SOL',
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -253,6 +253,13 @@ export const isEmptyObject = (value: object | null | undefined) => {
 
 export type ExplorerPathType = 'wallet' | 'tx' | 'token';
 
+// Helper function to append path while preserving query params
+const appendPathToUrl = (baseUrl: string, pathSegment: string): string => {
+  const url = new URL(baseUrl);
+  url.pathname = `${url.pathname}/${pathSegment}`;
+  return url.toString();
+};
+
 export const getExplorerUrl = (
   chain: Chain,
   path: string,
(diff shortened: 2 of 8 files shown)


src/utils/index.ts
```diff
@@ -268,26 +275,30 @@ export const getExplorerUrl = (
 
   switch (pathType) {
     case 'wallet':
-      return chain === 'Aptos'
-        ? `${baseUrl}account/${path}`
-        : `${baseUrl}address/${path}`;
+      switch (chain) {
```
priyajeet: Switch case within a switch case is wild...error prone.

Probably should move each section (getWalletExplorerUrl / getTxExplorerUrl etc) out to its own function.

src/utils/index.ts
```diff
@@ -253,6 +253,13 @@ export const isEmptyObject = (value: object | null | undefined) => {
 
 export type ExplorerPathType = 'wallet' | 'tx' | 'token';
 
+// Helper function to append path while preserving query params
+const appendPathToUrl = (baseUrl: string, pathSegment: string): string => {
+  const url = new URL(baseUrl);
+  url.pathname = `${url.pathname}/${pathSegment}`;
```
priyajeet: `new URL(pathSegment, baseUrl)`
https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

tonyjin: Don't think that works? https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#examples

> new URL("/a", "https://example.com/?query=1");
// => 'https://example.com/a' (see relative URLs)

You lose the query params from the base, which we need to preserve, since the base URL could be something like `explorer.solana.com/?cluster=devnet` or `https://seitrace.com/?chain=atlantic-2`

priyajeet: ok, ic, variable name being baseUrl got me confused. But I think you might have to check for trailing / since pathname when none exists (or when baseUrl has a trailing /) returns / at the end, so it might end up with //pathsegment. And also need to make sure that there isn't a prefixed / on pathsegment, otherwise another / will get added due to concatenation.

priyajeet: Regex replace might work
`url.pathname = url.pathname.replace(/\/$/, '') + '/' + pathSegment.replace(/^\//, '');`

tonyjin:
@priyajeet split into separate functions, handle trailing/prefixed `/`, update Solana devnet URL

src/config/testnet/chains.ts
```diff
@@ -43,7 +43,7 @@ export const TESTNET_CHAINS: ChainsConfig = {
   },
   Solana: {
     displayName: 'Solana',
-    explorerUrl: 'https://explorer.solana.com/',
+    explorerUrl: 'https://explorer.solana.com/?cluster=devnet',
```
tonyjin, line 46: @kev1n-peters this look right?

kev1n-peters: The `?cluster=devnet` part should come after the URL path, e.g.:
https://explorer.solana.com/address/7yLKkp1HS2v9eJtPqT3crvhv7usuNWmQ87GLx48Ck8jJ?cluster=devnet
The functions that create the URLs (e.g. `getTokenExplorerUrl`) will need to check the network (I think you can get it from the `config` object) and append `?cluster=devnet` to the URL if `network === 'Testnet'`. We should leave `explorerUrl` here as the base URL

tonyjin: I started down this path since I noticed Sei testnet vs mainnet included the `?chain=` pattern in the base Url.

The helper functions `appendPathToUrl` actually handle this correctly now where it will move any query params to the end of the URL.

`appendPathToUrl('explorer.solana.com/?cluster=devnet', 'address/XXX')` => `explorer.solana.com/address/XXX?cluster=devnet`

artursapek: kinda funky but seems alright to me.

artursapek (approved):

PR #3658, opened 2025-07-21
Route badges redesign

emreboga:
Please see the loom for details.
https://www.loom.com/share/7f188ace7e364c69862edc249a74c023?sid=cb5e6a79-9c26-4eb1-a516-d1d59ac7bc92

### DIFF
--- a/src/views/v3/Bridge/Routes/index.tsx
+++ b/src/views/v3/Bridge/Routes/index.tsx
@@ -5,18 +5,21 @@ import Box from '@mui/material/Box';
 import Link from '@mui/material/Link';
 import Skeleton from '@mui/material/Skeleton';
 import Stack from '@mui/material/Stack';
+import ToggleButton from '@mui/material/ToggleButton';
+import ToggleButtonGroup from '@mui/material/ToggleButtonGroup';
 import Typography from '@mui/material/Typography';
 import ChevronRightIcon from '@mui/icons-material/ChevronRight';
 import { routes } from '@wormhole-foundation/sdk';
 
 import config from 'config';
 import ClockIcon from 'icons/Clock';
+import RoutingIcon from 'icons/Routing';
 import { millisToHumanString } from 'utils';
+import { OPACITY } from 'utils/style';
+import type { RootState } from 'store';
 import { setToNativeToken } from 'store/relay';
 import RoutesMobile from 'views/v3/Bridge/Routes/RoutesBottomSheet';
 import RoutesDesktop from 'views/v3/Bridge/Routes/RoutesModal';
-
-import type { RootState } from 'store';
 
 type Props = {
   routes: string[];
@@ -34,7 +37,7 @@ function Routes({
   isLoading,
 }: Props) {
   const dispatch = useDispatch();
-  const theme = useTheme();
+  const theme: any = useTheme();
   const mobile = useMediaQuery(theme.breakpoints.down('sm'));
 
   const { toNativeToken } = useSelector((state: RootState) => ({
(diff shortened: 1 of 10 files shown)


src/views/v3/Redeem/TransactionDetails/index.tsx
```diff
@@ -323,7 +323,7 @@ const TransactionDetails = () => {
       <Card sx={styles.card}>
         <CardContent>
           <Typography color={theme.palette.text.secondary} marginBottom="12px">
-            {`Transaction #`}
+            {`Transaction `}
```
priyajeet: I think we should avoid whitespaces since hard to read. Probably try 

```jsx
      <Stack direction="row" spacing={1}>
        <Typography>
          Transaction
        </Typography>
        <Typography>
          {explorerUrl ? (
              <ExplorerLink url={explorerUrl} text={trimmedTx} />
            ) : (
              trimmedTx
            )}
        </Typography>
      </Stack>
```

src/views/v3/Bridge/Routes/index.tsx
```diff
@@ -51,16 +54,93 @@ function Routes({
...
+
+      if (quote.eta !== undefined && quote.eta < fastest.eta) {
+        return { name: route, eta: quote.eta };
+      }
+      return fastest;
+    },
+    { name: '', eta: Infinity },
+  );
+
+  const cheapestRoute = routesWithQuotes.reduce(
```
priyajeet: I think we should move these functions out of the render/component logic into a new file/utils/something importable as they aren't directly related to react rendering. And I think you can reduce 3 loops to 1. Line 60, 69, 82 can be combined.

```ts
function getFastestAndCheapest(routesList) {
  const cheapest = ...
  const fastest = ...

  for (const route of routesList) {
    ...
  }
}
```

Optionally, you can [fast-memoize](https://www.npmjs.com/package/fast-memoize) this function if there is a need if routeList doesn't change too often but parent component renders more often.

src/views/v3/Bridge/Routes/index.tsx
```diff
@@ -51,16 +54,93 @@ function Routes({
...
     // Reset the highlighted route when the selected route changes
-    if (selectedRoute && selectedRoute !== highlightedRoute) {
-      setHighlightedRoute(selectedRoute);
+    if (selectedRoute) {
+      if (selectedRoute !== highlightedRoute) {
+        setHighlightedRoute(selectedRoute);
+      }
+
+      // Set the selected route badge based on the selected route
+      if (selectedRoute === fastestRoute.name) {
```
priyajeet, line 84: onChange below calls onRouteChange, does that send in the new selectedRoute to this component?

If so, you might be able to simplify quite a bit and reduce extra renders as you wont need state for the badge.  As in, selectedRoute is probably already in some state somewhere and everything else can be derived from it?

```ts
function getSelectedRouteBadge(selectedRoute, fastestRoute, cheapestRoute) {
  if (selectedRoute === fastestRoute.name) {
    return 'fastest';
  } else if (selectedRoute === cheapestRoute.name) {
    return 'cheapest';
  } else {
    return '';
  }
}

```
Likewise, maybe even for highLighted route, because aren't they supposed to be one and the same which is what this effect does on line 101? And if so, this whole effect can be nuked.

src/views/v3/Bridge/Routes/index.tsx
```diff
@@ -140,79 +220,56 @@ function Routes({
...
+        <Box sx={{ maxWidth: '174px' }}>
+          <ToggleButtonGroup
+            exclusive
+            size="small"
+            value={selectedRouteBadge}
+            sx={styles.toggleGroup}
+            onChange={(_, value) => {
+              if (value === 'fastest' && fastestRoute.name) {
+                onRouteChange(fastestRoute.name);
+                setSelectedRouteBadge('fastest');
```
priyajeet, line 241: See above comment, you may not need to set badge at all if onRouteChange already sets the route in state and passes it down to this component.

priyajeet (changes requested):

src/views/v3/Bridge/Routes/index.tsx
emreboga: This effect is to handle the case when selectedRoute changes externally. For instance user updates the amount and previously selected route is no longer available. In that case it automatically select the best available route and update it here as well.

src/views/v3/Bridge/Routes/index.tsx
emreboga: Yes, but in one case, where fastest and cheapest are the same route, we need to setSelectedRouteBadge here as onRouteChange won't trigger the side-effect.

src/views/v3/Redeem/TransactionDetails/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3658/commits/8a06627a3c6fa11096eeb19658549790902eeffe

src/views/v3/Bridge/Routes/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3658/commits/8a06627a3c6fa11096eeb19658549790902eeffe

priyajeet (approved):

PR #3663, opened 2025-07-22
feat: support comma separators in the bridge amount input

kyleleighton:
This adds number-formatting functions based on a user's locale. This will convert numbers (and currencies) to have separators that match the user's locale. This specifically updates the UI of the Bridge V2 and V3 amount inputs and converted values. The input values themselves are still unformatted numbers.

Previously, we did not have number formatting. So if you were trading 123,456,789 tokens, that would display as 123456789, which is hard to read. Now it displays as 123,456,789, or whatever the equivalent is for your locale. 

Tested with fractional amounts (0.0001 is fine to enter) and huge amounts (1,000,000,000,000), and in between. All good.

Screenshot 2025-07-22 at 12 30 03 PM
Screenshot 2025-07-22 at 12 32 55 PM

> NOTE: we read the locale via fallback from config -> documentElement.lang -> navigator.language. However, we don't currently _store_ the locale in the config. This is something we will likely want to consider as the app expands and more markets enter.

Linear: https://linear.app/wormholelabs/issue/WFREQ-144/add-thousands-separator-comma-in-numerical-displays


src/config/types.ts
```diff
@@ -196,6 +196,8 @@ export interface InternalConfig<N extends Network> {
   guardianSet: GuardianSetData;
 
   transactionSettings: TransactionSettings;
+
+  locale?: string;
```
priyajeet: Don't have a strong opinion either way, but should this be nested under the `ui` sub-config thats on line 191?

priyajeet: @emreboga thoughts?

src/views/v2/Bridge/AmountInput/index.tsx
```diff
@@ -51,19 +56,17 @@ const DebouncedTextField = memo(
...
+        if (!isValidDecimalInput(value)) {
           return;
         }
 
-        setInnerValue(e.target.value);
-        onChange(e.target.value); // callback with no delay
-        deferredOnChange(e.target.value);
+        const formattedValue = formatWithCommas(value);
+
+        setInnerValue(formattedValue);
```
priyajeet, line 67: @emreboga 
Kind of curious, why we have 2 setStates that track each other. We already set the value in state, innerValue can just be derived directly? Which means, all we need to do is change line 88 to be `formatWithCommans(value)` and innerValue can just be deleted?

Is it due to the focus issue since there is a guard for it? But if a user is focused and typing, it should update what they type, no?

In any case @kyleleighton we probably should not set UI-formatted values in state, only raw data. How that is represented visually, should be at the render level.

src/utils/formatNumber.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (!value) return '';
+
+  const locale = getUserLocale();
+  const thousand = (1000).toLocaleString(locale).charAt(1);
+  const decimal = (1.1).toLocaleString(locale).charAt(1);
+
+  return value.split(thousand).join('').replace(decimal, '.');
+};
+
+export const isValidDecimalInput = (value: string): boolean => {
```
priyajeet, line 86: Probably should rename to say isValidPositiveDecimal, since there is a >= 0 check too.

src/utils/formatNumber.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (!value) return true;
+
+  // Allow valid decimal number patterns (including starting with ".")
+  if (!/^\d*\.?\d*$/.test(value)) {
```
priyajeet: Is this needed? since 57 takes care of it, no?

src/utils/formatNumber.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const locale = getUserLocale();
+  const thousand = (1000).toLocaleString(locale).charAt(1);
+  const decimal = (1.1).toLocaleString(locale).charAt(1);
+
+  return value.split(thousand).join('').replace(decimal, '.');
+};
+
+export const isValidDecimalInput = (value: string): boolean => {
+  if (!value) return true;
```
priyajeet: I think we should specifically check for string/empty string. And return false for anything else. There have been non-0 browser bugs where users have pasted or typed into the field things other than numbers + since this is a utility function which can be used everywhere incl UI, better to be safe.

priyajeet: BTW there is also `valueAsNumber` prop on input.type === number, in case that helps at all.

src/utils/formatNumber.ts
```diff
@@ -0,0 +1,58 @@
...
+    return false;
+  }
+
+  // Skip number validation for "." to allow "0." input
+  if (value === '.') {
+    return true;
+  }
+
+  const numValue = Number(value);
+  return !isNaN(numValue) && numValue >= 0;
```
priyajeet: `Number.isFinite(parseFloat(...))` might be a better check than isNaN.

src/utils/formatNumber.ts
```diff
@@ -0,0 +1,58 @@
+import config from 'config';
+
+const getUserLocale = (): string => {
+  return (
+    config?.locale ||
+    (typeof document !== 'undefined' && document.documentElement.lang) ||
+    (typeof navigator !== 'undefined' && navigator.language) ||
```
priyajeet: Can follow the style of line 6 too, using ?.?. instead of typeof navigator !== 'undefined'

priyajeet: On portal

document.documentElement.lang
>> 'en'

navigator.language
>> 'en-US'

Should we flip the check so that the specific comes 1st?

src/utils/formatNumber.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  return formatted;
+};
+
+export const removeCommas = (value: string): string => {
+  if (!value) return '';
+
+  const locale = getUserLocale();
+  const thousand = (1000).toLocaleString(locale).charAt(1);
```
priyajeet: To get the .

```
function getDecimalSeparator() {
  const numberFormat = new Intl.NumberFormat(locale);
  const parts = numberFormat.formatToParts(1.1);
  return parts.find(part => part.type === 'decimal').value;
}
```

though should cache it

src/utils/formatNumber.ts
```diff
@@ -0,0 +1,58 @@
...
+export const isValidDecimalInput = (value: string): boolean => {
+  if (!value) return true;
+
+  // Allow valid decimal number patterns (including starting with ".")
+  if (!/^\d*\.?\d*$/.test(value)) {
+    return false;
+  }
+
+  // Skip number validation for "." to allow "0." input
+  if (value === '.') {
```
priyajeet: === decimal, where decimal is derived from Intl, no?

src/utils/formatNumber.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (!value) return '';
+
+  const locale = getUserLocale();
+  const integerFormatter = new Intl.NumberFormat(locale, {
+    minimumFractionDigits: 0,
+    maximumFractionDigits: 0,
+    useGrouping: true,
+  });
+
+  const [integerPart, decimalPart] = value.split('.');
```
priyajeet, line 42: Does doing integerFormatter.format(value) not suffice?

src/utils/formatNumber.ts
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    config?.locale ||
+    (typeof document !== 'undefined' && document.documentElement.lang) ||
+    (typeof navigator !== 'undefined' && navigator.language) ||
+    'en-US'
+  );
+};
+
+export const formatWithCommas = (value: string): string => {
+  if (!value) return '';
```
priyajeet: Should we just call the internal isValidDecimalInput() here. Likewise for when removing commas.

src/utils/formatNumber.ts
kyleleighton: I'm fine with that, but will there ever be a case where a negative decimal is valid? If so, maybe one function where you can pass in `(value, { allowNegative: true })`? But again, not sure when that would be used.

src/utils/formatNumber.ts
kyleleighton: We could, but that returns a `boolean` and this returns a `string`, and it also means we combine concerns that I think should be separated. `isValidDecimalInput` checks whether a value is valid, whereas the comma functions format the already-validated `value` (or no `value` at all, which _could_ be considered valid).

src/utils/formatNumber.ts
kyleleighton: The regex prevents invalid inputs like "1.2.3" or "123abc" that `Number()` would partially parse incorrectly. Nothing would break if I removed it, but we wouldn't have preventative measures that are probably safer to have.

Edit: handled it differently

src/utils/formatNumber.ts
kyleleighton: Good catch, but the `integerFormatter` is set to zero `fractionDigits`, so `.format()` only gives us the `group`ing character (not the `decimal` or any fraction).

If we want to eventually handle locales that use commas or other symbols and to keep a trailing separator while someone’s typing, we need both the `group` and `decimal` characters (via `formatToParts`) and split on the real decimal symbol.

Increasing the fraction digits would have much bigger problems, like rounding and trailing separators.

src/utils/formatNumber.ts
kyleleighton: Nice! I did not know about `formatToParts`.

priyajeet (approved):

priyajeet (dismissed):

src/config/types.ts
emreboga: Yes, should be nested under UI; but before that, why do we need this prop in config? Was there a requirement to allow integrators to override the locale?

src/utils/formatNumber.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  return formattedInt;
+};
+
+/**
+ * Strip locale‑specific grouping separators and convert the decimal
+ * separator to "." so Number() will parse correctly.
+ */
+export const removeCommas = (value: string): string => {
+  if (!value) return '';
```
emreboga: Nitpick: For single line if statements I see both with and without brackets across this PR. Can you make them all with brackets? I think we should also enable the linter rule for that int he future for consistency.

src/views/v2/Bridge/AmountInput/index.tsx
emreboga: TBH I can't recall why did we have innerValue. DebouncedTextField component was added about a year ago by someone from the xLabs team to prevent the quote request every time user types in amount. WE haven't touched the logic there much since then, but 100% on revisiting this component.

emreboga (approved):
LGTM!

PR #3664, opened 2025-07-22
fix: trailing slash in block explorer urls

aapclark:
run linter

Addresses https://github.com/wormhole-foundation/wormhole-connect/issues/3661


artursapek:
this is handled by https://github.com/wormhole-foundation/wormhole-connect/pull/3650, I don't think we need both changes

PR #3667, opened 2025-07-23
Token loading time improvements

artursapek:


emreboga (dismissed):

src/hooks/useGetTokenBalances.ts
```diff
@@ -42,28 +42,24 @@ const useBalanceCache = () => {
...
-      return null;
-    },
-    [],
-  );
+  const getCached = useCallback((wallet: WalletData, token: Token) => {
+    const key = cacheKey(wallet.address, token);
+    const cached = cacheRef.current[key];
+    const now = Date.now();
+
+    if (cached && cached.lastUpdated > now - CACHE_DURATION_MS) {
```
priyajeet: Super long file, so I might have missed some logic/understanding.

If we check cached object staleness here before returning, is there is use case of the entire useEffect on line 80. Also that useEffect relies on intervals that are not guaranteed to run in the specified interval due to callback stacking, so wondering what happens if we just nuke out deleting things out of cache.

artursapek: I believe you are right - this could be removed.

artursapek: [84dd47e](https://github.com/wormhole-foundation/wormhole-connect/pull/3667/commits/84dd47e189d6893e91d054797863ad258d4000fe)

priyajeet (dismissed):

priyajeet (approved):

emreboga (approved):

PR #3679, opened 2025-07-28
Fix min amt error

artursapek:
- Passes failed quotes thru the quote pipeline, so amount validation has access to them
- Formats minimum amounts sanely (don't show Ethereum amounts to 18 decimal places)

image

image

### DIFF
--- a/src/hooks/useAmountValidation.ts
+++ b/src/hooks/useAmountValidation.ts
@@ -3,6 +3,7 @@ import { useMemo } from 'react';
 import { useSelector } from 'react-redux';
 import { QuoteResult } from 'routes/operator';
 import { RootState } from 'store';
+import { formatMinAmount } from 'utils/formatNumber';
 import { isMinAmountError } from 'utils/sdkv2';
 
 export type AmountValidationResult = {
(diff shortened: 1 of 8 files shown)


src/hooks/useAmountValidation.ts
```diff
@@ -88,11 +86,26 @@ export const useAmountValidation = (props: Props): AmountValidationResult => {
...
   }
 
   return {};
 };
+
+// Minimum amounts are approximations anyway so we don't need to display ultra-precise figures here.
+// For amounts >  999, we simply round up.
+// For amounts <= 999, we use toPrecision(3) which only shows the first 3 non-zero digits.
+// This way we're not showing excessive precision for any value
+const formatMinAmount = (minAmount: sdkAmount.Amount): string => {
```
priyajeet: We should move this inside `src/utils/formatNumber.ts`

src/hooks/useFetchQuotes.ts
```diff
@@ -411,8 +416,17 @@ export default (routes: string[], params: Params): HookReturn => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [unfilteredQuotes]);
 
+  const failedQuotes = useMemo(() => {
```
priyajeet: If you want, we can add es-toolkit npm (aka newer lodash) to take care of utilities like this. For example here and above we can use [partition](https://es-toolkit.dev/reference/array/partition.html) to get both success/failed.

priyajeet (approved):

PR #3695, opened 2025-08-05
Addressing Design QA feedback

emreboga:
This PR will address all design QA items in PROD-298 except for items 6 and 11. For these two items please see below.

6 -> Added a new config prop to hide the footer in Connect including Powered By icon. This will allow us to hide it in Portal and show them in the desired locations specified in the ticket.
11 -> This item is postponed to phase-2 of Connect UI refresh.


src/utils/index.ts
```diff
@@ -367,18 +367,18 @@ export const millisToRelativeTime = (ts: number): string => {
   if (ts > daysMultiplier) {
```
priyajeet, line 367: Maybe a task for backlog, to look at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat to see if some of this can be cleaned up.

[react-intl](https://formatjs.github.io/docs/react-intl/components/#formattedrelativetime) uses the above. There is also https://date-fns.org/

priyajeet (approved):

src/utils/index.ts
emreboga: Yes, this is an old utility, we can definitely revisit to use the browser intl libs.

emreboga: https://linear.app/wormholelabs/issue/PROD-447/use-browser-libs-for-relative-time-format

PR #3698, opened 2025-08-06
feat: added wallet provider interface

kev1n-peters:
The purpose of this PR is to allow integrators to use their own wallet provider like Privy or Dynamic instead of Connect’s internal wallet provider that only supports "external" / injected wallets (e.g. Metamask).


src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,213 @@
...
+}) => {
+  const [isConnecting, setIsConnecting] = useState(false);
+
+  const previousProvider = useRef<WormholeConnectWalletProvider | null>(null);
+
+  const dispatch = useDispatch();
+
+  const setupWalletConnection = useCallback(
+    (wallet: Wallet, chain: Chain, type: TransferWallet): void => {
+      console.log(`Setting up wallet connection for ${chain} (${type})...`);
```
kev1n-peters: Please ignore the console logs, using for debugging

src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,213 @@
...
+export const WalletContext = createContext<WalletContextType | undefined>(
+  undefined,
+);
+
+interface WalletProviderProps {
+  children: ReactNode;
+  provider?: WormholeConnectWalletProvider;
+}
+
+export const WalletProvider: React.FC<WalletProviderProps> = ({
```
priyajeet: Probably move this to its own file called `WalletProvider.tsx`, since this file is `WalletContext.tsx`

src/WormholeConnect.tsx
```diff
@@ -14,17 +14,25 @@ import type { WormholeConnectConfig } from './config/types';
 import type { WormholeConnectTheme } from 'theme';
 import { RouteProvider } from './contexts/RouteContext';
 import { TokensProvider } from './contexts/TokensContext';
+import { WalletProvider } from './contexts/WalletContext';
+import type { WormholeConnectWalletProvider } from './utils/wallet/types';
```
kev1n-peters, line 18: Is this type name too verbose? Prefixing with `WormholeConnect` to be consistent with the other props

src/contexts/WalletContext.tsx
priyajeet: I think for functional components, if you use this syntax, typing becomes a bit cleaner as you don't need to reference React.FC.

```tsx
function WalletProvider({ children, provider }: WalletProviderProps) {
   ...
}
```

src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,213 @@
...
+    connectLastUsed?: boolean,
+  ) => Promise<Wallet>;
+  swapWallets: () => boolean;
+  disconnectWallet: (chain: Chain, type: TransferWallet) => Promise<void>;
+  isConnecting: boolean;
+  walletProvider: WormholeConnectWalletProvider;
+  isInternalProvider: boolean;
+}
+
+export const WalletContext = createContext<WalletContextType | undefined>(
```
priyajeet: Should ideally be this file's default export.

```tsx
const WalletContext = createContext<WalletContextType | undefined>(...);

export default WalletContext;
```

src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,213 @@
...
+  );
+
+  return (
+    <WalletContext.Provider value={contextValue}>
+      {children}
+    </WalletContext.Provider>
+  );
+};
+
+export const useWallet = (): WalletContextType => {
```
priyajeet: Probably move it to its own file `useWallet.ts`

src/hooks/useConfirmTransaction.ts
```diff
@@ -155,6 +157,7 @@ const useConfirmTransaction = (props: Props): ReturnProps => {
           destChain,
           receivingWallet.address,
           destToken,
+          walletProvider,
```
kev1n-peters: Not the biggest fan of passing the `walletProvider` here, but it's needed in the SDKv2Signer class [here](https://github.com/wormhole-foundation/wormhole-connect/pull/3698/files#diff-f99342fb595ea5a206b2e09ea318ac2a8af6f59d9c8478b4244a26a90112e53aR129) to get the wallet.

src/utils/wallet/index.ts
```diff
@@ -59,206 +45,14 @@ export const walletAcceptedChains = (
     .map((c) => c.sdkName);
 };
 
-export const setWalletConnection = (type: TransferWallet, wallet: Wallet) => {
-  walletConnection[type] = wallet;
-};
-
-// Returns false if the wallet connection was rejected by the user
-export const connectWallet = async (
```
kev1n-peters, line 67: These functions have been moved to the WalletContext

src/utils/wallet/index.ts
```diff
@@ -59,206 +45,14 @@ export const walletAcceptedChains = (
...
-  if (name !== ReadOnlyWallet.NAME) {
-    localStorage.setItem(config.cacheKey(`wallet:${platform}`), name);
-  }
-
-  return true;
-};
-
-// Checks localStorage for previously used wallet for this chain
-// and connects to it automatically if it exists.
-export const connectLastUsedWallet = async (
```
kev1n-peters, line 157: This has been moved to useConnectToLastUsedWallet.ts

src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+      const payload = {
+        address,
+        type: chainToPlatform(chain),
+        icon: wallet.getIcon(),
+        name: wallet.getName(),
+      };
+
+      if (type === TransferWallet.SENDING) {
+        dispatch(connectSourceWallet(payload));
```
kev1n-peters: Ideally all wallet state would exist in this context or the wallet provider. But I don't want to refactor too much in this PR

src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,208 @@
...
+      }
+    },
+    [setupWalletConnection, provider, isInternal],
+  );
+
+  const swapWallets = useCallback((): boolean => {
+    const success = provider.swapWallets();
+    if (success) {
+      dispatch(swapWalletsAction());
+      dispatch(swapInputs());
```
priyajeet: Wondering if swapInputs and setAmount can belong somewhere else. Maybe inside the implementation of swapWalletsAction or part of some effect in the file that handles it. Reason being, the wallet context shouldn't care about amount or inputs directly here otherwise it couples wallets with other parts of ui.

src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,208 @@
...
+      try {
+        let wallet: Wallet | null = null;
+        if (
+          connectLastUsed &&
+          isInternal &&
+          provider instanceof InternalWalletProvider
+        ) {
+          wallet = await provider.connectLastUsedWallet(chain, type);
+          if (!wallet)
+            throw new Error(`No last used wallet found for ${chain}`);
```
priyajeet: Is it expected the parent components catch this error? Alternative would be to have an error value in state, and return it as part of the context value, like an error code + message. Code for checking against.

BTW, Error() constructor can take a 2nd arg now [{ cause }](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) which can be `{ code: string, error: Nested-Error-object }` or whatever shape you want.

priyajeet: Another question to the above, should we just do `else` part when !wallet above?

src/WormholeConnect.tsx
```diff
@@ -41,16 +50,27 @@ export default function WormholeConnect({
     [theme],
   );
 
+  const walletProviderData = React.useMemo(() => {
+    const provider = externalProvider ?? new InternalWalletProvider();
```
priyajeet: Can the provider itself have an `isInternal` flag on the object? Which is set by the InternalWalletProvider() constructor? If so, then we may not need an additional prop outside.

src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,208 @@
...
+    [dispatch],
+  );
+
+  useEffect(() => {
+    provider.on('walletConnected', setupWalletConnection);
+
+    return () => {
+      provider.off('walletConnected', setupWalletConnection);
+    };
+  }, [provider, setupWalletConnection]);
```
priyajeet: Wondering if we can rely on some immutable property on `provider` to trigger the effect. Since `provider` is a complex object and can be passed in, we cannot guarantee its immutability (unlikely, but non-0). So this effect risks running more than once. Maybe if there was some string `provider.name` that we can rely on. Or we just remove it from deps and eslint disable the exhaustive deps since provider should ideally never change.

src/contexts/WalletContext.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const disconnectWallet = useCallback(
+    async (chain: Chain, type: TransferWallet): Promise<void> => {
+      console.log(`Disconnecting wallet for ${type}...`);
+      try {
+        const wallet = provider.getWallet(chain, type);
+        if (wallet) {
+          wallet.disconnect();
+        }
+      } catch (error) {
+        console.error('Error disconnecting wallet:', error);
```
priyajeet: Will there be a throw here also like the above function? Will probably have to make the error state consistent for parents to act upon.

src/contexts/WalletContext.tsx
priyajeet: Likewise for others in the useMemo() / useCallback() functions below since they all rely on provider object.

src/contexts/WalletContext.tsx
priyajeet: Thinking more about this, I think since we are in the context's folder scope, we should break down the file to have a hierarchy something like

```
.../contexts
        |__ wallet
               |__ WalletContext.tsx
               |__ WalletProvider.tsx
               |__ useWalletProvider.ts
```

src/hooks/useConfirmTransaction.ts
```diff
@@ -48,6 +51,7 @@ const useConfirmTransaction = (props: Props): ReturnProps => {
   );
 
   const routeContext = useContext(RouteContext);
+  const { walletProvider, isInternalProvider } = useWallet();
```
priyajeet: `useWalletProvider()` terminoligy might better represent whats being returned, since there will be a useWallet() to return the wallet itself at some point.

src/hooks/useConnectToLastUsedWallet.ts
```diff
@@ -0,0 +1,53 @@
...
+import type { Chain } from '@wormhole-foundation/sdk';
+import { TransferWallet } from 'utils/wallet';
+import { useWallet } from 'contexts/WalletContext';
+
+export const useConnectToLastUsedWallet = (
+  sourceChain?: Chain,
+  destChain?: Chain,
+): { isConnecting: boolean } => {
+  const { connectWallet } = useWallet();
+  const [isConnecting, setIsConnecting] = useState(false);
```
priyajeet, line 11: I noticed that the wallet context provider also has boolean state like this. Since this was moved from the other file, guessing this is some legacy requirement?

src/utils/wallet/provider.ts
```diff
@@ -0,0 +1,234 @@
...
+import type { TransferWallet } from '.';
+import { getWalletOptions } from '.';
+import type {
+  WormholeConnectWalletProvider,
+  WalletProviderEvents,
+} from './types';
+import type { Wallet } from '@wormhole-labs/wallet-aggregator-core';
+import config from 'config';
+
+export class InternalWalletProvider implements WormholeConnectWalletProvider {
```
priyajeet: If you want to get away from class/new/this, you can have this file as a simple factory function. Normally I avoid classes whenever possible as they are harder to type, more boiler plate to test, and composition > inheritance.

```
function createWalletProvider() {
    let pendingConnect;
    let walletConnections;
      
    function setWalletConnection() {...}
    function getWalletConnection() {...}
    function connectWalletToChain() {...}
    function connectWallet() {...}
    ...

    return { connectWallet, getWallet }; // whatever is needed by WormholeConnectWalletProvider type
}
```

Parent will just be `const internalWalletProvider = createWalletProvider()`

priyajeet: In this particular case, you can even choose to not have the factory outer function `function createWalletProvider()` since its a 1-time use factory. And just have all the inner stuff at the top level. Just export the return the value {...} needed by `satisfies WormholeConnectWalletProvider` type.

src/utils/wallet/provider.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  private getWalletConnection(type: TransferWallet): Wallet | undefined {
+    return this.walletConnections[type]?.wallet;
+  }
+
+  private async connectWalletToChain(
+    wallet: Wallet,
+    chain: Chain,
+  ): Promise<void> {
+    const _chainId = nativeChainIds.networkChainToNativeChainId.get(
```
priyajeet: Can we use the `chainToChainId()` method from `@wormhole-foundation/sdk` here?

src/utils/wallet/provider.ts
```diff
@@ -0,0 +1,234 @@
...
+      config.network,
+      chain,
+    );
+    const chainId = typeof _chainId === 'bigint' ? Number(_chainId) : _chainId;
+
+    await wallet.connect({ chainId });
+
+    const address = wallet.getAddress();
+    if (!address) {
+      throw new Error(
```
priyajeet: Q: Would it simplify code in caller if instead of throwing, we return the address? As in, can the caller avoid needing a try / catch? Or is there other errors that can be thrown from say wallet.connect() that have to be try / catched

src/utils/wallet/provider.ts
```diff
@@ -0,0 +1,234 @@
...
+   * 3. When the user selects a wallet, the sidebar calls onWalletSelected()
+   * 4. onWalletSelected() resolves this promise with the selected wallet
+   * 5. If the user cancels, onWalletSelectCancelled() rejects the promise
+   */
+  connectWallet(chain: Chain, type: TransferWallet): Promise<Wallet> {
+    if (this.pendingConnect) {
+      return Promise.reject(new Error('Wallet connection already in progress'));
+    }
+
+    return new Promise<Wallet>((resolve, reject) => {
```
priyajeet: Take a look at [Promise.withResolvers()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers#description)

src/utils/wallet/provider.ts
```diff
@@ -0,0 +1,234 @@
...
+   * When this method is called:
+   * 1. A promise is created and returned immediately (does not resolve yet)
+   * 2. The calling UI components will open the wallet sidebar
+   * 3. When the user selects a wallet, the sidebar calls onWalletSelected()
+   * 4. onWalletSelected() resolves this promise with the selected wallet
+   * 5. If the user cancels, onWalletSelectCancelled() rejects the promise
+   */
+  connectWallet(chain: Chain, type: TransferWallet): Promise<Wallet> {
+    if (this.pendingConnect) {
+      return Promise.reject(new Error('Wallet connection already in progress'));
```
priyajeet: Should this just return `this.pendingConnect`

src/utils/wallet/provider.ts
```diff
@@ -0,0 +1,234 @@
...
+          console.log(
+            `Auto-connecting to wallet ${lastUsedWallet} for ${chain} (${type})...`,
+          );
+
+          await this.connectWalletToChain(wallet, chain);
+
+          this.setWalletConnection(chain, type, wallet);
+
+          return wallet;
+        } catch (error) {
```
priyajeet: You probably can combine the inner/outer try catch. If error thrown have a specific name or code, can test against that if things need to be done differently

src/utils/wallet/provider.ts
```diff
@@ -0,0 +1,234 @@
...
+          localStorage.removeItem(localStorageKey);
+          console.error(
+            `Failed to connect to wallet ${lastUsedWallet} for ${chain}:`,
+            error,
+          );
+          return null;
+        }
+      } else {
```
priyajeet: Can reduce the arrowness of the function by moving localStorage.removeItem(localStorageKey); out at the bottom I think. May not need else.

src/views/v3/Bridge/WalletConnector/index.tsx
```diff
@@ -84,18 +97,29 @@ const WalletConnector = (props: Props) => {
...
           {button}
-          <WalletSidebar
-            open={isOpen}
-            type={props.type}
-            onClose={() => {
-              setIsOpen(false);
-            }}
-            showAddressInput={props.type === TransferWallet.RECEIVING}
-          />
+          {isInternalProvider && (
```
priyajeet: If `isInternal` can go on the provider itself, this check may not be needed since WalletSidebar can just return null based off its walletProvider prop

src/WormholeConnect.tsx
priyajeet: Seems fine

src/views/v3/Redeem/index.tsx
```diff
@@ -916,13 +941,16 @@ function Redeem() {
...
       )}
       <PoweredByIcon color={theme.palette.text.primary} />
-      <WalletSidebar
-        open={isWalletSidebarOpen}
-        type={TransferWallet.RECEIVING}
-        onClose={() => {
-          setIsWalletSidebarOpen(false);
-        }}
-      />
+      {walletProvider instanceof InternalWalletProvider && (
```
priyajeet: Same as other comment

src/views/v3/Redeem/index.tsx
```diff
@@ -795,7 +808,16 @@ function Redeem() {
           <Button
             variant="primary"
             styleOverrides={styles.actionButton}
-            onClick={() => setIsWalletSidebarOpen(true)}
+            onClick={async () => {
```
priyajeet: Probably move it out to a handleClick() function. Ideally this button should have been its own file called ReceivingWalletButton.tsx or something

src/hooks/useConnectToLastUsedWallet.ts
priyajeet: Also wondering if this file can be just nuked in favor of the provider doing the connection. Since the provider already prefer's the last connection.

src/contexts/WalletContext.tsx
kev1n-peters: I think inside of the file that handles it should be okay (ui should care about ui things)

src/WormholeConnect.tsx
kev1n-peters: Yeah, added a readonly isInternal flag and a type guard

src/hooks/useConfirmTransaction.ts
kev1n-peters: agree

src/utils/wallet/provider.ts
kev1n-peters: This is giving us the native chain ID and not the Wormhole chain ID. There isn't a `chainToNativeChainId` function at the moment.

src/views/v3/Bridge/WalletConnector/index.tsx
kev1n-peters: good idea

src/utils/wallet/provider.ts
kev1n-peters: good call

src/utils/wallet/provider.ts
kev1n-peters: I think it can, but we should still make sure the pendingConnect is for the given chain and type here

src/utils/wallet/provider.ts
kev1n-peters: Agree that this is cleaner, but requires es2024 (Connect is currently using es2021 and I'm not sure if we should change that for this right now)

src/contexts/WalletContext.tsx
kev1n-peters: Connect did not previously have parent components catch any errors here, so I'm just maintaining parity for now by swallowing the errors. Maybe we can leave the behavior like this until we need the errors in parent components?

src/contexts/WalletContext.tsx
kev1n-peters: If the provider reference changes, then we want this effect's cleanup function to run and remove the listeners on the previous reference, no? Maybe I'm misunderstanding something

src/hooks/useConnectToLastUsedWallet.ts
kev1n-peters: For the first question, it's a legacy requirement. I just moved the hook to its own file without refactoring much. I would prefer to keep it as is for now. For the second question, this hook passes the `autoConnect` flag which tells the provider to connect to the last used wallet (which should not require any user action -- if the `autoConnect` flag were false, then Connect would require the user to select a wallet in the sidebar). I'm not sure how the provider would know which action to take without this flag (autoConnect vs manual user action)

src/utils/wallet/provider.ts
kev1n-peters: Will change to use a functional approach with closures instead of inheritance

kyleleighton (dismissed):

src/views/v3/Bridge/WalletConnector/Sidebar.tsx
```diff
@@ -151,23 +163,22 @@ const WalletSidebar = (props: Props) => {
...
+      );
+    } catch (error) {
+      console.error('Failed to select wallet:', error);
+    }
+  }, [address, selectedChain, onClose, dispatch, walletProvider]);
+
+  const handleClose = useCallback(() => {
+    walletProvider.onWalletSelectCancelled();
+    props.onClose?.();
+  }, [walletProvider, props]);
```
priyajeet: props.onClose, so that we aren't depending on the whole props object

src/views/v3/Bridge/WalletConnector/Sidebar.tsx
```diff
@@ -38,13 +40,15 @@ type Props = {
...
   showAddressInput?: boolean;
+  walletProvider: WormholeConnectWalletProvider;
 };
 
 // Renders the sidebar on the right-side to display the list of available wallets
 // for the selected source or destination chain.
 const WalletSidebar = (props: Props) => {
   const dispatch = useDispatch();
   const theme = useTheme();
+  const walletProvider = props.walletProvider as InternalWalletProvider;
```
priyajeet: Is the Sidebar not have access to the wallet provider context so that we can get it out of the hook?

src/views/v3/Bridge/WalletConnector/Sidebar.tsx
```diff
@@ -123,9 +127,17 @@ const WalletSidebar = (props: Props) => {
       }
 
       onClose?.();
```
priyajeet, line 130: Should onClose be called on error case below?

src/utils/wallet/InternalWalletProvider.ts
```diff
@@ -0,0 +1,280 @@
...
+  ) => Promise<void>;
+  onWalletSelectCancelled: () => void;
+};
+
+export type InternalWalletProvider = typeof internalWalletProvider;
+
+export function isInternalProvider(
+  provider: WormholeConnectWalletProvider,
+): provider is InternalWalletProvider {
+  return 'isInternal' in provider && provider.isInternal === true;
```
priyajeet: `Object.hasOwn(provider, "isInternal")`

src/utils/wallet/InternalWalletProvider.ts
```diff
@@ -0,0 +1,280 @@
...
+function swapWallets(): boolean {
+  const temp = walletConnections.sending;
+  walletConnections.sending = walletConnections.receiving;
+  walletConnections.receiving = temp;
+
+  if (walletConnections.sending?.wallet.getName() === 'ReadOnlyWallet') {
+    walletConnections.sending.wallet.disconnect();
+  }
+
+  return true;
```
priyajeet: Would this function ever return anything but true? If not, should it even return anything at all?

src/views/v3/Bridge/SwapInputs/index.tsx
```diff
@@ -77,10 +78,11 @@ function SwapInputs() {
       val === 'spinRight' ? 'spinLeft' : 'spinRight',
     );
 
-    dispatch(swapInputs());
-    dispatch(swapWallets());
-    dispatch(setAmount(''));
-  }, [canSwap, isTransactionInProgress, dispatch]);
+    if (swapWallets()) {
```
priyajeet: Should this be if()ed, question same as another comment

src/utils/wallet/types.ts
```diff
@@ -1,8 +1,98 @@
...
+   * Get an already connected wallet without requiring user interaction.
+   * This is a passive method that only returns existing connections.
+   *
+   * @param chain - The chain to check for an existing connection
+   * @param type - Whether to check for sending or receiving wallet
+   * @returns The connected wallet, or null if no wallet is currently connected
+   */
+  getWallet(chain: Chain, type: TransferWallet): Wallet | null;
+
+  signAndSendTransaction(
```
priyajeet, line 81: Thinking aloud, should we (can we?) move this to `Wallet` interface?

In which case, `WormholeConnectWalletProvider` only exposes `{ connect, disconnect, swap, get, on, off }`.

Then the `useWalletProvider()` hook and `WalletContextProvider` do not need to return the `walletProvider` object but instead just those 4 wallet-lifecycle functions + on/off.

signer.ts, will just use `wallet.signAndSendTransaction()` and this function need not take a chain or wallet input.

src/views/v3/Bridge/WalletConnector/Sidebar.tsx
kev1n-peters: I think it's fine for onClose() to be called on success and error cases for now (this is how it behaves in production)

src/utils/wallet/types.ts
kev1n-peters: I originally had it on the wallet interface, but the problem is that Connect doesn’t use the `signAndSendTransaction` method from the wallet aggregator wallets. Instead, Connect uses its own `signAndSendTransaction` function, which contains special logic for EVM and SVM that isn’t present in the wallet aggregator wallets’ methods. I like your proposal, but we would either need to move the special logic from Connect into the wallet aggregator wallets (which I think is out of scope for this PR since I don't think it would be a straight "copy paste" refactor) or create a new class that extends the wallet aggregator wallet class and overrides the signAndSendTransaction method. Wdyt? Is it fine to leave this for now and revisit later?

src/views/v3/Bridge/SwapInputs/index.tsx
kev1n-peters: removed the if

src/utils/wallet/InternalWalletProvider.ts
kev1n-peters: Our internal provider would probably never return anything besides true

src/utils/wallet/InternalWalletProvider.ts
kev1n-peters: changed, but needed to bump the tsconfig.json target from `es2021` to `es2022` to support it

src/views/v3/Bridge/WalletConnector/Sidebar.tsx
kev1n-peters: We can get it from the hook -- changed

src/views/v3/Bridge/WalletConnector/Sidebar.tsx
kev1n-peters: done

src/utils/wallet/types.ts
priyajeet: We can revisit when we integrate a WaaS provider. We would need some kind of abstraction layer in our new portal provider since it will have to choose between WaaS or fallback to wallet aggregator, depending upon the chain/network.

priyajeet (dismissed):

priyajeet (dismissed):

priyajeet (approved):

PR #3724, opened 2025-08-14
chore: Add manual tBTC to V2 route

kyleleighton:
The manual tBTC route was already showing on v3 100% of the time. This ensures the same behavior for v2.


kyleleighton:
Duplicate

PR #3730, opened 2025-08-15
feat: pull in Mayan route sdk

bradleyplaydon:
### Update
 - Pull the @mayanfinance/wormhole-sdk-route in


src/routes/mayan/evm/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+      return amountIn;
+    } else {
+      return 0n;
+    }
+  }
+
+  return { encodeFunctionData, getMsgValue };
+}
+
+function useMayanForwarderShim(
```
priyajeet, line 64: When removing the `isNewEvmReferralEnabled` flag, just rename the function to something like `isMayanForwarderShimUsed()` or something better. Since its not really a hook, but the linter thinks it is due to the `use` prefix.

priyajeet (approved):

src/routes/mayan/evm/utils.ts
bradleyplaydon: For now keeping everything as is, on my refactor branch i've changed this to getMayanForwarderShim, can change to getIsMayanForwarderShim since it returns boolean?

bradleyplaydon: For now keeping everything as is, on my refactor branch i've changed this to getMayanForwarderShim, can change to getIsMayanForwarderShim since it returns boolean?

PR #3740, opened 2025-08-22
fix: resolve repeated USDC output amount warning

kyleleighton:
This PR does 3 things:

1. Fixes an issue where all CCTP routes would show the high gas alert as long as one route had it
2. Changes the gas alert threshold to 10% rather than a fixed $20
3. Alerts the user with a more specific message that the difference is greater than 10%

Screenshot 2025-08-22 at 1 15 41 PM
Screenshot 2025-08-22 at 1 16 09 PM

https://linear.app/wormholelabs/issue/PROD-488/investigate-cctp-routes-warning-due-to-high-ethereum-gas-fees

### DIFF
--- a/src/views/v2/Bridge/Routes/SingleRoute.tsx
+++ b/src/views/v2/Bridge/Routes/SingleRoute.tsx
@@ -35,7 +35,7 @@ import { formatWithCommas } from 'utils/formatNumber';
 import TimeToDestination from './TimeToDestination';
 import Provider from './Provider';
 
-const HIGH_FEE_THRESHOLD = 20; // dollhairs
+const HIGH_FEE_PERCENT = 10;
 
 type Props = {
   route: string;
@@ -120,6 +120,7 @@ const SingleRoute = (props: Props) => {
   const {
     toChain: destChain,
     fromChain: sourceChain,
+    amount: inputAmount,
     isTransactionInProgress,
   } = useSelector((state: RootState) => state.transferInput);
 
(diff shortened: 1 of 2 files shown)


src/views/v2/Bridge/Routes/SingleRoute.tsx
```diff
@@ -137,20 +138,31 @@ const SingleRoute = (props: Props) => {
...
+    );
+    const outputAmount = amount.whole(quote.destinationToken.amount);
+    const outputUsd = calculateUSDPriceRaw(
+      getTokenPrice,
+      outputAmount,
+      destToken,
+    );
 
-    if (feePrice === undefined) {
+    if (inputUsd === undefined || outputUsd === undefined || inputUsd <= 0) {
```
priyajeet, line 157: inputUsd <= 0 should cover the undefined case

src/views/v3/Bridge/Routes/SingleRoute.tsx
```diff
@@ -113,21 +114,32 @@ const SingleRoute = (props: Props) => {
...
-      return false;
+    if (inputUsd === undefined || outputUsd === undefined || inputUsd <= 0) {
+      return { isHighFee: false };
     }
 
-    return feePrice > HIGH_FEE_THRESHOLD;
-  }, [getTokenPrice, quote?.relayFee]);
+    const delta = Math.max(0, inputUsd - outputUsd);
+    const pct = delta / inputUsd;
+    return { isHighFee: pct >= HIGH_FEE_PCT };
```
priyajeet: Do we need to return an object? Or is this for future to return some other thresholds?

priyajeet (dismissed):

src/views/v2/Bridge/Routes/SingleRoute.tsx
kyleleighton: `inputUsd <= 0` won’t catch `undefined` (it evaluates to `false`), so I believe we still need `inputUsd === undefined`.

priyajeet: ah yes

priyajeet (approved):

PR #3741, opened 2025-08-23
fix: Disallow unsupported routes from popping up

priyajeet:
when using same chain swaps

### DIFF
--- a/src/hooks/useFetchSupportedRoutes.ts
+++ b/src/hooks/useFetchSupportedRoutes.ts
@@ -81,15 +81,6 @@ const useFetchSupportedRoutes = ({
           );
         }
 
-        // HAX - Enable Mayan routes (except SHUTTLE) for all assets
-        // TODO token refactor
-        if (
-          route.rc.name.includes('Mayan') &&
-          route.rc.name !== 'MayanRouteSHUTTLE'
-        ) {
-          supported = true;
-        }
-
         if (supported) {
           _routes.push(name);
         }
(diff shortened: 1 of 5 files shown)


src/hooks/useFetchSupportedRoutes.ts
```diff
@@ -81,15 +81,6 @@ const useFetchSupportedRoutes = ({
           );
         }
 
-        // HAX - Enable Mayan routes (except SHUTTLE) for all assets
-        // TODO token refactor
-        if (
```
priyajeet, line 86: This should ideally just work with Line 94 in the file (sdkv2/route.ts) below.
@emreboga any gotchas you can think of?

src/views/v2/Bridge/Routes/SingleRoute.tsx
```diff
@@ -343,7 +343,7 @@ const SingleRoute = (props: Props) => {
                 fontSize={14}
                 lineHeight="18px"
               >
-                {`Price impact is greater than ${HIGH_FEE_PERCENT}%.`}
+                {`Price impact is greater than ${HIGH_FEE_PERCENT}%`}
```
priyajeet, line 346: Removed the period `.` from the header/title for this section. The description/body keeps it.

kev1n-peters (approved):
Seems good to me. Should wait for Emre to address the comment before merging

src/hooks/useFetchSupportedRoutes.ts
emreboga: This was a shortcut return for Mayan routes here, but since we have the same logic in isRouteSupported function, I think it should be fine.

emreboga (approved):

PR #3746, opened 2025-08-26
[Feat]: Add XRPL mainnet  & bump NTT sdk to 2.0.3

bradleyplaydon:
- Add XRPL Mainnet to config
- Bump NTT sdk to 2.0.3


priyajeet (dismissed):

src/config/testnet/tokens.ts
```diff
@@ -413,4 +413,22 @@ export const TESTNET_TOKENS: TokenConfig[] = [
     icon: TokenIcon.PLUME,
     tokenId: { chain: 'Plume', address: 'native' },
   },
+  {
+    symbol: 'XRP',
+    tokenId: {
+      chain: 'XRPLEVM',
+      address: 'native',
```
kev1n-peters: The `XRP` token is an ERC20 token which has an address and is also the gas token. Since it has an address, we should change the address here to `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` and remove the `WXRP` entry. Similar to Celo (CELO is an ERC20 token and also the gas token), we should add a case here to return `XRP` when the chain is `WXRP`: https://github.com/wormhole-foundation/wormhole-connect/blob/612b290bacd52f55aad8ddcf698fa479b6fb3f07/src/config/tokens.ts#L353

priyajeet (approved):

PR #3747, opened 2025-08-27
Fix destination token search and filtering

kyleleighton:
Problem
  - Users couldn't paste contract addresses to search for destination tokens
  - Source token wasn't appearing as option on destination chain

---

  Solution
  - Consolidated v2/v3 token list logic into shared `useTokenListWithSearch` hook
  - Fixed balance filtering - only filter source tokens, never destination tokens
  - Updated Mayan route to return all tokens on destination chain
  - Refactored `useComputeDestinationTokens` with memoization for better performance

---

  Result
  - Can paste any contract address in destination field
  - Can select tokens not owned on destination chain
  -  Source token appears in destination options when available
  
Screenshot 2025-08-27 at 11 53 40 AM
Screenshot 2025-08-27 at 11 53 51 AM
Screenshot 2025-08-28 at 3 35 37 PM


src/routes/utils/tokenHelpers.ts
```diff
@@ -0,0 +1,10 @@
+import type { Chain, TokenId } from '@wormhole-foundation/sdk';
+import config from 'config';
+
+/**
+ * Returns all token IDs for a given chain.
+ * Used by routes that support swapping between any tokens on a chain.
+ */
+export const getAllTokenIdsForChain = (chain: Chain): TokenId[] => {
```
priyajeet: Should we move this file under src/utils? There is a tokens file in there also that can host this maybe?

src/hooks/useComputeDestinationTokens.ts
```diff
@@ -13,95 +14,119 @@ type Props = {
...
+  }
+
+  // Both chains selected - fetch supported tokens from routes
+  const supportedIds = await fetchSupportedDestTokenIds(
+    sourceToken?.key,
+    sourceChain,
+    destChain,
+  );
+
+  const supported: Token[] = [];
```
bradleyplaydon: pref: supportedTokens

src/hooks/useComputeDestinationTokens.ts
```diff
@@ -13,95 +14,119 @@ type Props = {
...
   isFetching: boolean;
 };
 
+/**
+ * Fetch supported destination token IDs from routes.
+ * Memoized to avoid redundant API calls.
+ */
+const fetchSupportedDestTokenIds = memoize(
+  async (
+    sourceTokenKey: string | undefined,
```
bradleyplaydon: Is this the source token address?

src/hooks/useComputeDestinationTokens.ts
```diff
@@ -13,95 +14,119 @@ type Props = {
...
+  const supportedIds = await fetchSupportedDestTokenIds(
+    sourceToken?.key,
+    sourceChain,
+    destChain,
+  );
+
+  const supported: Token[] = [];
+  await Promise.all(
+    supportedIds.map(async (tokenId) => {
+      const t = await getOrFetchToken(tokenId);
```
bradleyplaydon: pref: tokenPromise

src/routes/mayan/index.ts
```diff
@@ -186,9 +180,9 @@ class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
   static async supportedDestinationTokens<N extends Network>(
     _token: TokenId,
     _fromChain: ChainContext<N>,
-    _toChain: ChainContext<N>,
+    toChain: ChainContext<N>,
```
bradleyplaydon, line 183: is the _ supposed to of been dropped?

src/routes/sdkv2/route.ts
```diff
@@ -87,14 +86,6 @@ export class SDKv2Route {
       return false;
     }
 
-    const isMayan = name.includes('Mayan');
```
priyajeet, line 90: Since we are removing this, wondering if we can fast-memoize the supportedDestTokens() function below. That function takes only primitives as input, so might be a good memoization candidate and will prevent running those loops unnecessarily. There are loops in there for filtering, and then this function also does a loop using find() below.

src/routes/mayan/index.ts
```diff
@@ -186,9 +180,9 @@ class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
...
   ): Promise<TokenId[]> {
-    return [];
+    return getAllTokenIdsForChain(toChain.chain);
```
priyajeet, line 186: Can you add a TODO comment here for reference that we need to update this function to only allow HyperCore/USDC when we come to implementing that.

src/hooks/useComputeDestinationTokens.ts
```diff
@@ -13,95 +14,119 @@ type Props = {
...
 type ReturnProps = {
   supportedDestTokens: Token[];
   isFetching: boolean;
 };
 
+/**
+ * Fetch supported destination token IDs from routes.
+ * Memoized to avoid redundant API calls.
+ */
+const fetchSupportedDestTokenIds = memoize(
```
priyajeet: Since we are inside an hook, and only being referenced in here, might as well use React.useMemo() since fast-memoize is only useful outside hooks.

src/hooks/useComputeDestinationTokens.ts
```diff
@@ -13,95 +14,119 @@ type Props = {
...
+      : undefined;
+
+    try {
+      return await config.routes.allSupportedDestTokens(
+        sourceToken,
+        sourceChain,
+        destChain,
+      );
+    } catch (e) {
+      console.error(e);
```
priyajeet: Remove console.error

src/hooks/useTokenListWithSearch.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    if (searchQuery) {
+      try {
+        const address = toNative(chain, searchQuery);
+
+        if (address) {
+          const existing = config.tokens.get(chain, searchQuery);
+
+          const addTokenIfNotExists = (token: Token) => {
```
bradleyplaydon: Worth moving outside the useEffect? Can't see why it needs to be inside it and would mean the functions not redefined everytime a dep changes

src/hooks/useTokenListWithSearch.ts
```diff
@@ -0,0 +1,150 @@
...
+            });
+          } else {
+            addTokenIfNotExists(existing);
+          }
+        }
+      } catch (_e) {
+        // Failed to parse as address - expected behavior
+      }
+    } else {
+      setSearchedTokens([]);
```
bradleyplaydon: Can't this condition go at the top condition on lines 54

src/hooks/useComputeDestinationTokens.ts
priyajeet: OR we can move the non-hook helper functions along with the other tokens helper you have below.

src/hooks/useTokenListWithSearch.ts
```diff
@@ -0,0 +1,150 @@
...
+    walletAddress,
+  ]);
+
+  // Get token prices for all tokens
+  const allTokens = useMemo(
+    () => [...baseTokenList, ...searchedTokens],
+    [baseTokenList, searchedTokens],
+  );
+
+  useEffect(() => {
```
bradleyplaydon: Could this be memoized instead?

```
const tokenPrices = useMemo(() => getTokenPrices(allTokens), [allTokens, getTokenPrices, lastTokenPriceUpdate]);
```

src/hooks/useComputeDestinationTokens.ts
```diff
@@ -13,95 +14,119 @@ type Props = {
...
+    }
+  },
+);
+
+/**
+ * Compute the destination tokens for a given source token and chains.
+ * This handles three cases:
+ * 1. No source chain selected - returns all tokens on destination chain
+ * 2. Both chains selected with source token - fetches supported destination tokens from routes
+ * 3. Both chains selected without source token - returns empty array
```
priyajeet, line 28: In this logic we should also make sure that if source and destination chain are the same, the source token is never included in destination

src/routes/mayan/index.ts
kyleleighton: Yes. We use it now. The `_` is for unused.

src/hooks/useComputeDestinationTokens.ts
priyajeet: This is the code that seems to add tokens to the config, can there be a race condition of this adding to the config and we reading config.tokens in other calls?

src/hooks/useComputeDestinationTokens.ts
```diff
@@ -13,95 +14,119 @@ type Props = {
...
+
+  // Both chains selected - fetch supported tokens from routes
+  const supportedTokenIds = await fetchSupportedDestTokenIds(
+    sourceToken?.key,
+    sourceChain,
+    destChain,
+  );
+
+  const supportedTokens: Token[] = [];
+  await Promise.all(
```
priyajeet: Wondering if this can be simplified to something like

```ts
const tokens = await Promise.all(supportedIds.map(id) => getOrFetchToken(tokenId));
const supported = tokens.filter(token => !!token);
```

src/hooks/useComputeDestinationTokens.ts
priyajeet: Though maybe thats already taken care of downstream inside route.supportedDestTokens()'s filter function

src/hooks/useComputeDestinationTokens.ts
```diff
@@ -13,95 +14,119 @@ type Props = {
...
+ * Memoized to avoid redundant API calls.
+ */
+const fetchSupportedDestTokenIds = memoize(
+  async (
+    sourceTokenKey: string | undefined,
+    sourceChain: Chain,
+    destChain: Chain,
+  ): Promise<TokenId[]> => {
+    // Find the actual token from the key
+    const sourceToken = sourceTokenKey
```
priyajeet: Can't we just pass the sourceToken from parent caller? instead of this lookup

src/hooks/useComputeDestinationTokens.ts
```diff
@@ -13,95 +14,119 @@ type Props = {
...
 
   const dispatch = useDispatch();
+  const { getOrFetchToken, lastTokenCacheUpdate } = useTokens();
 
   const [supportedDestTokens, setSupportedDestTokens] = useState<Token[]>([]);
   const [isFetching, setIsFetching] = useState(false);
 
-  const { getOrFetchToken, lastTokenCacheUpdate } = useTokens();
+  const computeDestTokens = useCallback(async () => {
+    setSupportedDestTokens([]);
```
priyajeet, line 69: If we remove this line, does it fix the issue where going from chain to chain, the asset picker flashes in size because it makes the list empty and then fills it again?

As opposed to in-place switch. Though this might also mean that when isFetching is true, that list should not be clickable. Aka list disabled.

src/hooks/useComputeDestinationTokens.ts
kyleleighton: No. `sourceTokenKey` is not the address. It's a JSON string containing [chain, address] from [this token file](https://github.com/wormhole-foundation/wormhole-connect/blob/d52e96b22c1b63370eeb8a2f6c7a91d21f4eba3c/src/config/tokens.ts#L49)

bradleyplaydon: Makes sense!

src/hooks/useComputeDestinationTokens.ts
kyleleighton: It seems to have no impact on that.

src/hooks/useTokenListWithSearch.ts
```diff
@@ -0,0 +1,148 @@
...
+    });
+  }, []);
+
+  useEffect(() => {
+    if (!chain || !tokenPastingEnabled || !searchQuery) {
+      setSearchedTokens([]);
+      return;
+    }
+
+    if (searchQuery) {
```
priyajeet: Can remove since its checked on line 57

src/hooks/useTokenListWithSearch.ts
```diff
@@ -0,0 +1,148 @@
...
+    chain,
+    getOrFetchToken,
+    tokenPastingEnabled,
+    addTokenIfNotExists,
+  ]);
+
+  const sortedTokens = useMemo(() => {
+    const mergedTokens = [...baseTokenList];
+
+    // Add searched tokens that aren't already in the list
```
priyajeet: We have es-toolkit available in this repo, see if unionBy/intersectionBy helps here
https://es-toolkit.dev/reference/array/unionBy.html
A union followed by filter might be linear time.

src/hooks/useTokenListWithSearch.ts
```diff
@@ -0,0 +1,148 @@
...
+        } else {
+          // For destination tokens or when searching, add all searched tokens
+          mergedTokens.push(searchedToken);
+        }
+      }
+    }
+
+    // For destination token list in same-chain swaps, filter out the source token
+    if (!isSource && isSameChainSwap && sourceToken) {
+      return mergedTokens.filter(
```
priyajeet: Might be able to combine this filtering with balance filtering

src/hooks/useTokenListWithSearch.ts
```diff
@@ -0,0 +1,148 @@
...
+            getOrFetchToken({ chain, address }).then((fetchedToken) => {
+              if (fetchedToken) {
+                addTokenIfNotExists(fetchedToken);
+              }
+            });
+          } else {
+            addTokenIfNotExists(existing);
+          }
+        }
+      } catch (_e) {
```
priyajeet: can remove `(_e)` part and it will work

src/hooks/useTokenListWithSearch.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    if (searchQuery) {
+      try {
+        const address = toNative(chain, searchQuery);
+
+        if (address) {
+          const existing = config.tokens.get(chain, searchQuery);
+
+          if (!existing) {
+            getOrFetchToken({ chain, address }).then((fetchedToken) => {
```
priyajeet: Maybe dump a comment here that we aren't waiting on this promise

src/hooks/useTokenListWithSearch.ts
```diff
@@ -0,0 +1,148 @@
...
+  balances,
+  walletAddress,
+  tokenPastingEnabled = true,
+}: UseTokenListWithSearchParams): UseTokenListWithSearchReturn => {
+  const [searchedTokens, setSearchedTokens] = useState<Token[]>([]);
+  const { getOrFetchToken, getTokenPrices } = useTokens();
+
+  const addTokenIfNotExists = useCallback((token: Token) => {
+    setSearchedTokens((prev) => {
+      const alreadyExists = prev.some((t) => isSameToken(t, token));
```
priyajeet: Do we need to do existence check? Can we just just concat and dedupe on return in the filtering we are doing? Might prevent loop here.

src/hooks/useTokenListWithSearch.ts
```diff
@@ -0,0 +1,148 @@
...
+          } else {
+            addTokenIfNotExists(existing);
+          }
+        }
+      } catch (_e) {
+        // Failed to parse as address - expected behavior
+      }
+    }
+  }, [
+    searchQuery,
```
priyajeet: Does this need to be debounced? es-toolkit has the utility

priyajeet: There is also the native useDeferredValue

src/utils/tokenHelpers.ts
```diff
@@ -0,0 +1,10 @@
+import type { Chain, TokenId } from '@wormhole-foundation/sdk';
+import config from 'config';
+
+/**
+ * Returns all token IDs for a given chain.
+ * Used by routes that support swapping between any tokens on a chain.
+ */
+export const getAllTokenIdsForChain = (chain: Chain): TokenId[] => {
```
priyajeet: Should we fast-memoize this?

priyajeet (dismissed):

src/views/v2/Bridge/AssetPicker/TokenItem.tsx
```diff
@@ -153,16 +154,25 @@ function TokenItem(props: TokenItemProps) {
           {props.isFetchingBalance && props.balance === null ? (
```
priyajeet: I think this is definitely something we can extract out into its own components `<TokenPrice>` and `<TokenBalance>`. With nested ?:?: hard to understand whats really happening.

But we can leave it as-is in the v2 folder, just do it for v3

src/views/v3/Bridge/AssetPicker/TokenItem.tsx
```diff
@@ -148,22 +148,12 @@ function TokenItem(props: TokenItemProps) {
           </Box>
         </div>
       </Box>
-      <Stack alignItems="flex-end">
```
priyajeet, line 151: 😇

priyajeet (approved):

PR #3748, opened 2025-08-27
chore: sort routes by speed/ETA rather than price

kyleleighton:
- Sort routes by pure ETA instead of bucketing routes under 60 seconds together
  - Ensures faster routes always rank higher regardless of price differences

### DIFF
--- a/src/hooks/useSortedRoutesWithQuotes.ts
+++ b/src/hooks/useSortedRoutesWithQuotes.ts
@@ -102,48 +102,44 @@ export const useSortedRoutesWithQuotes = ({
 
   // Only routes with quotes are sorted.
   const sortedRoutesWithQuotes = useMemo(() => {
-    return bucketByEta(routesWithQuotes)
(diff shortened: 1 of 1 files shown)


src/hooks/useSortedRoutesWithQuotes.ts
```diff
@@ -102,48 +102,44 @@ export const useSortedRoutesWithQuotes = ({
...
-          // when the preferred route gets autoselected
-          if (preferredRouteName) {
-            if (routeA.route === preferredRouteName) {
-              return -1;
-            } else if (routeB.route === preferredRouteName) {
-              return 1;
-            }
-          }
+      // 1. Sort by ETA (fastest first)
+      const etaA = routeA.quote.eta ?? Infinity;
```
priyajeet, line 120: If ETA is not there, it can also mean instant. For example same chain swaps is considered instant.

priyajeet (approved):

PR #3754, opened 2025-08-29
feat: add new event type to WormholeConnectEventCore

bradleyplaydon:
**Introduces breaking changes**:
- Updated defaultInput prop from 'fromChain, toChain, fromToken, toToken' to 'source' and 'destination'

**Updates**
- Added a new event type to WormholeConnectEventCore to support remembering the last chain and token the user selected so if they refresh it persists
- Fix condition which inside validateDefaults as it no longer stands because we support same chain swaps

### DIFF
--- a/src/config/utils.ts
+++ b/src/config/utils.ts
@@ -153,10 +153,11 @@ export const validateDefaults = (
       delete defaults.destination;
     }
   }
-  if (defaults.source?.chain && defaults.destination?.chain) {
-    if (defaults.source.chain === defaults.destination.chain) {
+
+  if (defaults.source?.token && defaults.destination?.token) {
+    if (defaults.source.token === defaults.destination.token) {
       error(
-        `Source and destination chain cannot be the same, check the defaultInputs configuration`,
+        `Source and destination token cannot be the same, check the defaultInputs configuration`,
       );
     }
   }
--- a/src/config/utils.ts
+++ b/src/config/utils.ts
@@ -135,33 +135,37 @@ export const validateDefaults = (
   tokens: TokenCache,
 ) => {
   if (!defaults) return;
-  if (defaults.from?.chain) {
-    const chain = chains[defaults.from.chain];
+  if (defaults.source?.chain) {
+    const chain = chains[defaults.source.chain];
     if (!chain) {
       error(
-        `Invalid chain name "${defaults.from.chain}" specified for defaultInputs.fromChain`,
+        `Invalid chain name "${defaults.source.chain}" specified for defaultInputs.source.chain`,
       );
-      delete defaults.from;
+      delete defaults.source;
     }
   }
-  if (defaults.to?.chain) {
-    const chain = chains[defaults.to.chain];
+  if (defaults.destination?.chain) {
+    const chain = chains[defaults.destination.chain];
     if (!chain) {
       error(
-        `Invalid chain name "${defaults.to.chain}" specified for defaultInputs.toChain`,
+        `Invalid chain name "${defaults.destination.chain}" specified for defaultInputs.destination.chain`,
       );
-      delete defaults.to;
+      delete defaults.destination;
     }
   }
-  if (defaults.from?.chain && defaults.to?.chain) {
-    if (defaults.from.chain === defaults.to.chain) {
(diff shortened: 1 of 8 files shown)


src/telemetry/utils.ts
```diff
@@ -0,0 +1,35 @@
+import type { InternalConfig } from 'config';
+import type { Chain, Token } from 'exports';
+import { UserActions } from 'exports';
+
+export const handleSelectChain = (
+  chain: Chain,
+  isSource: boolean,
+  config: InternalConfig<'Mainnet' | 'Testnet' | 'Devnet'>,
```
kev1n-peters: Can these just be `InternalConfig<Network>`?

priyajeet: Also can we just import this? instead of passing in

src/views/v2/Bridge/AssetPicker/index.tsx
```diff
@@ -26,6 +26,7 @@ import TokenList from './TokenList';
 import AssetBadge from 'components/AssetBadge';
 import type { Token } from 'config/tokens';
 import { useTokenList } from 'hooks/useTokenList';
+import { handleSelectChain, handleSelectToken } from 'telemetry/utils';
```
priyajeet: Need to update the v3 (views/v3) folder also

priyajeet: I won't even bother changing V2 at this point if it makes life simpler for new features

src/telemetry/utils.ts
bradleyplaydon: > Also can we just import this? 

Wouldn't this be counter intuitive since it's already imported in the component and being used for other things not just telemetry like the useTokenList hook

src/telemetry/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+    details: {
+      value: chain,
+      action: isSource
+        ? UserActions.SelectSrcChain
+        : UserActions.SelectDestChain,
+    },
+  });
+};
+
+export const handleSelectToken = (
```
priyajeet: Should we also rename the exported functions to indicate they are telemetry related? Since these words are common and can also be used in other places with same meaning. 

Maybe something along the lines of `handleTelemetryOnChainSelect` and `handleTelemetryOnTokenSelect`. Something like this will likely avoid casting to a new name if there is a conflict + person reading them will not need to see where its imported from to understand usecase/functionality

src/telemetry/utils.ts
bradleyplaydon: > > Also can we just import this?
> 
> Wouldn't this be counter intuitive since it's already imported in the component and being used for other things not just telemetry like the useTokenList hook

Unless I do a dynamic import based upon if the param is passed and make it optional?

src/telemetry/utils.ts
bradleyplaydon: Love this, the more explicit the better imo

src/telemetry/utils.ts
bradleyplaydon: For now leaving as it, think if we run into multiple cases where configs having to be imported into the component that imports these functions we can import it here

src/telemetry/utils.ts
```diff
@@ -0,0 +1,33 @@
+import config from 'config';
+import type { Chain, Token } from 'exports';
```
priyajeet: Lets fix these imports. Should come from the actual file next to this file using relative import and not exports module. Chain/Token should come from wormhole sdk or wherever?

src/config/utils.ts
```diff
@@ -135,72 +135,76 @@ export const validateDefaults = (
...
+        `Invalid chain name "${defaults.destination.chain}" specified for defaultInputs.destination.chain`,
       );
-      delete defaults.fromChain;
+      delete defaults.destination;
     }
   }
-  if (defaults.fromChain && defaults.toChain) {
-    if (defaults.fromChain === defaults.toChain) {
+  if (defaults.source?.chain && defaults.destination?.chain) {
+    if (defaults.source.chain === defaults.destination.chain) {
```
priyajeet: change check to token check, not chain check. Chains can be same

priyajeet (approved):

PR #3755, opened 2025-08-29
feat: Change Asset picker layout LTR to align with Monad designs

emreboga:
Other than adding back the balance information to the UI, this will not change any behavior of the Asset Picker, but the layout of the components within.

Screenshot 2025-08-29 at 6 00 32 PM

### DIFF
--- a/src/views/v3/Bridge/AssetPicker/index.tsx
+++ b/src/views/v3/Bridge/AssetPicker/index.tsx
@@ -95,6 +95,28 @@ function AssetPicker(props: Props) {
     return null;
   }, [props.isSource, props.balances, props.token]);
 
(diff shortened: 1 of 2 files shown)


bradleyplaydon (approved):

src/views/v3/Bridge/AssetPicker/index.tsx
```diff
@@ -95,6 +95,28 @@ function AssetPicker(props: Props) {
     return null;
   }, [props.isSource, props.balances, props.token]);
 
+  const tokenBalanceDisplay = useMemo(() => {
```
priyajeet, line 98: This should have been its own file/component for future reference. And no need to memoize this.

PR #3767, opened 2025-09-03
fix: UI polish per product feedback

emreboga:
Before:
Screenshot 2025-09-03 at 8 23 13 AM

After:
Screenshot 2025-09-03 at 8 22 07 AM

### DIFF
--- a/src/views/v3/Bridge/AssetPicker/index.tsx
+++ b/src/views/v3/Bridge/AssetPicker/index.tsx
@@ -100,7 +100,7 @@ function AssetPicker(props: Props) {
       return null;
     }
     const displayValue = `${sdkAmount.display(tokenBalance)} ${
-      props.token?.symbol || ''
+      props.token ? getTokenSymbol(props.token) : ''
     }`;
     return (
       <Typography
@@ -118,7 +118,7 @@ function AssetPicker(props: Props) {
         {displayValue}
       </Typography>
     );
-  }, [props.token?.symbol, theme.palette.text.secondary, tokenBalance]);
+  }, [props.token, theme.palette.text.secondary, tokenBalance]);
 
   // Side-effect to reset chain search visibility.
   // Popover and drawer close has an animation, which requires to wait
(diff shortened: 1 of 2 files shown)


src/views/v3/Bridge/AssetPicker/index.tsx
```diff
@@ -523,11 +538,7 @@ function AssetPicker(props: Props) {
             gap: '8px',
           }}
         >
-          {props.isSource ? (
-            <Box>{tokenBalanceDisplay}</Box>
-          ) : (
-            <Box>{destTokenUnitPrice}</Box>
-          )}
+          {<Box>{!props.isSource && destTokenUnitPrice}</Box>}
```
kyleleighton, line 541: Should we move the check outside of the `<Box>` so that we don't render an empty `Box`?

src/views/v3/Bridge/AssetPicker/index.tsx
```diff
@@ -99,23 +99,26 @@ function AssetPicker(props: Props) {
     if (!tokenBalance) {
       return null;
     }
-    const displayValue = `Balance: ${sdkAmount.display(tokenBalance)}`;
+    const displayValue = `${sdkAmount.display(tokenBalance)} ${
+      props.token?.symbol || ''
```
kyleleighton: Can you use the new [getTokenSymbol](https://github.com/wormhole-foundation/wormhole-connect/blob/2100ffee68c8a79d8a6e3f4f7ac8a7360eaf017a/src/utils/index.ts#L72) here? This accounts for token naming/symbol overrides (which currently only includes WSOL -> SOL)

src/views/v3/Bridge/AssetPicker/index.tsx
emreboga: If we do so, the parent's "space-between" styling will push the content to left for source asset picker, but we want to the right in both source and destination pickers. We could either have the check in styles or have the box render regardless. Both is fine for me.

src/views/v3/Bridge/AssetPicker/index.tsx
emreboga: Sure!

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3767/commits/ee44bfd41530a6b84d9fa22dec070ae9d8cdfede

kyleleighton (approved):

PR #3770, opened 2025-09-04
feat: group assets in selector by owned vs all

kyleleighton:
1. When USD balances are < $0.01, they showed as $0.00. Now they show as < $0.01
2. All assets now show in the source selector, divided by owned versus all, with all dimmed and your owned ones undimmed.
3. Removes the 'Select a Token' and 'Tokens on X' text for cleanliness
4. Changes 'Choose a network' with 'Choose chain'

Screenshot 2025-09-03 at 7 13 17 PM
Screenshot 2025-09-03 at 7 13 23 PM

https://linear.app/wormholelabs/issue/PROD-611/possibly-show-all-tokens-in-the-source-asset-picker-when-connected-to

https://linear.app/wormholelabs/issue/PROD-612/possibly-show-something-other-than-dollar0-as-the-token-balance-amount

### DIFF
--- a/src/hooks/useTokenListGrouping.ts
+++ b/src/hooks/useTokenListGrouping.ts
@@ -1,0 +1,47 @@
+import { useMemo } from 'react';
+import type { Token } from 'config/tokens';
+import { amount as sdkAmount } from '@wormhole-foundation/sdk';
+import type { Balances } from 'utils/wallet';
+
+export function useTokenListGrouping({
+  sortedTokens,
+  isGroupingEnabled,
+  isWalletConnected,
+  balances,
+}: {
+  sortedTokens: Token[];
+  isWalletConnected: boolean;
+  isGroupingEnabled: boolean;
+  balances: Balances;
+}): { listItems: Token[]; ownedCount: number } {
+  return useMemo(() => {
+    if (!isGroupingEnabled) {
+      return { listItems: sortedTokens, ownedCount: 0 };
+    }
+
+    const getHasPositiveBalance = (t: Token) => {
+      if (!isWalletConnected) return false;
+      const bal = balances?.[t.key]?.balance;
+      return bal != null && sdkAmount.units(bal) > 0n;
+    };
+
+    const native = sortedTokens.find((t) => t.isNativeGasToken);
+    const nativeKey = native?.key;
+
+    const ownedTokens = sortedTokens.filter(getHasPositiveBalance);
+    const ownedKeys = new Set(ownedTokens.map((t) => t.key));
+
+    const listItems = [
+      // 1) Owned
+      ...ownedTokens,
+      // 2) Native (only if not owned)
+      ...(native && nativeKey && !ownedKeys.has(nativeKey) ? [native] : []),
+      // 3) The rest
+      ...sortedTokens.filter(
+        (t) => !ownedKeys.has(t.key) && t.key !== nativeKey,
+      ),
+    ];
+
+    return { listItems, ownedCount: ownedTokens.length };
+  }, [sortedTokens, isGroupingEnabled, isWalletConnected, balances]);
+}
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -203,6 +203,10 @@ export const getUSDFormat = (price: number | undefined): string => {
     return '$0';
   }
 
+  if (price > 0 && price < 0.01) {
+    return '<$0.01';
+  }
+
   return Intl.NumberFormat('en-EN', {
     style: 'currency',
     currency: 'USD',
--- a/src/views/v3/Bridge/AssetPicker/TokenItem.tsx
+++ b/src/views/v3/Bridge/AssetPicker/TokenItem.tsx
@@ -27,6 +27,7 @@ type TokenItemProps = {
   isSelected?: boolean;
   isFetchingBalance?: boolean;
   isSource?: boolean;
+  dimmed?: boolean;
 };
 
 function TokenItem(props: TokenItemProps) {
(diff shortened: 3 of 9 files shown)


src/utils/index.ts
```diff
@@ -203,6 +203,10 @@ export const getUSDFormat = (price: number | undefined): string => {
     return '$0';
   }
 
+  if (price > 0 && price < 0.01) {
+    return '<$0.01';
```
priyajeet, line 207: It probably doesn't matter for now since its all USD....but I would say still use the Intl.NumberFormat here for the 0.01, can prefix it with <$.

Passing in { maximumFractionDigits: n } to NumberFormat can help.

bradleyplaydon, line 207: Should we not also be using Intl like on lines 210 - 213? Or toLocale?

src/views/v3/Bridge/AssetPicker/TokenItem.tsx
```diff
@@ -27,6 +27,7 @@ type TokenItemProps = {
   isSelected?: boolean;
   isFetchingBalance?: boolean;
   isSource?: boolean;
+  dimmed?: boolean;
```
bradleyplaydon: nit: isDimmed

src/views/v3/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -119,38 +115,53 @@ const TokenList = (props: Props) => {
...
-    props.isFetching,
-    sortedTokens.length,
-  ]);
+  }, [props.isFetching, sortedTokens.length]);
 
   const shouldShowLoadingState = listState === 'loading';
   const shouldShowEmptyMessage = listState === 'empty';
 
+  // Build sectioned list for source picker when not searching
+  const isGroupingEnabled = props.isSource && !props.searchQuery;
```
bradleyplaydon, line 125: Worth extracting this logic to a hook? 124 - 157

src/views/v3/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -161,32 +172,51 @@ const TokenList = (props: Props) => {
...
-              props.onSelectToken(token);
-            }}
-            isSource={props.isSource}
-            balance={balance}
-            price={price}
-            isSelected={token.key === props.selectedToken?.key}
-            isFetchingBalance={props.isFetchingBalances}
-          />
+          <Fragment key={token.key}>
+            {isGroupingEnabled && index === 0 && ownedCount > 0 && (
```
bradleyplaydon: Would be nice to extract this into a TokenListHeader component to reduce number of lines, could also move away from the ternary then.

src/views/v3/Bridge/AssetPicker/TokenList.tsx
kyleleighton: I don't see this being used anywhere else soon, and it's such a simple determination that I don't think a hook is useful.

src/views/v3/Bridge/AssetPicker/TokenList.tsx
kyleleighton: I can, but it won't cut it down, since the component definition will be longer than the two instances

src/utils/index.ts
kyleleighton: No. That's specifically what this is overriding. Those return 0.00 for low balances, which is the problem.

src/views/v3/Bridge/AssetPicker/TokenList.tsx
bradleyplaydon: > I don't see this being used anywhere else soon, and it's such a simple determination that I don't think a hook is useful.

Doesn't need to be used elsewhere, we could have a directory specific for TokenList hooks e.g hooks/TokenList/...

It's more about reducing overall lines in parent component here.

src/views/v3/Bridge/AssetPicker/TokenList.tsx
bradleyplaydon: My bad, should of been more specific I mean from lines 193 - 206

src/hooks/useTokenListGrouping.ts
```diff
@@ -0,0 +1,47 @@
+import { useMemo } from 'react';
+import type { Token } from 'config/tokens';
+import { amount as sdkAmount } from '@wormhole-foundation/sdk';
+import type { Balances } from 'utils/wallet';
+
+export function useTokenListGrouping({
```
bradleyplaydon, line 6: ❤️

bradleyplaydon (approved):

src/views/v3/Bridge/AssetPicker/TokenList.tsx
```diff
@@ -161,32 +149,42 @@ const TokenList = (props: Props) => {
...
-              props.onSelectToken(token);
-            }}
-            isSource={props.isSource}
-            balance={balance}
-            price={price}
-            isSelected={token.key === props.selectedToken?.key}
-            isFetchingBalance={props.isFetchingBalances}
-          />
+          <Fragment key={token.key}>
+            <TokenSectionHeader
```
priyajeet, line 170: Can this move inside the <TokenItem> in which case you won't need wrapping Fragment/key? Might need hasHeader prop added.

src/hooks/useTokenListGrouping.ts
priyajeet: I think for future reference, this doesn't need to be a hook. Can be a simple plain old JS function as they are normally less expensive that hooks that tie into react lifecycles/closure. Since we have access to fast-memoize, if there is no real react usage in a function, best to keep it as a simple function over a hook. I will admit that, performance penalty of using a hook over a function is negligible, but non-0.

From claude

```
Why hooks have more overhead:

- Hooks tie into React's internal state management and lifecycle systems
- They create closures and maintain references in React's fiber tree
- Each hook call adds to React's internal bookkeeping during renders
- Custom hooks that use multiple built-in hooks compound this overhead

When to prefer plain functions:

- Pure utility functions - calculations, data transformations, formatting
- Stateless logic - anything that doesn't need React state or effects
- Performance-critical code that runs frequently
- Reusable logic that might be used outside React components
```

PR #3772, opened 2025-09-08
test: Add unit test for AmountInput

emreboga:
This will be the first unit test for v3 components. It verifies basic functionality for the amount input and we can use this as a template to add more tests quicker using Claude as well.


src/contexts/wallet/WalletContext.test.tsx
```diff
@@ -91,23 +91,22 @@ describe('WalletContext with InternalWalletProvider', () => {
     );
 
     // Sending wallet (Ethereum)
-    let sendingConnectPromise: Promise<any>;
-    act(() => {
-      sendingConnectPromise = result.current!.connectWallet(
+    let sendingWallet: any;
+    await act(async () => {
```
emreboga, line 95: Testing lib's act is a wrapper around React's act function, which is async: `await act(async actFn)`

vitest.config.ts
```diff
@@ -13,6 +14,7 @@ export default defineConfig({
       '**/e2e/**',
       '**/tests/**',
     ],
+    watch: false,
```
emreboga, line 17: I believe the default setting should be no-watch.

src/views/v3/Bridge/AmountInput.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+  supportedSourceTokens: [],
+  tokenBalance: null,
+  receiveAmount: undefined,
+  error: undefined,
+  warning: undefined,
+  onChange: vi.fn(),
+  onDebouncedChange: vi.fn(),
+};
+
+const wrapper = ({ children }: { children: React.ReactNode }) => (
```
bradleyplaydon: Could add the store as a param and then wouldn't have to redefine the nodes? On line 88-90 for example, also think the name should be 'AppWrapper'?

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3772/commits/d7e717ec78980f14ee4d9eea960e810edd23423e

bradleyplaydon (approved):

PR #3779, opened 2025-09-09
fix: Fetch quote only for selected route during tx confirmation

emreboga:
This will ensure to fetch quote only for the selected route after user clicks "Confirm transaction".


emreboga:
No longer needed as Confirm action already makes a single quote call.

PR #3786, opened 2025-09-10
Globally Replace WSOL with SOL

kyleleighton:
- Globally renames WSOL to SOL.
- Prevents swapping same-chain native/wrapped tokens
- Cleans up a prior PR

### DIFF
--- a/src/routes/sdkv2/route.ts
+++ b/src/routes/sdkv2/route.ts
@@ -11,7 +11,6 @@ import {
   routes,
   chainToPlatform,
   isSameToken,
-  isNative,
   TransferState,
 } from '@wormhole-foundation/sdk';
 import type { Token } from 'config/tokens';
@@ -22,6 +21,10 @@ import config, { getWormholeContextV2 } from 'config';
 import { sleep } from 'utils';
 import { isFrankensteinToken } from 'utils';
 import { isNttToken } from 'utils/ntt';
+import {
+  getWrappedNativeToken,
+  shouldFilterSameChainToken,
+} from 'utils/wrappedNativeTokens';
 
 type Amount = sdkAmount.Amount;
 
@@ -146,18 +149,10 @@ export class SDKv2Route {
       routeSupportedTokenFetcher,
     );
 
-    // Pre-compute wrapped native address when filtering same-chain swaps
-    let wrappedNativeAddr: string | undefined = undefined;
-    if (isSameChain) {
-      try {
-        const tb = await toContext.context.getTokenBridge();
-        wrappedNativeAddr = (await tb.getWrappedNative())
-          .toString()
-          .toLowerCase();
-      } catch {
-        // No-op
-      }
-    }
+    // Get wrapped native address for same-chain swaps
+    const wrappedNativeAddr = isSameChain
+      ? getWrappedNativeToken(config.network, toContext.chain)?.toLowerCase()
+      : undefined;
 
     const filteredTokens = destTokens.filter((t) => {
       const token = config.tokens.get(t);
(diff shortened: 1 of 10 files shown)


src/routes/sdkv2/route.ts
```diff
@@ -148,6 +149,17 @@ export class SDKv2Route {
...
     );
 
+    // Pre-compute wrapped native address when filtering same-chain swaps
+    let wrappedNativeAddr: string | undefined = undefined;
+    if (isSameChain) {
+      try {
+        const tb = await toContext.context.getTokenBridge();
+        wrappedNativeAddr = tb.getWrappedNative().toString().toLowerCase();
+      } catch {
+        console.log('Unable to fetch wrapped native token');
```
priyajeet: Lets remove these logs

src/routes/sdkv2/route.ts
```diff
@@ -158,6 +170,24 @@ export class SDKv2Route {
         return false;
       }
 
+      // Block same-chain swaps between native gas token and its wrapped native token
+      if (isSameChain && wrappedNativeAddr) {
+        const srcIsNative = sourceToken.isNativeGasToken;
+        const srcIsWrappedNative =
```
priyajeet: nit: can we omit the word Native from the wrapped variable :)

src/routes/sdkv2/route.ts
```diff
@@ -158,6 +170,24 @@ export class SDKv2Route {
         return false;
       }
 
+      // Block same-chain swaps between native gas token and its wrapped native token
+      if (isSameChain && wrappedNativeAddr) {
+        const srcIsNative = sourceToken.isNativeGasToken;
```
priyajeet: Should this also use isNative() like below?

src/routes/sdkv2/route.ts
```diff
@@ -148,6 +149,17 @@ export class SDKv2Route {
       routeSupportedTokenFetcher,
     );
 
+    // Pre-compute wrapped native address when filtering same-chain swaps
+    let wrappedNativeAddr: string | undefined = undefined;
+    if (isSameChain) {
+      try {
+        const tb = await toContext.context.getTokenBridge();
+        wrappedNativeAddr = tb.getWrappedNative().toString().toLowerCase();
```
priyajeet: (await tb.getWrappedNative()) since its a promise

src/routes/sdkv2/route.ts
```diff
@@ -158,6 +170,24 @@ export class SDKv2Route {
...
+      if (isSameChain && wrappedNativeAddr) {
+        const srcIsNative = sourceToken.isNativeGasToken;
+        const srcIsWrappedNative =
+          sourceToken.address.toString().toLowerCase() === wrappedNativeAddr;
+
+        const destIsWrappedNative =
+          t.address.toString().toLowerCase() === wrappedNativeAddr;
+        const destIsNative = isNative(t.address);
+
+        if (
```
priyajeet: ~My one other question is, should this logic be actually inside the new useTokenListWithSearch() where we build and filter the list and also includes search. And if not, should the filtering logic over there be removed since we are doing it here. Searching of the tokens doesn't hit this code path.~

priyajeet: ~But also probably take a glance at this line~ https://github.com/wormhole-foundation/wormhole-connect/blob/59ec738b421a563c7bac8ced9d507faf967f03dc/src/hooks/useTokenListWithSearch.ts#L80 ~since we want the search to also filter out the wrapped (or native) token~

priyajeet: I think this is fine as you have it. When we search for SOL its fine to show both. Just that the user will not find any route for it because of this logic here. If we want to hide it from the search also, then same logic will be over here https://github.com/wormhole-foundation/wormhole-connect/blob/59ec738b421a563c7bac8ced9d507faf967f03dc/src/hooks/useTokenListWithSearch.ts#L143

priyajeet (approved):

PR #3787, opened 2025-09-11
fix: remove double quoting when initiating a transaction

priyajeet:


src/routes/sdkv2/route.ts
```diff
@@ -172,24 +173,18 @@ export class SDKv2Route {
...
       destToken,
       sourceChain,
       destChain,
       recipient,
     );
 
-    const wh = await getWormholeContextV2();
-    const route = new this.rc(wh);
-    const validationResult = await route.validate(req, {
+    const router = await this.createRouter();
```
priyajeet: router? routeInstance?
The word route is overloaded

src/routes/operator.ts
```diff
@@ -216,17 +229,69 @@ export default class RouteOperator {
...
+  ) {
+    const route = this.get(routeName);
+
+    const quoteParams: QuoteParams = {
+      amount,
+      sourceChain,
+      sourceToken,
+      destChain,
+      destToken,
+      nativeGas: options.nativeGas,
```
priyajeet, line 251: Is this controlled by the toggle button on a quote? Can change after a quote is cached?

src/routes/operator.ts
```diff
@@ -216,17 +229,69 @@ export default class RouteOperator {
...
 // async functions asking for the same quote gracefully.
 //
 // If we are already fetching a quote and a second hook requests the same quote elsewhere,
-// we queue up a Promise in `QuoteCacheEntry.pending` that we resolve when the original
+// we queue up a Promise in `QuoteMetadataEntry.pending` that we resolve when the original
 // quote request is resolved. This just prevents us from making redundant API calls when
 // multiple components or hooks are interested in a quote.
-class QuoteCache {
-  cache: Record<string, QuoteCacheEntry>;
+class QuoteMetadataCache {
```
priyajeet, line 280: Will move to its own file in a followup

src/routes/operator.ts
```diff
@@ -360,31 +413,41 @@ interface QuotePromiseHandlers {
   reject: (err: Error) => void;
 }
 
-class QuoteCacheEntry {
-  // Last quote we received (the cached value)
-  result: QuoteResult;
+class QuoteMetadataEntry {
```
priyajeet, line 417: Will move to its own file in a followup

src/routes/sdkv2/route.ts
kev1n-peters: What if we call it:
`const routeInstance = this.createRouteInstance()`

kev1n-peters: `this.rc` is a `routes.RouteConstructor` so I think calling it `routeInstance` makes sense (to me at least)

src/routes/operator.ts
kev1n-peters: Yeah, this is controlled by the toggle button and can change after a quote is cached. If it changes, the quoteParamsKey (below) should change triggering a quote re-fetch

src/routes/operator.ts
```diff
@@ -360,31 +413,41 @@ interface QuotePromiseHandlers {
...
-  constructor(result: QuoteResult) {
-    this.result = result;
+  // Optional route used for the quote
+  router?: routes.Route<Network>;
+  // Optional request used for the quote
+  request?: routes.RouteTransferRequest<Network>;
+
+  constructor(
+    quote: QuoteResult,
+    router?: routes.Route<Network>,
```
kev1n-peters: Should these be required?

priyajeet: These may not be available during an unhandled error or quote error. I could pass it back from route.ts for quote error, but unhandled error can have it missing

src/routes/operator.ts
```diff
@@ -243,96 +296,88 @@ class QuoteCache {
...
-              result.expires = new Date(now + 60_000);
-            }
-          }
-
-          console.debug(`Fetched quote`, routeName, result);
-
-          this.cache[key] = new QuoteCacheEntry(result);
-        })
-        .catch((err: any) => {
-          console.debug(`Error fetching quote`, routeName, err);
```
kev1n-peters, line 320: Do you think it would be helpful to keep in these console.debugs for quick debugging purposes?

kev1n-peters (dismissed):

kev1n-peters:
I think the changes look good

src/routes/operator.ts
priyajeet: Should this be sent as an event to log in portal?

kev1n-peters: I think it's fine to leave it as a console log for now, maybe we can create an issue and follow-up?

kev1n-peters (approved):

PR #3792, opened 2025-09-16
refactor: prepare for mayan route cleanup

priyajeet:
moves index.ts to MayanRouteBase.ts to
1. Free up the index name
2. Move the largest code (class MayanRouteBase) that will need a diff later, into the file it will eventually belong


src/routes/mayan/MayanRouteBase.ts
```diff
@@ -0,0 +1,1152 @@
+import type {
```
priyajeet, line 1: copy/paste/no-changes

kev1n-peters (approved):

bradleyplaydon (approved):

PR #3796, opened 2025-09-16
chore: improve low gas balance message on Solana

kyleleighton:
Adds a low gas balance message on Solana when the error regex `simulation failed` comes through. This actually does absolutely mean for sure that the error is, in fact, due to this. However, that is apparently by far the most common reason we see it and this should clean up Sentry.


src/utils/errors.ts
```diff
@@ -74,17 +78,20 @@ export function interpretTransferError(
...
+      const gasChain = transferDetails.fromChain;
+      try {
+        const gasToken = getGasToken(gasChain);
+        const gasSymbol = getTokenSymbol(gasToken);
+        const chainName = chainDisplayName(gasChain);
+        const chainSuffix = chainName ? ` on ${chainName}` : '';
+        uiErrorMessage = `Insufficient ${gasSymbol} for fees${chainSuffix}, please add more ${gasSymbol} and try again`;
+      } catch {
+        uiErrorMessage =
+          'Insufficient funds for network fees, please add more funds and try again';
```
priyajeet: probably make this a const and re-use

src/utils/errors.ts
```diff
@@ -74,17 +78,20 @@ export function interpretTransferError(
...
+      SIMULATION_ACCOUNT_NOT_FOUND_REGEX.test(e?.message) ||
+      INSUFFICIENT_LAMPORTS_REGEX.test(e?.message)
+    ) {
+      const gasChain = transferDetails.fromChain;
+      try {
+        const gasToken = getGasToken(gasChain);
+        const gasSymbol = getTokenSymbol(gasToken);
+        const chainName = chainDisplayName(gasChain);
+        const chainSuffix = chainName ? ` on ${chainName}` : '';
+        uiErrorMessage = `Insufficient ${gasSymbol} for fees${chainSuffix}, please add more ${gasSymbol} and try again`;
```
priyajeet: fees${chainSuffix}, typo?

kyleleighton: Not a typo. In case there's a chain name failure, this will say "Insufficient BTC for fees, please..." without a space before the comma and if it works it will say "Insufficient ETH for fees on ARB, please..."

priyajeet (dismissed):

priyajeet (approved):

PR #3802, opened 2025-09-17
fix: findBySymbol not returning a token on multiple symbols

bradleyplaydon:
- ~findBySymbol on the tokens class for getting a Token by symbol wasn't working correctly because we override WSOL there's 2 SOL's and it was hitting a line to filter out the wrapped token, but because tokenBridgeOriginalTokenId was never set it wasn't filtering the wrapped token.~ We decided instead to just filter out non native tokens if `matching.length > 1` could be improved more in the future. NOTE: **tokenBridgeOriginalTokenId** isn't being set on buildTokenCache but is in a loop below.
- Deconstruct Tokens class constructor for readability
- Updated getTokenSymbol to getTokenDisplaySymbolByTokenAddress to be more explicit 
- Fixed WSOL to SOL on testnet

### DIFF
--- a/src/config/tokens.ts
+++ b/src/config/tokens.ts
@@ -9,9 +9,8 @@ import {
   UniversalAddress,
 } from '@wormhole-foundation/sdk';
 import type { TokenIcon, TokenConfig, WrappedTokenAddresses } from './types';
-import { getWormholeContextV2 } from './index';
+import config, { getWormholeContextV2 } from './index';
 import { isValidSuiType } from '@wormhole-foundation/sdk-sui';
-
 import { fetchTokenMetadata } from 'utils/coingecko';
 
 const TOKEN_CACHE_VERSION = 1;
@@ -389,9 +388,16 @@ export class TokenCache extends TokenMapping<Token> {
   // This should be used sparingly/never... use addresses instead.
   // Excludes wrapped tokens
   findBySymbol(chain: Chain, symbol: string): Token | undefined {
-    let matching = this.getAllForChain(chain).filter(
-      (t) => t.symbol.toLowerCase() === symbol.toLowerCase(),
-    );
+    const chainOverrides = config.ui?.tokenNameOverrides?.[chain];
+    const lowerCaseSymbol = symbol.toLowerCase();
+
+    let matching = this.getAllForChain(chain).filter((t) => {
+      const symbolMatch = lowerCaseSymbol === t.symbol.toLowerCase();
+      const overrideMatch =
+        lowerCaseSymbol ===
+        chainOverrides?.[t.address.toString()]?.toLowerCase();
+      return symbolMatch || overrideMatch;
+    });
 
     if (matching.length > 1) {
       // Exclude wrapped tokens if there's multiple matches
--- a/src/config/tokens.ts
+++ b/src/config/tokens.ts
@@ -496,7 +496,7 @@ export class TokenCache extends TokenMapping<Token> {
       decimals,
       symbol,
       name,
-      icon: image || undefined,
+      icon: image,
       tokenBridgeOriginalTokenId,
       coingeckoId,
     });
--- a/src/config/tokens.ts
+++ b/src/config/tokens.ts
@@ -1,9 +1,4 @@
-import type {
-  Chain,
-  TokenId,
-  TokenAddress,
-  Network,
-} from '@wormhole-foundation/sdk';
+import type { Chain, TokenId, TokenAddress } from '@wormhole-foundation/sdk';
 import {
   canonicalAddress,
   isTokenId,
@@ -13,17 +8,10 @@ import {
   chainToPlatform,
   UniversalAddress,
 } from '@wormhole-foundation/sdk';
-import type {
-  TokenIcon,
-  TokenConfig,
-  WrappedTokenAddresses,
-  WormholeConnectConfig,
-} from './types';
+import type { TokenIcon, TokenConfig, WrappedTokenAddresses } from './types';
 import config, { getWormholeContextV2 } from './index';
 import { isValidSuiType } from '@wormhole-foundation/sdk-sui';
 import { fetchTokenMetadata } from 'utils/coingecko';
-import { getWrappedNativeToken } from 'utils/wrappedNativeTokens';
-import { capitalize } from './utils';
 
 const TOKEN_CACHE_VERSION = 1;
 
@@ -425,12 +413,11 @@ export class TokenCache extends TokenMapping<Token> {
     if (matching.length > 1) {
       // Exclude wrapped tokens if there's multiple matches
       matching = matching.filter((t) => {
-        return !t.isTokenBridgeWrappedToken;
+        return !t.isTokenBridgeWrappedToken && t.address !== 'native';
       });
     }
 
     if (matching.length === 1) {
-      console.log(`Found token by symbol: ${symbol} -> ${matching[0]}`);
       return matching[0];
     } else if (matching.length > 1) {
       // This means there's more than one native token (not wrapped) with this symbol
@@ -568,19 +555,9 @@ export function buildTokenCache(
   tokens: TokenConfig[],
   wrappedTokens: WrappedTokenAddresses,
   cacheKey: string,
-  customConfig: WormholeConnectConfig = {},
 ): TokenCache {
   const cache = TokenCache.load(cacheKey);
-  const network = capitalize(
-    customConfig.network ||
-      import.meta.env.REACT_APP_CONNECT_ENV?.toLowerCase() ||
-      'Mainnet',
-  ) as Network;
   for (const { tokenId, symbol, name, icon, decimals } of tokens) {
-    const wrappedToken = getWrappedNativeToken(network, tokenId.chain);
-    const isWrappedToken =
-      wrappedToken &&
-      tokenId.address.toLowerCase() === wrappedToken.toLowerCase();
     const token = new Token({
       chain: tokenId.chain,
       address: tokenId.address.toString(),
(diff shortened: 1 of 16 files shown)


src/hooks/useTokenListWithSearch.ts
```diff
@@ -119,7 +119,8 @@ export const useTokenListWithSearch = ({
 
     if (deferredSearch) {
       tokens = tokens.filter((token) => {
-        const overrideName = getTokenSymbol(token)?.toLowerCase();
+        const overrideName =
+          getTokenDisplaySymbolByTokenAddress(token)?.toLowerCase();
```
bradleyplaydon, line 123: Just naming convention change to be more explicit

src/views/v3/TxHistory/Item/index.tsx
```diff
@@ -78,7 +78,8 @@ const TxHistoryItem = (props: Props) => {
         <AssetBadge chainConfig={sourceChainConfig} token={fromToken} />
         <Stack direction="column" marginLeft="12px">
           <Typography fontSize={16}>
-            {amount} {fromToken ? getTokenSymbol(fromToken) : ''}
+            {amount}{' '}
```
priyajeet, line 81:
```jsx
<Stack direction="row" spacing={2 or whatever}>
  <Item>{amount}</Item>
  <Item>{fromToken ? getTokenDisplaySymbolByTokenAddress(fromToken) : ''}</Item>
</Stack>
```

src/views/v3/Redeem/TransactionDetails/index.tsx
```diff
@@ -164,7 +164,8 @@ const TransactionDetails = () => {
         <AssetBadge chainConfig={destChainConfig} token={destToken} />
         <Stack direction="column" marginLeft="12px">
           <Typography fontSize={16}>
-            {formattedReceiveAmount} {getTokenSymbol(destToken)}
+            {formattedReceiveAmount}{' '}
```
priyajeet, line 167: Same as other comment

src/views/v3/Bridge/Routes/SingleRoute.tsx
```diff
@@ -383,7 +383,7 @@ const SingleRoute = (props: Props) => {
         component="div"
         marginBottom="6px"
       >
-        {receiveAmountTrunc} {getTokenSymbol(destToken)}
+        {receiveAmountTrunc} {getTokenDisplaySymbolByTokenAddress(destToken)}
```
priyajeet, line 386: Same as other comment

src/views/v3/Bridge/AssetPicker/index.tsx
```diff
@@ -398,7 +404,9 @@ function AssetPicker(props: Props) {
           color={theme.palette.text.secondary + OPACITY[50]}
           fontSize="12px"
           fontWeight={500}
-        >{`1 ${getTokenSymbol(props.token)} = ${unitPrice}`}</Typography>
+        >{`1 ${getTokenDisplaySymbolByTokenAddress(
```
priyajeet, line 407: getTokenDisplaySymbolByTokenAddress(props.token) can be a const on top.
And all business logic ?: can move to consts too on top. So that JSX is just simple {variable}

src/config/tokens.ts
```diff
@@ -389,16 +410,27 @@ export class TokenCache extends TokenMapping<Token> {
...
     if (matching.length > 1) {
       // Exclude wrapped tokens if there's multiple matches
-      matching = matching.filter((t) => !t.isTokenBridgeWrappedToken);
+      matching = matching.filter((t) => {
+        return !t.isTokenBridgeWrappedToken;
+      });
     }
 
     if (matching.length === 1) {
+      console.log(`Found token by symbol: ${symbol} -> ${matching[0]}`);
```
priyajeet: delete

src/config/tokens.ts
```diff
@@ -536,18 +568,33 @@ export function buildTokenCache(
   tokens: TokenConfig[],
   wrappedTokens: WrappedTokenAddresses,
   cacheKey: string,
+  customConfig: WormholeConnectConfig = {},
 ): TokenCache {
   const cache = TokenCache.load(cacheKey);
-
+  const network = capitalize(
```
priyajeet: Shouldn't this be done when config is being built? Not here? Or did you have for debugging...in any case remove

src/config/tokens.ts
```diff
@@ -536,18 +568,33 @@ export function buildTokenCache(
...
+      'Mainnet',
+  ) as Network;
   for (const { tokenId, symbol, name, icon, decimals } of tokens) {
-    const token = new Token(
-      tokenId.chain,
-      tokenId.address.toString(),
+    const wrappedToken = getWrappedNativeToken(network, tokenId.chain);
+    const isWrappedToken =
+      wrappedToken &&
+      tokenId.address.toLowerCase() === wrappedToken.toLowerCase();
```
priyajeet: We really should add a helper method `isSameTokenAddress()` for this instead of lower casing everywhere. Can put it next to tokents.ts::isSameToken() and update this to also use `isSameTokenAddress()`.

src/config/tokens.ts
```diff
@@ -536,18 +568,33 @@ export function buildTokenCache(
...
+  const network = capitalize(
+    customConfig.network ||
+      import.meta.env.REACT_APP_CONNECT_ENV?.toLowerCase() ||
+      'Mainnet',
+  ) as Network;
   for (const { tokenId, symbol, name, icon, decimals } of tokens) {
-    const token = new Token(
-      tokenId.chain,
-      tokenId.address.toString(),
+    const wrappedToken = getWrappedNativeToken(network, tokenId.chain);
```
priyajeet: Lets rename function to `getWrappedTokenAddress()` or `getWrappedTokenAddressString()` to be explicit what its returning.

src/config/tokens.ts
```diff
@@ -536,18 +568,33 @@ export function buildTokenCache(
...
+      chain: tokenId.chain,
+      address: tokenId.address.toString(),
       decimals,
       symbol,
       name,
       icon,
-    );
+      tokenBridgeOriginalTokenId: isWrappedToken
+        ? {
+            chain: tokenId.chain,
```
priyajeet: Shouldn't the address be of the original native token? aka `native` ?

src/store/transferInput.ts
```diff
@@ -123,13 +121,10 @@ const performModificationsIfToChainChanged = (state: TransferInputState) => {
...
     if (destToken && toChain) {
       if (destToken.chain !== toChain && destToken.symbol) {
-        const withSameSymbol = config.tokens.findBySymbol(
-          toChain,
-          destToken.symbol,
-        );
+        const withSameSymbol = getTokenDisplaySymbolByTokenAddress(destToken);
 
         if (withSameSymbol) {
-          state.destToken = withSameSymbol.tuple;
```
priyajeet: 🙏🏻

src/config/tokens.ts
kev1n-peters: Remove this or change to debug?

src/config/tokens.ts
```diff
@@ -536,18 +568,33 @@ export function buildTokenCache(
...
+      tokenId.address.toLowerCase() === wrappedToken.toLowerCase();
+    const token = new Token({
+      chain: tokenId.chain,
+      address: tokenId.address.toString(),
       decimals,
       symbol,
       name,
       icon,
-    );
+      tokenBridgeOriginalTokenId: isWrappedToken
```
kev1n-peters: A couple of things:
1. We should not set the `tokenBridgeOriginalTokenId` property unless this is actually a token bridge wrapped token. The `getWrappedNativeToken` function above returns the wrapped version of the native/gas token, which is not a token bridge wrapped token. To determine if a token is a token bridge wrapped token, we have to call:

```typescript
const tb = await chain.getTokenBridge(); // where `chain` is the token's chain
if (await tb.isWrappedAsset(tokenId.address)) {
  tokenBridgeOriginalTokenId = await tb.getOriginalAsset(tokenId.address);
```
But since we can't and don't want to call this for every token when building the cache, we have to do something different.

2. What did adding this fix? Maybe we can work backwards from there to find a fix.

bradleyplaydon: Discussed - for now we just filter out non native tokens if matching.length > 1 for findBySymbol. See PR description.

src/config/tokens.ts
```diff
@@ -389,13 +398,23 @@ export class TokenCache extends TokenMapping<Token> {
...
+        chainOverrides?.[t.address.toString()]?.toLowerCase();
+
+      return symbolMatch || overrideMatch;
+    });
 
     if (matching.length > 1) {
       // Exclude wrapped tokens if there's multiple matches
-      matching = matching.filter((t) => !t.isTokenBridgeWrappedToken);
+      matching = matching.filter((t) => {
+        return !t.isTokenBridgeWrappedToken && t.address !== 'native';
```
kev1n-peters: I think we should leave this as it was before. We can do this below:
```
    } else if (matching.length > 1) {
      const gasToken = matching.find((t) => t.address === 'native');
      // prefer the gas token if there are multiple native tokens with this symbol
      if (gasToken) return gasToken;
      // This means there's more than one native token (not wrapped) with this symbol
      console.error(`Ambiguous token symbol: ${symbol}`);
    }
```
If an integrator needs to get wrapped SOL from a URL, then they can pass in the wrapped SOL address.

priyajeet (approved):

src/views/v3/TxHistory/Item/index.tsx
bradleyplaydon: Just a name change for now.

PR #3808, opened 2025-09-21
fix: resolve issue #3804

ktan-wolf:
fixed the issue #3804 , pls review it , would be lovely to hear back from you , attaching the fix video also 

https://github.com/user-attachments/assets/0ef8d580-cd25-4084-9384-bd78b95c6553

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,6 +149,7 @@
     "@vitejs/plugin-react-swc": "^3.5.0",
     "@vitest/ui": "^3.2.4",
     "commitizen": "^4.3.1",
+    "cross-env": "^10.0.0",
     "cz-conventional-changelog": "^3.3.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -149,6 +149,7 @@
     "@vitejs/plugin-react-swc": "^3.5.0",
     "@vitest/ui": "^3.2.4",
     "commitizen": "^4.3.1",
+    "cross-env": "^10.0.0",
```
kev1n-peters, line 152: What is the reason for adding `cross-env`?

ktan-wolf: using the windowOS , running it in 'NODE_OPTIONS' gives the error as external command ,hence used cross-env.
Screenshot 2025-09-22 213743

PR #3812, opened 2025-09-23
fix: fixed and refactored signAndSendTransaction on solana

bradleyplaydon:
- Refactored retry and resend logic on solana signAndSendTransaction
  - Changed promise race delay from 5 seconds -> 1 second, this means we resend every second if confirmation promise doesn't resolve.
  - Remove custom retry loop for retry from es-toolkit and added function recoverBlockheightExceededTransaction

### DIFF
--- a/src/utils/inProgressTxCache.ts
+++ b/src/utils/inProgressTxCache.ts
@@ -1,6 +1,6 @@
 import config from 'config';
 import type { TransactionLocal } from 'config/types';
-import { isEmptyObject, JSONReplacer } from 'utils';
+import { isEmptyObject, stringifyWithBigInt } from 'utils';
 
 const LOCAL_STORAGE_KEY = 'transactions:inprogress';
 const LOCAL_STORAGE_MAX = 3;
@@ -169,7 +169,10 @@ export const addTxToLocalStorage = (
 
   // Update the list
   try {
-    ls.setItem(config.cacheKey(LOCAL_STORAGE_KEY), JSONReplacer(newList));
+    ls.setItem(
+      config.cacheKey(LOCAL_STORAGE_KEY),
+      stringifyWithBigInt(newList),
+    );
   } catch (e: any) {
     // We can get two different errors:
     // 1- TypeError from JSON.stringify
@@ -193,7 +196,10 @@ export const removeTxFromLocalStorage = (txHash: string) => {
       // remove the item and update localStorage
       items.splice(removeIndex, 1);
       try {
-        ls.setItem(config.cacheKey(LOCAL_STORAGE_KEY), JSONReplacer(items));
+        ls.setItem(
+          config.cacheKey(LOCAL_STORAGE_KEY),
+          stringifyWithBigInt(items),
+        );
       } catch (e: any) {
         // We can get two different errors:
         // 1- TypeError from JSON.stringify
@@ -223,7 +229,10 @@ export const updateTxInLocalStorage = (
       // Update item property and put back in local storage
       items[idx][key] = value;
       try {
-        ls.setItem(config.cacheKey(LOCAL_STORAGE_KEY), JSONReplacer(items));
+        ls.setItem(
+          config.cacheKey(LOCAL_STORAGE_KEY),
+          stringifyWithBigInt(items),
+        );
       } catch (e: any) {
         // We can get two different errors:
         // 1- TypeError from JSON.stringify
--- a/src/utils/inProgressTxCache.ts
+++ b/src/utils/inProgressTxCache.ts
@@ -1,6 +1,6 @@
 import config from 'config';
 import type { TransactionLocal } from 'config/types';
-import { isEmptyObject } from 'utils';
+import { isEmptyObject, JSONReplacer } from 'utils';
 
 const LOCAL_STORAGE_KEY = 'transactions:inprogress';
 const LOCAL_STORAGE_MAX = 3;
@@ -8,8 +8,6 @@ const LOCAL_STORAGE_MAX = 3;
 // Bigint types cannot be serialized to a string
 // That's why we need to provide a replacer function to handle it separately
 // Please see for details: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/BigInt_not_serializable
-const JSONReplacer = (_, value: any) =>
-  typeof value === 'bigint' ? value.toString() : value;
 
 // Checks the existance of the props with the given types in a parent object
 const validateChildPropTypes = (
@@ -171,10 +169,7 @@ export const addTxToLocalStorage = (
 
   // Update the list
   try {
-    ls.setItem(
-      config.cacheKey(LOCAL_STORAGE_KEY),
-      JSON.stringify(newList, JSONReplacer),
-    );
+    ls.setItem(config.cacheKey(LOCAL_STORAGE_KEY), JSONReplacer(newList));
   } catch (e: any) {
     // We can get two different errors:
     // 1- TypeError from JSON.stringify
@@ -198,10 +193,7 @@ export const removeTxFromLocalStorage = (txHash: string) => {
       // remove the item and update localStorage
       items.splice(removeIndex, 1);
       try {
-        ls.setItem(
-          config.cacheKey(LOCAL_STORAGE_KEY),
-          JSON.stringify(items, JSONReplacer),
-        );
+        ls.setItem(config.cacheKey(LOCAL_STORAGE_KEY), JSONReplacer(items));
       } catch (e: any) {
         // We can get two different errors:
         // 1- TypeError from JSON.stringify
@@ -231,10 +223,7 @@ export const updateTxInLocalStorage = (
       // Update item property and put back in local storage
       items[idx][key] = value;
       try {
-        ls.setItem(
-          config.cacheKey(LOCAL_STORAGE_KEY),
-          JSON.stringify(items, JSONReplacer),
-        );
+        ls.setItem(config.cacheKey(LOCAL_STORAGE_KEY), JSONReplacer(items));
       } catch (e: any) {
         // We can get two different errors:
         // 1- TypeError from JSON.stringify
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -424,3 +424,9 @@ export const isExecutorRoute = (route: string | undefined) => {
   }
   return route.endsWith('ExecutorRoute');
 };
+
+export const JSONReplacer = (json: any) => {
+  return JSON.stringify(json, (_key, value) =>
+    typeof value === 'bigint' ? value.toString() : value,
+  );
+};
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -425,7 +425,7 @@ export const isExecutorRoute = (route: string | undefined) => {
   return route.endsWith('ExecutorRoute');
 };
 
-export const JSONReplacer = (json: any) => {
+export const stringifyWithBigInt = (json: any) => {
   return JSON.stringify(json, (_key, value) =>
     typeof value === 'bigint' ? value.toString() : value,
   );
--- a/src/utils/wallet/solana.ts
+++ b/src/utils/wallet/solana.ts
@@ -29,10 +29,13 @@ import {
 
 import config from 'config';
 
+const CONFIRMATION_PROMISE_TIMER = 3_000; // How long to wait for confirmation before resending
+
 import type { SolanaUnsignedTransaction } from '@wormhole-foundation/sdk-solana';
 import type { Chain, Network } from '@wormhole-foundation/sdk';
 import { setPriorityFeeInstructions } from 'utils/solana';
 import { retry } from 'es-toolkit';
+import { JSONReplacer } from 'utils';
 
 const getWalletName = (wallet: Wallet) =>
   wallet.getName().toLowerCase().replaceAll('wallet', '').trim();
@@ -140,79 +143,17 @@ export async function signAndSendTransactionWithResends(
     commitment,
   );
 
-  const confirmationPromiseTimer = 3_000; // How long to wait for confirmation before resending
+  const transaction = { serializedTransaction, sendOptions };
 
-  const signature = await signAndConfirmTransactionWhilstResending(
-    { serializedTransaction, sendOptions },
+  const signature = await resendTransactionUntilConfirmed(
     connection,
+    transaction,
     blockhash,
     lastValidBlockHeight,
     commitment,
-    confirmationPromiseTimer,
   );
 
   return signature;
-}
-
-async function signAndConfirmTransactionWhilstResending(
-  transaction: {
-    serializedTransaction: Uint8Array | Buffer | number[];
-    sendOptions?: SendOptions;
-  },
-  connection: Connection,
-  blockHash: string,
-  lastValidBlockHeight: number,
-  commitment: Commitment,
-  confirmationPromiseTimer: number,
-): Promise<string> {
-  const { signature, confirmPromise } =
-    await sendTransactionAndGetConfirmPromise(
-      transaction,
-      connection,
-      blockHash,
-      lastValidBlockHeight,
-      commitment,
-    );
-
-  await resendTransactionUntilConfirmed(
-    signature,
-    connection,
-    transaction,
-    confirmationPromiseTimer,
-    confirmPromise,
-  );
-
-  return signature;
-}
-
-async function sendTransactionAndGetConfirmPromise(
-  transaction: {
-    serializedTransaction: Uint8Array | Buffer | number[];
-    sendOptions?: SendOptions;
-  },
-  connection: Connection,
-  blockHash: string,
-  lastValidBlockHeight: number,
-  commitment: Commitment,
-): Promise<{
-  signature: string;
-  confirmPromise: Promise<RpcResponseAndContext<SignatureResult>>;
-}> {
-  const signature = await connection.sendRawTransaction(
-    transaction.serializedTransaction,
-    transaction.sendOptions,
-  );
-
-  const confirmPromise = connection.confirmTransaction(
-    {
-      signature,
-      blockhash: blockHash,
-      lastValidBlockHeight,
-    },
-    commitment,
-  );
-
-  return { signature, confirmPromise };
 }
 
 /**
@@ -228,33 +169,42 @@ async function sendTransactionAndGetConfirmPromise(
  */
 
 async function resendTransactionUntilConfirmed(
-  signature: string,
   connection: Connection,
   transaction: {
     serializedTransaction: Uint8Array | Buffer | number[];
     sendOptions?: SendOptions;
   },
-  confirmationPromiseTimer: number,
-  confirmTransactionPromise: Promise<RpcResponseAndContext<SignatureResult>>,
-) {
+  blockhash: string,
+  lastValidBlockHeight: number,
+  commitment: Commitment,
+): Promise<string> {
   let isTransactionConfirmed: RpcResponseAndContext<SignatureResult> | null =
     null;
+
+  const signature = await connection.sendRawTransaction(
+    transaction.serializedTransaction,
+    transaction.sendOptions,
+  );
+
   try {
     while (!isTransactionConfirmed) {
+      const confirmPromise = connection.confirmTransaction(
+        { signature, blockhash, lastValidBlockHeight },
+        commitment,
+      );
+
       isTransactionConfirmed = await Promise.race([
-        confirmTransactionPromise,
+        confirmPromise,
         new Promise<null>((resolve) =>
-          setTimeout(() => {
-            resolve(null);
-          }, confirmationPromiseTimer),
+          setTimeout(() => resolve(null), CONFIRMATION_PROMISE_TIMER),
         ),
       ]);
+
       if (isTransactionConfirmed) {
         break;
       }
 
-      // This is the same signature so don't need response.
-      // Need to await since we are explicilty resending before confirming again
+      // Resend transaction as it's not yet confirmed
       await connection.sendRawTransaction(
         transaction.serializedTransaction,
         transaction.sendOptions,
@@ -263,18 +213,20 @@ async function resendTransactionUntilConfirmed(
   } catch (e) {
     if (e instanceof Error) {
       if (e.name === 'TransactionExpiredBlockheightExceededError') {
-        recoverBlockheightExceededTransaction(e, connection, signature);
+        await recoverBlockheightExceededTransaction(connection, signature);
       }
       console.error('Failed to resend transaction:', e);
     }
   }
 
-  if (isTransactionConfirmed && isTransactionConfirmed.value.err) {
+  if (isTransactionConfirmed?.value.err) {
     const errorMessage = formatConfirmationError(
       isTransactionConfirmed.value.err,
     );
     throw new Error(`Transaction failed: ${errorMessage}`);
   }
+
+  return signature;
 }
 
 /**
@@ -288,23 +240,22 @@ async function resendTransactionUntilConfirmed(
  * @returns The signature once found, or throws if retries are exhausted
  */
 async function recoverBlockheightExceededTransaction(
-  e: unknown,
   connection: Connection,
   signature: string,
   { retries = 5, delay = 2000 }: { retries?: number; delay?: number } = {},
-): Promise<string | null> {
-  const findTransaction = async (): Promise<string> => {
+) {
+  const findTransaction = async () => {
     const tx = await connection.getTransaction(signature, {
       commitment: 'confirmed',
       maxSupportedTransactionVersion: 0,
     });
 
-    if (tx) return signature;
+    if (tx) return;
 
     throw new Error('Transaction not yet found on chain');
   };
 
-  return retry(findTransaction, { retries, delay });
+  retry(findTransaction, { retries, delay });
 }
 
 async function createSolanaTransaction(
@@ -340,9 +291,7 @@ function formatConfirmationError(err: TransactionError): string {
 
   if (typeof err === 'object') {
     try {
-      return JSON.stringify(err, (_key, value) =>
-        typeof value === 'bigint' ? value.toString() : value,
-      );
+      return JSONReplacer(err);
     } catch {
       return 'Unstringifiable error object';
     }
--- a/src/utils/wallet/solana.ts
+++ b/src/utils/wallet/solana.ts
@@ -35,7 +35,7 @@ import type { SolanaUnsignedTransaction } from '@wormhole-foundation/sdk-solana'
 import type { Chain, Network } from '@wormhole-foundation/sdk';
 import { setPriorityFeeInstructions } from 'utils/solana';
 import { retry } from 'es-toolkit';
-import { JSONReplacer } from 'utils';
+import { stringifyWithBigInt } from 'utils';
 
 const getWalletName = (wallet: Wallet) =>
   wallet.getName().toLowerCase().replaceAll('wallet', '').trim();
@@ -186,13 +186,13 @@ async function resendTransactionUntilConfirmed(
     transaction.sendOptions,
   );
 
+  const confirmPromise = connection.confirmTransaction(
+    { signature, blockhash, lastValidBlockHeight },
+    commitment,
+  );
+
   try {
     while (!isTransactionConfirmed) {
-      const confirmPromise = connection.confirmTransaction(
-        { signature, blockhash, lastValidBlockHeight },
-        commitment,
-      );
-
       isTransactionConfirmed = await Promise.race([
         confirmPromise,
         new Promise<null>((resolve) =>
@@ -291,7 +291,7 @@ function formatConfirmationError(err: TransactionError): string {
 
   if (typeof err === 'object') {
     try {
-      return JSONReplacer(err);
+      return stringifyWithBigInt(err);
     } catch {
       return 'Unstringifiable error object';
     }
--- a/src/utils/wallet/solana.ts
+++ b/src/utils/wallet/solana.ts
@@ -134,7 +134,7 @@ export async function signAndSendTransactionWithRetry(
     sendOptions,
   );
 
-  confirmTransactionWithRetry(
+  const newSignature = await confirmTransactionWithRetry(
     signature,
     { serializedTransaction, sendOptions },
     connection,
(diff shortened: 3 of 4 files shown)


src/utils/wallet/solana.ts
```diff
@@ -81,122 +76,204 @@ export function fetchOptions(chain: Chain) {
...
-          txRetryInterval * txSendAttempts++
-        }ms, resending`,
-      );
-      try {
-        await connection.sendRawTransaction(serializedTx, sendOptions);
-      } catch (e) {
-        console.error('Failed to resend transaction:', e);
-      }
-    }
+    let hasTransactionBeenResent = false;
```
bradleyplaydon: Flag outside retry so we only resend the transaction once.

src/utils/wallet/solana.ts
```diff
@@ -81,122 +76,204 @@ export function fetchOptions(chain: Chain) {
...
+
+              try {
+                currentSignature = await connection.sendRawTransaction(
+                  transaction.serializedTransaction,
+                  transaction.sendOptions,
+                );
+
+                hasTransactionBeenResent = true;
+              } catch (resendError) {
+                console.error('Failed to resend transaction:', resendError);
```
bradleyplaydon: This code is from before just moved here, if the resend fails then we throw

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,205 @@ export function fetchOptions(chain: Chain) {
...
   request: SolanaUnsignedTransaction<Network>,
   wallet: Wallet | undefined,
   options?: ConfirmOptions,
-) {
+): Promise<string> {
   if (!wallet) throw new Error('Wallet not found');
   const rpc = config.rpcs[request.chain];
   if (!rpc) throw new Error(`${request.chain} RPC not found`);
 
   const commitment = options?.commitment ?? 'finalized';
```
priyajeet: Should we change the default to `confirmed` which will be faster?
In executor I think we default to confirmed since the fork/rehash rate was low?

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,205 @@ export function fetchOptions(chain: Chain) {
...
 
   try {
-    // This loop will break once the transaction has been confirmed or the block height is exceeded.
-    // An exception will be thrown if the block height is exceeded by the confirmTransactionPromise.
-    // The transaction will be resent if it hasn't been confirmed after the interval.
-    const txRetryInterval = 5000;
-    while (!confirmedTx) {
-      confirmedTx = await Promise.race([
-        confirmTransactionPromise,
-        new Promise<null>((resolve) =>
```
kev1n-peters, line 161: The purpose of this method is to repeatedly resend the transaction until the `confirmTransaction` promise either resolves (meaning the transaction was confirmed) or rejects (meaning the transaction's block hash expired). I no longer see us resending the transaction while waiting for `confirmTransaction`, which increases the likelihood of the block hash expiring. Where did the retry logic block confirming the transaction?

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,205 @@ export function fetchOptions(chain: Chain) {
...
-    preFlightCommitment: commitment, // See PR and linked issue for why setting this matters: https://github.com/anza-xyz/agave/pull/483
-  };
-  signature = await connection.sendRawTransaction(serializedTx, sendOptions);
-  confirmTransactionPromise = connection.confirmTransaction(
+  const signature = await connection.sendRawTransaction(
+    serializedTransaction,
+    sendOptions,
+  );
+
+  confirmTransactionWithRetry(
```
priyajeet: Does this need to be await-ed

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,205 @@ export function fetchOptions(chain: Chain) {
...
+        }
+
+        return signature;
+      },
+      { retries, delay },
+    ).catch(() => null); // return null if all retries fail
   }
 
-  return signature;
+  return Promise.resolve(null);
```
priyajeet: `return null` should suffice

src/utils/wallet/solana.ts
bradleyplaydon: Ah so couple of things.

- Firstly, the first promise was being awaited 5 seconds after sending because of the timeout. Meant no matter what we'd only check if the tx was confirmed 5 seconds later, this changes it so we check for confirmation straight away.
- I may of misunderstood what the method was trying to do, in terms of what you just said 'resending **until** the tx is confirmed', I see it being resent but the response from resending isn't stored so although it gets resent the UI wouldn't react
- Can change the approach from 5 retries and one resend to keep retrying / resending.

src/utils/wallet/solana.ts
bradleyplaydon: Need to remove this as @kev1n-peters commented, the desired logic is to resend until it's successful

src/utils/wallet/solana.ts
bradleyplaydon: Yes good spot.

src/utils/wallet/solana.ts
bradleyplaydon: Can't comment on this, don't know what impact is has. Will update if you think it's necessary
?

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,205 @@ export function fetchOptions(chain: Chain) {
...
-            commitment: 'confirmed',
-            maxSupportedTransactionVersion: 0,
-          });
-          if (tx) return signature; // Transaction actually landed
-        } catch {
-          // Ignore errors, we will retry
-        }
-        await sleep(retryDelay);
-      }
+    const recoveredSignature = await recoverBlockheightExceededTransaction(
```
priyajeet: Lets change the catch block here to check for `e` instead of passing it down

```ts
catch (e: unknown) {
  if (Error.isError(e) && e.name === 'TransactionExpiredBlockheightExceededError')

    const recoveredSignature = await recoverBlockheightExceededTransaction(
      connection,
      signature,
    );
    
    if (recoveredSignature) {
      return recoveredSignature;
    }
  }

  throw e;
}
```

src/utils/wallet/solana.ts
bradleyplaydon: this should change the logic to what you explained: https://github.com/wormhole-foundation/wormhole-connect/pull/3812/commits/4e01b1b1d0fd651c1781f968890aa372e6ee8bfb

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,216 @@ export function fetchOptions(chain: Chain) {
...
+        });
+
+        if (!tx) {
+          throw new Error('Transaction not yet found on chain');
+        }
+
+        return signature;
+      },
+      { retries, delay },
+    ).catch(() => null); // return null if all retries fail
```
priyajeet: Personally, I think we shouldn't mix match async/await syntax with then/catch imo. Makes it a bit more confusing reading.

```ts
async function getSignature() {
  const tx = await connection.getTransaction(signature, {
    commitment: 'confirmed',
    maxSupportedTransactionVersion: 0,
  });

  if (!tx) {
    throw new Error('Transaction not yet found on chain');
  }

  return signature;
}

try {
  const signature = await retry(getSignature, { retries, delay });
  return signature;
} catch {
  // noop
}

return null;
```

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,216 @@ export function fetchOptions(chain: Chain) {
...
-  const tx = await wallet.signTransaction(unsignedTx as Transaction);
-  const serializedTx = tx.serialize();
-  const sendOptions = {
-    skipPreflight: true,
-    maxRetries: 0,
-    preFlightCommitment: commitment, // See PR and linked issue for why setting this matters: https://github.com/anza-xyz/agave/pull/483
-  };
-  signature = await connection.sendRawTransaction(serializedTx, sendOptions);
-  confirmTransactionPromise = connection.confirmTransaction(
+  const signature = await connection.sendRawTransaction(
```
priyajeet: Check if this is needed, if retry() calls on leading edge

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,216 @@ export function fetchOptions(chain: Chain) {
...
-      );
-      try {
-        await connection.sendRawTransaction(serializedTx, sendOptions);
-      } catch (e) {
-        console.error('Failed to resend transaction:', e);
-      }
-    }
+    const RETRY_DELAY = 1000;
+
+    await retry(
```
priyajeet: Should definitely try to reduce the function arrowness/nesting here. Makes it hard to keep track of { }

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,216 @@ export function fetchOptions(chain: Chain) {
...
+      () => {
```
priyajeet: Can move this out to its own function, so that retry() is just calling that, similar to the other comment. If you need to build a closure, it can still be gotten like

```ts
getConfirmTransaction(connection) {
  async function confirmTransaction() {
    // the body
  }

  return confirmTransaction;
}
```

src/utils/wallet/solana.ts
priyajeet: We should clean it up, there is way too much try {} catch {} nesting happening making the code pretty unreadable and prone to bugs.

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,216 @@ export function fetchOptions(chain: Chain) {
...
+            );
+
+            const isTransactionSuccessful =
+              !confirmTransactionPromise.value.err;
+
+            if (isTransactionSuccessful) {
+              return currentSignature;
+            }
+
+            if (!isTransactionSuccessful) {
```
priyajeet: Is this check useful? Considering line 170 above

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,216 @@ export function fetchOptions(chain: Chain) {
...
+                'Transaction confirmation failed, resending transaction...',
+              );
+
+              try {
+                currentSignature = await connection.sendRawTransaction(
+                  transaction.serializedTransaction,
+                  transaction.sendOptions,
+                );
+
+                const retriedTx = await connection.confirmTransaction(
```
priyajeet: connection.sendRawTransaction and connection.confirmTransaction are done twice, can it be simplified by moving them to their own mini functions and re-used?

src/utils/wallet/solana.ts
priyajeet: If desired, can also move out the getSignature at top level

```ts
function getSignatureGetter(connection) {
  async function getSignature() {
    const tx = await connection.getTransaction(signature, {
      commitment: 'confirmed',
      maxSupportedTransactionVersion: 0,
    });

    if (!tx) {
      throw new Error('Transaction not yet found on chain');
    }

    return signature;
  }

  return getSignature;
}
```

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,216 @@ export function fetchOptions(chain: Chain) {
...
+                console.error('Failed to resend transaction:', resendError);
+
+                // If resend failed, still throw the original confirmation error
+                let errorMessage = `Transaction failed: ${confirmTransactionPromise.value.err}`;
+                if (typeof confirmTransactionPromise.value.err === 'object') {
+                  try {
+                    errorMessage = `Transaction failed: ${JSON.stringify(
+                      confirmTransactionPromise.value.err,
+                      (_key, value) =>
+                        typeof value === 'bigint' ? value.toString() : value,
```
priyajeet: Can move to a utility and re-used here and https://github.com/wormhole-foundation/wormhole-connect/blob/ef0d70646f74cab908d86e48c17649b6c94e2650/src/utils/inProgressTxCache.ts#L11

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,216 @@ export function fetchOptions(chain: Chain) {
...
+                if (!retriedTx.value.err) {
+                  return currentSignature;
+                }
+                throw new Error(`Resent tx failed: ${retriedTx.value.err}`);
+              } catch (resendError) {
+                console.error('Failed to resend transaction:', resendError);
+
+                // If resend failed, still throw the original confirmation error
+                let errorMessage = `Transaction failed: ${confirmTransactionPromise.value.err}`;
+                if (typeof confirmTransactionPromise.value.err === 'object') {
```
priyajeet: Can reduce 1 level of nesting by inverting this check.
Also `typeof null === "object"`

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,216 @@ export function fetchOptions(chain: Chain) {
...
+
+            const isTransactionSuccessful =
+              !confirmTransactionPromise.value.err;
+
+            if (isTransactionSuccessful) {
+              return currentSignature;
+            }
+
+            if (!isTransactionSuccessful) {
+              console.log(
```
priyajeet: Maybe change to .debug()

src/utils/wallet/solana.ts
bradleyplaydon: Good shout

src/utils/wallet/solana.ts
```diff
@@ -81,122 +81,216 @@ export function fetchOptions(chain: Chain) {
...
+    serializedTransaction: Uint8Array | Buffer | number[];
+    sendOptions?: SendOptions;
+  },
+  connection: Connection,
+  blockHash: string,
+  lastValidBlockHeight: number,
+  commitment?: Commitment,
+) {
+  let currentSignature = signature;
+  let confirmTransactionPromise = await connection.confirmTransaction(
```
kev1n-peters: We should not use `await` here, since `confirmTransaction` blocks until the transaction is either confirmed or expired, and we won’t be resending the transaction in the meantime. Instead, we should call `confirmTransaction` *once* in this function and periodically check whether it has resolved or rejected *while also* resending the transaction. This improves the chances of the transaction being confirmed during periods of high congestion on Solana. Why are we changing this from the previous approach? I don’t think I understand what issue this PR is meant to address. See this article for why the previous approach is implemented: https://docs.triton.one/chains/solana/cascade/sending-txs

src/utils/wallet/solana.ts
kev1n-peters: For the first thing, the first promise is awaited immediately when we call `Promise.race()`. Either `confirmTransactionPromise` or the timeout will resolve first. If the timeout resolved first, then we resend the transaction and the loop repeats until `confirmTransactionPromise` resolves or rejects.

src/utils/wallet/solana.ts
bradleyplaydon: Wait but this makes sense no? Why would you want to resend the transaction when you've not confirmed that the transactions expired or confirmed? Won't you result in potentially two tx's if you send before confirming the transaction didn't go through?

The previous code did the following:

The tx would be resent but the response from resending wasn't stored, it also as I mentioned above because it was in a timeout the first confirmation check wouldn't happen until after 5 seconds blocking the transaction from being confirmed resulting in the UI not doing anything for 5 seconds everytime. Try a swap on Solana - this massively speeds things up

src/utils/wallet/solana.ts
bradleyplaydon: You'll see if you do a same chain swap on Solana, the tx isn't confirmed everytime until 5 seconds after the tx is sent making the UI feel stuck because you confirm and have to wait 5 seconds until anything happens

src/utils/wallet/solana.ts
bradleyplaydon: This still abides by the triton docs?

kev1n-peters: We are resending the same transaction, it would not result in two independent transactions. The resent transactions all have the same signature as the original.

src/utils/wallet/solana.ts
kev1n-peters: The amount of time it takes for the tx to confirm depends on current Solana network conditions (e.g. congestion and current priority fees), not the timer. It could take up to a minute for the transaction to confirm in the worst case

src/utils/wallet/solana.ts
bradleyplaydon: Ahhh okay this makes sense now. Will make sure we keep the same logic but just more explicit in code so it's easier to understand

src/utils/wallet/solana.ts
bradleyplaydon: I think the issue here was just the 5 second timeout was running on first confirmation instead of the subsequent confirmation calls

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
-  const tx = await wallet.signTransaction(unsignedTx as Transaction);
-  const serializedTx = tx.serialize();
-  const sendOptions = {
-    skipPreflight: true,
-    maxRetries: 0,
-    preFlightCommitment: commitment, // See PR and linked issue for why setting this matters: https://github.com/anza-xyz/agave/pull/483
-  };
-  signature = await connection.sendRawTransaction(serializedTx, sendOptions);
-  confirmTransactionPromise = connection.confirmTransaction(
+  const confirmationPromiseTimer = 3_000; // How long to wait for confirmation before resending
```
priyajeet: Can move to the top of file as a const?

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
-  const commitment = options?.commitment ?? 'finalized';
+  const commitment = options?.commitment ?? 'confirmed';
   const connection = new Connection(rpc);
+
   const { blockhash, lastValidBlockHeight } =
     await connection.getLatestBlockhash(commitment);
 
-  const unsignedTx = await setPriorityFeeInstructions(
+  // Create tx
+  const { serializedTransaction, sendOptions } = await createSolanaTransaction(
```
priyajeet, line 137: This call isn't used in this function other than just passing into the next. Might as well move it in there?

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
+  return { serializedTransaction: serializedTx, sendOptions };
+}
+
+function formatConfirmationError(err: TransactionError): string {
+  if (!err) return 'Unknown error';
+
+  if (typeof err === 'object') {
+    try {
+      return JSON.stringify(err, (_key, value) =>
+        typeof value === 'bigint' ? value.toString() : value,
```
priyajeet: can use src/utils/inProgressTxCache.ts/JSONReplacer function (maybe move it to a better place)

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
+ * by polling until it appears on chain.
+ *
+ * @param connection Solana RPC connection
+ * @param signature The transaction signature to look up
+ * @param retries Number of retries (default: 5)
+ * @param delay Delay between retries in ms (default: 2000)
+ * @returns The signature once found, or throws if retries are exhausted
+ */
+async function recoverBlockheightExceededTransaction(
+  e: unknown,
```
priyajeet: unused?

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
+  connection: Connection,
+  signature: string,
+  { retries = 5, delay = 2000 }: { retries?: number; delay?: number } = {},
+): Promise<string | null> {
+  const findTransaction = async (): Promise<string> => {
+    const tx = await connection.getTransaction(signature, {
+      commitment: 'confirmed',
```
priyajeet: Does commitment need to be passed in from the top function or was this supposed to be hardcoded to confirmed

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
-    // This loop will break once the transaction has been confirmed or the block height is exceeded.
-    // An exception will be thrown if the block height is exceeded by the confirmTransactionPromise.
-    // The transaction will be resent if it hasn't been confirmed after the interval.
-    const txRetryInterval = 5000;
-    while (!confirmedTx) {
-      confirmedTx = await Promise.race([
+    while (!isTransactionConfirmed) {
+      isTransactionConfirmed = await Promise.race([
         confirmTransactionPromise,
         new Promise<null>((resolve) =>
```
priyajeet, line 198: Can use sleep  from 'utils'

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
-          });
-          if (tx) return signature; // Transaction actually landed
-        } catch {
-          // Ignore errors, we will retry
-        }
-        await sleep(retryDelay);
+  } catch (e) {
+    if (e instanceof Error) {
+      if (e.name === 'TransactionExpiredBlockheightExceededError') {
+        recoverBlockheightExceededTransaction(e, connection, signature);
```
priyajeet: The function returns, but nothing is being done with the return?

src/utils/wallet/solana.ts
priyajeet: Also curious, if we are already in a retry while loop on top, why do we need to retry down here for errors? Won't the loop already take care of this?

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
+  let isTransactionConfirmed: RpcResponseAndContext<SignatureResult> | null =
+    null;
   try {
-    // This loop will break once the transaction has been confirmed or the block height is exceeded.
-    // An exception will be thrown if the block height is exceeded by the confirmTransactionPromise.
-    // The transaction will be resent if it hasn't been confirmed after the interval.
-    const txRetryInterval = 5000;
-    while (!confirmedTx) {
-      confirmedTx = await Promise.race([
+    while (!isTransactionConfirmed) {
```
priyajeet, line 195: Lets say isTransactionConfirmed fails due to condition below, what updates it in the loop after? `confirmTransactionPromise` is never reset?

src/utils/wallet/solana.ts
priyajeet: Also can we verify when `TransactionExpiredBlockheightExceededError` is gotten, the transaction can actually be retried? Per google/claude/chatgpt it cannot be and it needs re-signing by the wallet. Not sure if we should have that check. Don't see it being done on the executor git repo (src/handlers/svm/utils.ts#L276).

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
+): Promise<{
+  signature: string;
+  confirmPromise: Promise<RpcResponseAndContext<SignatureResult>>;
+}> {
+  const signature = await connection.sendRawTransaction(
+    transaction.serializedTransaction,
+    transaction.sendOptions,
+  );
+
+  const confirmPromise = connection.confirmTransaction(
```
priyajeet: This method is deprecated, should we use the replacement `getSignatureStatuses`

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
+ * @returns The signature once found, or throws if retries are exhausted
+ */
+async function recoverBlockheightExceededTransaction(
+  e: unknown,
+  connection: Connection,
+  signature: string,
+  { retries = 5, delay = 2000 }: { retries?: number; delay?: number } = {},
+): Promise<string | null> {
+  const findTransaction = async (): Promise<string> => {
+    const tx = await connection.getTransaction(signature, {
```
priyajeet: This method is deprecated, should we use the replacement `getSignatureStatuses`

src/utils/wallet/solana.ts
```diff
@@ -102,125 +105,248 @@ export function fetchFogoOptions() {
...
+  },
+  connection: Connection,
+  blockHash: string,
+  lastValidBlockHeight: number,
+  commitment: Commitment,
+): Promise<{
+  signature: string;
+  confirmPromise: Promise<RpcResponseAndContext<SignatureResult>>;
+}> {
+  const signature = await connection.sendRawTransaction(
```
priyajeet: Can line 201-213 just move into the loop below at the top? So that we send the transaction, check confirmation, sleep, repeat?

src/utils/wallet/solana.ts
bradleyplaydon: Splits up the logic so it's easier to process when reading imo being here.

src/utils/wallet/solana.ts
bradleyplaydon: Annoyingly this method getTransaction is a different type than 'Commitment' it uses a 'Finality' type which doesn't work for the other methods, so yeah supposed to be hardcoded to confirmed.

This is only for the tx exceeded block height since it could still of landed or be landing

src/utils/wallet/solana.ts
bradleyplaydon: It's not about being retried for this error, it's about verifying whether the tx actually lands on chain. This just checks a few times to see if it did land even though we got this error.

> The transaction might still have landed in a block just before expiry, but your RPC node didn’t return confirmation fast enough.

src/utils/wallet/solana.ts
bradleyplaydon: It's not but because the signature is the same as the transaction params are the same the new send still updates the promise.

We can make it more explicit by overriding the promise after sending so it's more explicit and makes more sense reading otherwise I agree it's confusing and it tripped me up too

Will add the promise and first send into the function to make things clearer

src/utils/wallet/solana.ts
bradleyplaydon: One for another day, might end up requiring another rewrite

src/utils/wallet/solana.ts
bradleyplaydon: See above

src/utils/wallet/solana.ts
bradleyplaydon: We can but this splits up the functions abit more and makes it easier to follow whats going on, also could be a use case for wanting this too methods together for other things.

For now i'll bring it in to make things clearer.

src/utils/index.ts
```diff
@@ -424,3 +424,9 @@ export const isExecutorRoute = (route: string | undefined) => {
   }
   return route.endsWith('ExecutorRoute');
 };
+
+export const JSONReplacer = (json: any) => {
```
priyajeet: A more appropriate name now would be `stringifyWithBigInt()` since the replacer part is just the `(_key, value) => {}`

src/utils/wallet/solana.ts
```diff
@@ -102,125 +108,194 @@ export function fetchFogoOptions() {
...
+    sendOptions?: SendOptions;
+  },
+  blockhash: string,
+  lastValidBlockHeight: number,
+  commitment: Commitment,
+): Promise<string> {
+  let isTransactionConfirmed: RpcResponseAndContext<SignatureResult> | null =
+    null;
+
+  const signature = await connection.sendRawTransaction(
```
priyajeet, line 184: Why not move this in the loop on top and remove line 208-210

src/utils/wallet/solana.ts
```diff
@@ -102,125 +108,194 @@ export function fetchFogoOptions() {
...
-            commitment: 'confirmed',
-            maxSupportedTransactionVersion: 0,
-          });
-          if (tx) return signature; // Transaction actually landed
-        } catch {
-          // Ignore errors, we will retry
-        }
-        await sleep(retryDelay);
+  } catch (e) {
+    if (e instanceof Error) {
```
priyajeet: Don't think you need this nested.
It can be outside like

```ts
if (!isTransactionConfirmed) {
  await recoverBlockheightExceededTransaction(connection, signature);
}
```

priyajeet: And you need to try catch this also, since retry will throw on failed retries

src/utils/wallet/solana.ts
priyajeet: just use withTimeout() from es-toolkit

try {
isTransactionConfirmed = await withTimeout(confirmPromise, CONFIRMATION_PROMISE_TIMER)
} catch { noop }

src/utils/wallet/solana.ts
bradleyplaydon: We do no? We only want to do these RPC calls for getting tx TransactionExpiredBlockheightExceededError, if we do it generically we waste RPC calls?

if (!isTransactionConfirmed) {
  await recoverBlockheightExceededTransaction(connection, signature);
}

src/utils/wallet/solana.ts
bradleyplaydon: Means having to change the function return signature no and then having to use a let?

src/utils/wallet/solana.ts
```diff
@@ -102,125 +108,194 @@ export function fetchFogoOptions() {
...
+      if (e.name === 'TransactionExpiredBlockheightExceededError') {
+        await recoverBlockheightExceededTransaction(connection, signature);
       }
+      console.error('Failed to resend transaction:', e);
     }
-    throw e;
```
priyajeet, line 203: add back

bradleyplaydon: Done this and added the try catch back too inside recover, so it's same as original

priyajeet (approved):

src/utils/inProgressTxCache.ts
```diff
@@ -1,15 +1,13 @@
 import config from 'config';
 import type { TransactionLocal } from 'config/types';
-import { isEmptyObject } from 'utils';
+import { isEmptyObject, stringifyWithBigInt } from 'utils';
 
 const LOCAL_STORAGE_KEY = 'transactions:inprogress';
 const LOCAL_STORAGE_MAX = 3;
 
 // Bigint types cannot be serialized to a string
```
kev1n-peters, line 8: Should we remove this comment (or move it to the stringifyWithBigInt function?)

src/utils/wallet/solana.ts
```diff
@@ -102,125 +108,194 @@ export function fetchFogoOptions() {
...
-            maxSupportedTransactionVersion: 0,
-          });
-          if (tx) return signature; // Transaction actually landed
-        } catch {
-          // Ignore errors, we will retry
-        }
-        await sleep(retryDelay);
-      }
+      await recoverBlockheightExceededTransaction(connection, signature);
+      return signature;
```
kev1n-peters, line 219: We should only return the `signature` if the transaction was actually confirmed. To preserve the functionality from before, we could do:
```
const recoveredSignature = await recoverBlockheightExceededTransaction(connection, signature);
if (recoveredSignature) return recoveredSignature;
throw e;
```
We would then change `recoverBlockheightExceededTransaction` to return Promise<string | null>

PR #3822, opened 2025-09-25
Prod 575/add hypercore

kyleleighton:


src/config/mainnet/chains.ts
```diff
@@ -217,6 +217,14 @@ export const MAINNET_CHAINS: ChainsConfig = {
     icon: 'HyperEVM',
     symbol: 'HYPE',
   },
+  HyperCore: {
+    displayName: 'HyperCore',
+    sdkName: 'HyperCore',
+    explorerUrl: 'https://arbiscan.io/', // Uses Arb USDC
```
priyajeet: 1 thing to check, once we are on the receipt page in portal/connect, how do the links work there. Because on mayan page, they link to hyperliquid's explorer. But the token that we show on the asset picker, links to ARB usdc.

src/routes/mayan/MayanRouteBase.ts
```diff
@@ -140,8 +146,12 @@ export class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
     return ['Mainnet', 'Testnet'];
   }
 
-  static supportedChains(network: Network): Chain[] {
-    return supportedChains(network);
+  static isEvmChain(chain: Chain): boolean {
```
priyajeet: this can move to src/utils/evm, since its not mayan specific. Double check if this ever throws.

src/routes/mayan/MayanRouteBase.ts
```diff
@@ -150,14 +160,40 @@ export class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
     return supportedChains(chain.network).includes(chain.chain);
   }
 
+  /**
+   * Check if a token on a specific chain is USDC
+   */
+  private static isUSDCToken(chain: Chain, tokenAddress: string): boolean {
```
priyajeet: This can also move to src/utils/tokenHelpers

kev1n-peters: If this function is meant to check if the token is a canonical USDC deployment, then we should be checking the (chain, address) pair against these: https://github.com/wormhole-foundation/wormhole-sdk-ts/blob/152b6fb02df1ad1ea735b241ccb5beb18ae8cc6b/core/base/src/constants/circle.ts#L13. Somebody could override a token's symbol to USDC, or a token might have USDC as the symbol, but it's not canonical (i.e. deployed by Circle)

src/utils/hypercore.ts
```diff
@@ -0,0 +1,141 @@
...
+        valid: false,
+        params,
+        error: new routes.UnavailableError(
+          new Error('HyperCore only supports USDC as destination token'),
+        ),
+      };
+    }
+  }
+
+  if (isHyperCoreChain(fromChain.chain)) {
```
priyajeet: This should just return line 71, there should be no source token check. HyperCore cannot be a source chain for any token

src/utils/hypercore.ts
```diff
@@ -0,0 +1,141 @@
...
+ */
+export function validateHyperCoreTransfer<N extends Network>(
+  request: routes.RouteTransferRequest<N>,
+  params: TransferParams,
+  validators: HyperCoreValidators,
+): ValidationResult | null {
+  const { fromChain, toChain, source, destination } = request;
+  const { isEvmChain, isCanonicalUSDCToken } = validators;
+
+  if (isHyperCoreChain(toChain.chain)) {
```
priyajeet: Probably add 1 more check in here to throw error when network isn't mainnet

src/utils/hypercore.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/**
+ * Enforces HyperCore routing rules: only EVM -> HyperCore flows with USDC
+ * inbound, and disallow HyperCore-originated USDC.
+ */
+export function validateHyperCoreTransfer<N extends Network>(
+  request: routes.RouteTransferRequest<N>,
+  params: TransferParams,
+  validators: HyperCoreValidators,
```
priyajeet: No need to pass in validators, they should come from respective co-located utils/XXX files

src/views/v3/Bridge/index.tsx
```diff
@@ -264,7 +266,10 @@ function Bridge(props: BridgeProps) {
 
   // Supported chains for the source network
   const supportedSourceChains = useMemo(() => {
-    return getFilteredChains(supportedChains, destChain);
+    // Prevent HyperCore from appearing as a source since it's destination-only
+    return getFilteredChains(supportedChains, destChain).filter(
```
priyajeet: Might be worth to just change getFilteredChains() to take an arg saying isSource, and filter there. Otherwise we will repeat.

src/views/v3/Bridge/index.tsx
```diff
@@ -254,7 +253,10 @@ function Bridge(props: BridgeProps) {
 
   const sourceTokens = useMemo(() => {
     if (sourceChain) {
-      return config.tokens.getAllForChain(sourceChain);
+      // Filter out HyperCore tokens from source list since it's destination-only
+      return config.tokens
+        .getAllForChain(sourceChain)
+        .filter((token) => token.nativeChain !== 'HyperCore');
```
priyajeet: Probably not needed when sourceChain will get already filtered out?

src/views/v2/Bridge/index.tsx
```diff
@@ -246,7 +248,10 @@ const Bridge = () => {
 
   // Supported chains for the source network
   const supportedSourceChains = useMemo(() => {
-    return getFilteredChains(supportedChains, destChain);
+    // Prevent HyperCore from appearing as a source since it's destination-only
```
priyajeet: same

src/utils/wallet/evm.ts
```diff
@@ -14,12 +14,16 @@ import type {
...
 } from '@wormhole-foundation/sdk-evm';
-import type { Network } from '@wormhole-foundation/sdk';
+import type { Chain, Network } from '@wormhole-foundation/sdk';
+import { chainToPlatform } from '@wormhole-foundation/sdk';
 
 import config from 'config';
 import * as ethers from 'ethers';
 import { sleep } from 'utils';
 
+export const isEvmChain = (chain: Chain): boolean =>
```
priyajeet: I would move this to a new file evm.ts 1 level above, since this is not wallet related

src/routes/mayan/MayanRouteBase.ts
```diff
@@ -140,24 +150,38 @@ export class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
...
   static isProtocolSupported<N extends Network>(
     chain: ChainContext<N>,
   ): boolean {
     return supportedChains(chain.network).includes(chain.chain);
   }
 
+  /**
+   * Filter a list of tokens to only include USDC tokens
+   */
+  private static filterUSDCTokens(chain: Chain, tokens: TokenId[]): TokenId[] {
```
priyajeet: Not needed, as discussed before, we can just return the USDC toke

```ts
export function getUSDCTokenId(
  chain: Chain,
  network: Network,
): TokenId | undefined {
  const usdcContract = circle.usdcContract.get(network, chain);

  if (!usdcContract) {
    return undefined;
  }

  return Wormhole.tokenId(chain, usdcContract);
}

export function isUSDCToken(
  chain: Chain,
  network: Network,
  tokenAddress: string,
): boolean {
  const usdc = getUSDCTokenId(chain, network);
  return !!usdc && tokenAddress === canonicalAddress(usdc);
}
```

and then

```ts
  static async supportedDestinationTokens<N extends Network>(
    _token: TokenId,
    fromChain: ChainContext<N>,
    toChain: ChainContext<N>,
  ): Promise<TokenId[]> {
    // For HyperCore, only allow USDC as destination token
    if (isHyperCoreChain(toChain.chain)) {
      const usdc = getUSDCTokenId(toChain.chain, "Mainnet");
      return [usdc];
    }

    const tokens = getAllTokenIdsForChain(toChain.chain);

    return tokens;
  }
```

src/routes/mayan/MayanRouteBase.ts
```diff
@@ -140,24 +150,38 @@ export class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
...
     toChain: ChainContext<N>,
   ): Promise<TokenId[]> {
-    // TODO update this function to only allow HyperCore/USDC
-    return getAllTokenIdsForChain(toChain.chain);
+    const tokens = getAllTokenIdsForChain(toChain.chain);
+
+    // For HyperCore, only allow USDC as destination token
+    if (isHyperCoreChain(toChain.chain)) {
+      if (!isEvmChain(fromChain.chain)) {
+        return [];
```
priyajeet: This particular check may not be needed if the destination chain gets filtered out, since the chain won't show up in the list. But double check.

src/utils/tokenHelpers.ts
```diff
@@ -16,3 +17,23 @@ export const getAllTokenIdsForChain = memoize(
       `${String(args[0])}|${config.tokens.lastUpdate.getTime()}`,
   },
 );
+
+const normalizeAddress = (value: string): string => {
+  return value.startsWith('0x') ? value.toLowerCase() : value;
+};
+
+export const isCanonicalUSDCToken = (
```
priyajeet: Maybe these (and any specific usdc stuff) can be moved to utilt/usdc.ts so that the next person knows where to get usdc stuff

src/utils/wallet/evm.ts
bradleyplaydon: > I would move this to a new file evm.ts 1 level above, since this is not wallet related

Do we not already have this util - think i've seen this done before

src/routes/sdkv2/signer.ts
```diff
@@ -139,6 +148,112 @@ export class SDKv2Signer<N extends Network, C extends Chain>
     return txHashes;
   }
 
+  async signTypedData(
```
priyajeet: Remove this function and use signTypedData() available from the wallet via the signer

src/routes/sdkv2/signer.ts
priyajeet, line 42: Expose this via a getter like

provider() {
    return this._walletProvider;
}

src/utils/hypercore.ts
```diff
@@ -0,0 +1,141 @@
...
+      destinationAddress,
+      arbitrumRpc,
+    ));
+  } catch (e) {
+    throw new Error(
+      `Failed to fetch HyperCore USDC permit params: ${(e as Error).message}`,
+    );
+  }
+
+  if (typeof (signer as any).signTypedData === 'function') {
```
priyajeet, line 175: Replace lines 129-140 with below (with better error messages)

```
  const wallet = signer
    .provider()
    .getWallet('Arbitrum', TransferWallet.SENDING) as Eip6963Wallet;

  if (!wallet) {
    throw new Error('something something');
  }

  try {
    await wallet.switchChain(42161);
  } catch {
    throw new Error('something something');
  }

  const nativeSigner = await wallet.getSigner();

  return await nativeSigner.signTypedData(domain, types, value);
```

PR #3823, opened 2025-09-26
chore: Adding development option to build lib

emreboga:
This will be utilized when pulling Connect into Portal for an easier local development and debugging.


emreboga:
build:minimal already have `sourcemaps`.

PR #3825, opened 2025-09-28
LiFi Tx History

emreboga:
This PR integrates LiFi as a third transaction history source alongside existing Mayan and WHScan hooks and adds comprehensive unit tests for all transaction history hooks.
Fixes PROD-723

**Future improvements:** Tx History view started with the list of transactions from WHScan APIs first. We've added Mayan transactions from their own API. This introduces the problem of time sorting of items from different paginated APIs. We have previously solved this for two APIs and this PR adds a third one. We may need to revisit this problem and find a more scalable solution.


vitest.config.mjs
emreboga, line 1: Extension`.mjs` forces Node.js to treat the file as an ES module, resolving the compatibility issue between:
- vitest's config loader (expecting CommonJS)
- vite's ES module exports

src/config/index.ts
```diff
@@ -105,6 +105,7 @@ export function buildConfig(
     rpcs,
     evmIndexers: customConfig.evmIndexers,
     mayanApi: 'https://explorer-api.mayan.finance',
+    lifiApi: 'https://li.quest',
```
kev1n-peters: Can we change this to lifiExplorerUrl? I know you're keeping it consistent with the other properties here, but mayanApi is misleading as it’s an explorer link, not an API URL. We’ll need to pass in a custom URL for the lifiApi in the future (different from this explorer URL), so would like to avoid any confusion.

emreboga: Sure, np!

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/71a33f037d6047ae68f574f0c6abd072d8bbe506

src/hooks/useTransactionHistoryWHScan.ts
```diff
@@ -157,7 +157,7 @@ const useTransactionHistoryWHScan = (
         token = await getOrFetchToken(
           Wormhole.tokenId(tokenChain, standarizedProperties.tokenAddress),
         );
-      } catch (e) {
+      } catch (_e) {
```
priyajeet, line 160: You can remove `(_e)`

src/views/v3/TxHistory/index.tsx
```diff
@@ -160,7 +144,33 @@ const TxHistory = () => {
...
         </Box>
-      )}
+      );
+    }
+
+    // No transactions found (after fetch completed)
+    if (!isFetching && (!transactions || transactions.length === 0)) {
+      return (
+        <Typography color={theme.palette.text.secondary} textAlign="center">
+          {`No transactions found for the wallet ${trimAddress(
```
priyajeet: outer `{ }` may not be needed, since react can already do substitution

src/views/v3/TxHistory/index.tsx
```diff
@@ -119,33 +107,29 @@ const TxHistory = () => {
...
-        transactionList
-      ) : (
+  const content = useMemo(() => {
+    // Has transactions to display
+    if (transactions?.length) {
+      return transactionList;
+    }
+
+    // Loading state
+    if (isFetching || (transactions && transactions.length === 0)) {
```
priyajeet: transactions.length > 0 already checked above

src/views/v3/TxHistory/index.tsx
```diff
@@ -160,7 +144,33 @@ const TxHistory = () => {
           <Skeleton variant="rounded" height={48} width="100%" />
           <Skeleton variant="rounded" height={48} width="100%" />
         </Box>
-      )}
+      );
+    }
+
+    // No transactions found (after fetch completed)
+    if (!isFetching && (!transactions || transactions.length === 0)) {
```
priyajeet: same, and might as well use transactions?. to keep it more readable like 126/122

src/views/v3/TxHistory/index.tsx
```diff
@@ -97,20 +97,8 @@ const TxHistory = () => {
...
-        <Typography color={theme.palette.text.secondary} textAlign="center">
-          {`No transactions found for the wallet ${trimAddress(
-            sendingWallet.address,
-          )}`}
-        </Typography>
-      );
-    }
-
-    return (
+  const transactionList = useMemo(
```
priyajeet: Should be moved to `TransactionList.tsx`

src/views/v3/TxHistory/index.tsx
```diff
@@ -119,33 +107,29 @@ const TxHistory = () => {
...
   }, [mobile, styles.container, styles.containerMobile]);
 
-  return (
-    <Box sx={{ ...containerStyles }}>
-      <ConfigurablePageHeader />
-      {txHistoryHeader}
-      {transactions && transactions.length && !isFetching ? (
-        transactionList
-      ) : (
+  const content = useMemo(() => {
```
priyajeet: Should be moved to `TransactionListContent.tsx`

src/hooks/useTransactionHistory.ts
```diff
@@ -125,48 +169,71 @@ const useTransactionHistory = (
...
+          item: mayanItem,
+          source: 'mayan',
+          time: mayanItem ? new Date(mayanItem.senderTimestamp) : null,
+        },
+        {
+          item: lifiItem,
+          source: 'lifi',
+          time: lifiItem ? new Date(lifiItem.senderTimestamp) : null,
+        },
+      ].filter((i) => i.item !== undefined);
```
priyajeet: What is this line doing

src/hooks/useTransactionHistory.ts
```diff
@@ -125,48 +169,71 @@ const useTransactionHistory = (
...
+
+      // Sort by timestamp (most recent first)
+      items.sort((a, b) => {
```
priyajeet: Probably move the sorter function to a common util file, I think we might have this in a couple of other places

src/hooks/useTransactionHistoryMayan.ts
```diff
@@ -164,7 +164,6 @@ const useTransactionHistoryMayan = (
         if (res.status !== 200) {
           setTransactions([]);
           setHasMore(false);
-          setIsFetching(false);
```
emreboga, line 167: We reset isFetching in the finally block down below, no need to do it there.

vitest.config.mjs
priyajeet: Can it be due to tsconfig not having nodenext / esnext as module resolution? 
portal has it in a .ts file

src/hooks/useTransactionHistoryLiFi.ts
```diff
@@ -0,0 +1,328 @@
+import { useCallback, useEffect, useState } from 'react';
```
bradleyplaydon, line 1: Slightly concerned about this hook, is this a copy paste of another hook but renamed?

Are the interfaces the same in other places too?

src/hooks/useTransactionHistoryLiFi.ts
```diff
@@ -0,0 +1,328 @@
...
+  const [transactions, setTransactions] = useState<
+    Array<Transaction> | undefined
+  >();
+  const [error, setError] = useState('');
+  const [isFetching, setIsFetching] = useState(false);
+  const [hasMore, setHasMore] = useState(true);
+
+  const { address, page = 0, pageSize = 30 } = props;
+
+  const parseSingleTx = useCallback(
```
priyajeet: I think this is a side effects free function glancing at it. Can be moved out into a utils/lifi.ts file and imported?

src/hooks/useTransactionHistoryLiFi.ts
```diff
@@ -0,0 +1,328 @@
...
+        : undefined;
+
+      if (!fromChain || !toChain) {
+        return undefined;
+      }
+
+      let fromToken: Token | undefined;
+      let toToken: Token | undefined;
+
+      try {
```
priyajeet: finding a fromToken and toToken can be extracted to the utils/tokens file. I think this logic may be used in other places.

Also can have just 1 function that finds the token and just pass it from/to

src/hooks/useTransactionHistory.ts
bradleyplaydon: agreed, this is probably something used somewhere else i would imagine

src/hooks/useTransactionHistoryLiFi.ts
emreboga: Yup, you can checkout the future improvement in the PR description, which also gives a bit of history and how ended up here.

src/hooks/useTransactionHistoryLiFi.ts
```diff
@@ -0,0 +1,328 @@
...
+              }
+              return parsedTxs;
+            });
+
+            // LiFi returns max 1000 results, if we get less than pageSize, no more data
+            if (resData.length < pageSize) {
+              setHasMore(false);
+            }
+          } else {
+            setTransactions([]);
```
priyajeet: setting [] to [] will trigger a re-render
correct way `setTransactions(current => current.length === 0 ? current : []);`

src/hooks/useTransactionHistoryLiFi.ts
```diff
@@ -0,0 +1,328 @@
...
+          setHasMore(false);
+          return;
+        }
+
+        let resPayload;
+        try {
+          resPayload = await res.json();
+        } catch (_e) {
+          setError('Invalid response from LiFi service');
+          setTransactions([]);
```
priyajeet: same. Ideally just have a resetTransactions() function outside and call that

src/hooks/useTransactionHistoryLiFi.test.tsx
```diff
@@ -0,0 +1,296 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { renderHook, waitFor } from '@testing-library/react';
+
+const mockToken = {
```
bradleyplaydon, line 4: Do we not have a mocks file this can go in?

src/hooks/useTransactionHistoryLiFi.ts
```diff
@@ -0,0 +1,328 @@
...
+          if (res.status === 429) {
+            setError('Rate limit exceeded. Please try again later.');
+          } else if (res.status >= 500) {
+            setError('LiFi service is temporarily unavailable.');
+          } else if (res.status !== 404) {
+            // 404 just means no transactions, not an error
+            setError('Failed to fetch LiFi transactions.');
+          }
+
+          setTransactions([]);
```
priyajeet: same

src/hooks/useTransactionHistoryLiFi.test.tsx
```diff
@@ -0,0 +1,296 @@
...
+      const value = BigInt(amountObj.amount) / divisor;
+      return value.toString();
+    }),
+  },
+}));
+
+import useTransactionHistoryLiFi from './useTransactionHistoryLiFi';
+
+// Sample test data
+const mockLiFiTransaction = {
```
bradleyplaydon, line 55: same here, mocks file?

bradleyplaydon: Also be good to add LifiTransaction type too incase in the future we update

src/hooks/useTransactionHistoryLiFi.ts
```diff
@@ -0,0 +1,328 @@
...
+
+        const params = new URLSearchParams({
+          wallet: address,
+          fromTimestamp: fromTimestamp.toString(),
+          toTimestamp: toTimestamp.toString(),
+          status: 'ALL',
+        });
+
+        const res = await fetch(
+          `${config.lifiExplorerUrl}/v1/analytics/transfers?${params}`,
```
bradleyplaydon, line 210: Might be worth breaking the string out into a const again if Lifi updates there endpoint versioning just makes it easy to know what version we are on

src/hooks/useTransactionHistoryLiFi.ts
```diff
@@ -0,0 +1,328 @@
...
+      if (!hasMore && page > 0) {
+        return;
+      }
+
+      setIsFetching(true);
+
+      try {
+        // LiFi API doesn't have traditional pagination,
+        // but we can simulate it using timestamp filtering
+        const now = Date.now() / 1000; // Current time in seconds
```
bradleyplaydon, line 189: Do we have date-fns or something? Instead

vitest.config.mjs
priyajeet: or type: module in package.json

src/hooks/useTransactionHistoryLiFi.ts
```diff
@@ -0,0 +1,328 @@
...
+          return;
+        }
+
+        let resPayload;
+        try {
+          resPayload = await res.json();
+        } catch (_e) {
+          setError('Invalid response from LiFi service');
+          setTransactions([]);
+          setHasMore(false);
```
bradleyplaydon, line 236: hasMore what :D

vitest.config.mjs
bradleyplaydon: i'd stick to mts just to keep the types flowing

emreboga: Just setting the target and module to ESNext (as portal settings) doesn't do it by itself. I may need to try/test with other TS or vite settings; but considering just the ts -> mjs change works, I'll go with it.

src/hooks/useTransactionHistoryWHScan.ts
emreboga: It expects an identifier. TS settings in Connect might not be the same as in Portal. It can be a separate item to update ES2022 to ESNext in Connect.

src/views/v3/TxHistory/index.tsx
emreboga: Yes and here we check if it's fetching or we have an empty transactions. I can remove `transactions && ` but we need to check the empty array.

src/views/v3/TxHistory/index.tsx
emreboga: `transactions?.length` === 0 and `!transactions || transactions.length === 0` are not the same.
The case where it's not fetching and we never had `transactions` initialized will be lost otherwise.

src/hooks/useTransactionHistory.ts
emreboga: Filtering out undefined, which happened before when APi returns empty results. But we don't need the filter, just won't add to the array if the item is not there.

src/hooks/useTransactionHistoryLiFi.ts
emreboga: You know what :) but I can make it more clear

src/hooks/useTransactionHistoryLiFi.ts
emreboga: I'm reluctant to add a new lib just for a division :) We can evaluate that for the whole Connect if that's going to make things much easier 👍

src/hooks/useTransactionHistoryLiFi.test.tsx
emreboga: Let me check If we can share mocks between unit tests, so having a mocks file will make it easier. Otherwise I prefer having the mock in the test file if it's just used there.

src/hooks/useTransactionHistoryLiFi.ts
emreboga: Isn't it already obvious which version? :) The URL is used in only one place, that's why I didn't want to break it down further. I added the explorer URL to the config because it can be reused to construct Tx links etc.

src/hooks/useTransactionHistoryLiFi.ts
emreboga: Actually, I think maybe leave it as hasMore and from the context I think it's clear what it has more of.

src/hooks/useTransactionHistoryLiFi.test.tsx
emreboga: The only mock that is shared is the token one:
```
const mockToken = {
  key: 'USDC',
  chain: 'Ethereum',
  address: '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48',
  decimals: 6,
  symbol: 'USDC',
  name: 'USD Coin',
};
```
I've created a separate [ticket](https://linear.app/wormholelabs/issue/PROD-807/share-mocks-for-unit-tests-in-connect) for this to go over all sharable mocks in Connect.

src/hooks/useTransactionHistoryLiFi.test.tsx
emreboga: Nope. We didn't have much of unit testing in Connect until recently. We can start moving the shared ones to a single location (added a separate [ticket](https://linear.app/wormholelabs/issue/PROD-807/share-mocks-for-unit-tests-in-connect) for that), but we need to be careful there. Having all those mocks in one place may produce other maintenance issues.

src/hooks/useTransactionHistory.ts
emreboga: We have other places that use sort but all of them are sorting different types of things (tokens, chains, quotes, balances, dates etc.). I will add a utility function to sort by time in this case, but won't expand that to all other Array.sort usage as it's a bit too much for this PR.

src/hooks/useTransactionHistoryLiFi.ts
emreboga: We don't need this to be memoized. I'll remove that but prefer keeping it in the same file as the hook. I don't see this function being utilized anywhere else.

emreboga: Also as I mentioned this in the PR description, the way Tx History works today is not very scalable. It started with WHScan as the single source but we had to use Mayan's own API for performance and reliability (WHScan data was delayed a few mins for Mayan transactions). On top of that now we have a third transaction source. We need to rethink and redesign this for the nth added transaction source. I've added a separate [ticket](https://linear.app/wormholelabs/issue/PROD-808/redesign-transaction-aggregation-for-tx-history-view) for this.

src/views/v3/TxHistory/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/2126847a2e36f039dc5b7835be60560ad3ae65a8

src/views/v3/TxHistory/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/2126847a2e36f039dc5b7835be60560ad3ae65a8

src/hooks/useTransactionHistoryLiFi.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/2126847a2e36f039dc5b7835be60560ad3ae65a8

src/hooks/useTransactionHistoryLiFi.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/2126847a2e36f039dc5b7835be60560ad3ae65a8

src/hooks/useTransactionHistoryLiFi.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/2126847a2e36f039dc5b7835be60560ad3ae65a8

src/views/v3/TxHistory/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/ce7a75cbb271ea1effe6ff6cd0d84405bec65d38

src/views/v3/TxHistory/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/commit/ce7a75cbb271ea1effe6ff6cd0d84405bec65d38

src/views/v3/TxHistory/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/7339b19ad10b05f1b99dcd51aa748e2998c14d02

src/views/v3/TxHistory/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/7339b19ad10b05f1b99dcd51aa748e2998c14d02

src/hooks/useTransactionHistory.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/2126847a2e36f039dc5b7835be60560ad3ae65a8

src/hooks/useTransactionHistoryLiFi.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3825/commits/ecd9065fb81218704fc228eb4471ebfb7a17ceec

vitest.config.mjs
bradleyplaydon: > Just setting the target and module to ESNext (as portal settings) doesn't do it by itself. I may need to try/test with other TS or vite settings; but considering just the ts -> mjs change works, I'll go with it.

I don't want to lose typescript on this config we shouldn't need too. What's the problem with switching to .mts?

src/hooks/useTransactionHistoryLiFi.ts
bradleyplaydon: I think this is more just to make it easier to maintain. If there's other endpoints we use of LIFI's it would be nice to see the versions we use in a consts. 

But less about version being split out and more about the analytics/transfers incase there's somewhere else we need to use this endpoint

src/hooks/useTransactionHistoryLiFi.ts
bradleyplaydon: I'm not sure I still think it's better to be explicit here

hasMoreTransactions makes sense to me

vitest.config.mjs
emreboga: @bradleyplaydon I must have missed your earlier comment, yes mts works :)

src/hooks/useTransactionHistoryLiFi.ts
emreboga: Thanks for the clarification. I get the point about maintainability. Since this URL is only used in a single place right now, I’d prefer to keep it inline to avoid introducing extra indirection prematurely. If we end up adding more LiFi endpoints in the future, or features that will use this endpoint; I agree it’ll make sense to centralize them at that point. For now, I’ll keep this as is.

src/hooks/useTransactionHistoryLiFi.ts
bradleyplaydon: I would avoid the comment however in this case and just make the variable name more obvious like

nowInSeconds

src/hooks/useTransactionHistoryLiFi.ts
emreboga: I get your point about being explicit. My take is that `hasMore` is a pretty common pattern (especially for pagination flags) and in this context it’s clear what it refers to, just like how we name `isFetching` in most (if not all) places. If we later introduce multiple “hasMore” states, then I’d be on-board with renaming it more explicitly. For now, I’d prefer to keep it short and consistent with other transaction history hooks and the usual pagination naming.

src/hooks/useTransactionHistoryLiFi.ts
emreboga: `nowInSeconds` works better 👍

kev1n-peters (approved):

PR #3827, opened 2025-09-29
feat: Add HyperCore

kyleleighton:
This adds Hypercore to the list of supported chains.

1. Hypercore can ONLY be a destination chain
2. The ONLY Hypercore asset is USDC
3. Hypercore does not appear in wallets. The way you verify it is by checking the balance of your _hyperliquid_ perps account at app.hyperliquid.xyz

> Note: HyperCore bridge = Arbitrum bridge + custom payload for USDC deposit to Hyperliquid

Example:
https://explorer.mayan.finance/tx/0xeb0a8f618f13c91e4eff54db4dff3bcbc511526c829e5906a6b0c296676ca926

https://app.hyperliquid.xyz/explorer/address/0x49887a216375fded17dc1aaad4920c3777265614

Screenshot 2025-09-29 at 12 00 56 PM

Screenshot 2025-09-29 at 12 02 46 PM

---
Separate txn

Screenshot 2025-09-29 at 12 56 39 PM


src/utils/sdkv2.ts
```diff
@@ -606,6 +608,10 @@ export function getFilteredChains(
       return false;
     }
 
+    if (isSource && isHyperCoreChain(chain.sdkName)) {
```
priyajeet: Can you add a comment right above, something along the lines of
```
// If we see more exceptions being added like this, a better solution would be to update the route interface in the sdk to have explicit `supportedSourceChains()` and `supportedDestinationChains()` function where routes can specify it directly. Current `supportedChains()` doesn't suffice.
```

priyajeet: + lets move this check above line 607 as this is a faster than .includes()

src/utils/hypercore.ts
```diff
@@ -0,0 +1,187 @@
...
+  const quoteDetails = quote.details;
+
+  if (!quoteDetails) {
+    throw new Error('Missing HyperCore quote details required for permit');
+  }
+
+  const arbitrumRpc = (await request.toChain.getRpc()) ?? null;
+
+  if (!arbitrumRpc) {
+    throw new Error('Could not resolve HyperCore RPC connection');
```
priyajeet: Maybe `Could not resolve Arbitrum RPC connection needed for HyperCore`

src/utils/hypercore.ts
```diff
@@ -0,0 +1,187 @@
...
+    return {
+      valid: false,
+      params,
+      error: new routes.UnavailableError(
+        new Error('HyperCore is only available on Mainnet'),
+      ),
+    };
+  }
+
+  if (!isEvmChain(fromChain.chain)) {
```
priyajeet, line 67: Lets add a comment on top here too that this is temporary as SUI/Solana support is coming soon

src/routes/sdkv2/signer.ts
```diff
@@ -139,11 +148,121 @@ export class SDKv2Signer<N extends Network, C extends Chain>
     return txHashes;
   }
 
+  async signTypedData(
```
priyajeet: This function is not needed anymore, right?

src/utils/hypercore.ts
```diff
@@ -0,0 +1,187 @@
...
+    ));
+  } catch (e) {
+    throw new Error(
+      `Failed to fetch HyperCore USDC permit params: ${(e as Error).message}`,
+    );
+  }
+
+  const signerWithProvider = signer as SignerWithProvider<N>;
+
+  if (typeof signerWithProvider.provider === 'function') {
```
priyajeet: This should always be true I think

src/utils/hypercore.ts
```diff
@@ -0,0 +1,187 @@
...
+    const nativeSigner = await wallet.getSigner();
+
+    if (!nativeSigner) {
+      throw new Error('Failed to access signer for HyperCore permit');
+    }
+
+    return nativeSigner.signTypedData(domain, types, value);
+  }
+
+  if (typeof (signer as any).signTypedData === 'function') {
```
priyajeet: 175-186 should not be needed

src/config/mainnet/chains.ts
```diff
@@ -217,6 +217,14 @@ export const MAINNET_CHAINS: ChainsConfig = {
     icon: 'HyperEVM',
     symbol: 'HYPE',
   },
+  HyperCore: {
+    displayName: 'HyperCore',
+    sdkName: 'HyperCore',
+    explorerUrl: 'https://arbiscan.io/', // Uses Arb USDC
```
priyajeet: Should these point back to what they were before, the hyperliquid explorer

src/hooks/useGetTokenBalances.ts
```diff
@@ -156,6 +156,12 @@ const useGetTokenBalances = ({
       tokens: Token[],
     ): Promise<Balances> => {
       const chainConfig = config.chains[chain];
+
+      // HyperCore is a destination-only chain, skip balance fetching
```
priyajeet, line 160: I'll add a backlog task to get balances from hyperliquid perps account

src/views/v3/Bridge/AssetPicker/TokenItem.tsx
```diff
@@ -21,6 +20,7 @@ import {
 import ChainIcon from 'icons/ChainIcons';
 import Color from 'color';
 import TokenBalance from 'components/TokenBalance';
+import { isHyperCoreChain } from 'utils/hypercore';
```
priyajeet: Is the changes in this file needed if we are returning {} from useTokenGetBalances() ?

src/routes/mayan/MayanRouteBase.ts
```diff
@@ -156,8 +162,15 @@ export class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
     _fromChain: ChainContext<N>,
     toChain: ChainContext<N>,
   ): Promise<TokenId[]> {
-    // TODO update this function to only allow HyperCore/USDC
-    return getAllTokenIdsForChain(toChain.chain);
+    const tokens = getAllTokenIdsForChain(toChain.chain);
```
priyajeet: Probably move this more expensive line below the  if () block

src/utils/hypercore.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  if (!requiresPermit) return undefined;
+
+  const quoteDetails = quote.details;
+
+  if (!quoteDetails) {
+    throw new Error('Missing HyperCore quote details required for permit');
+  }
+
+  const arbitrumRpc = (await request.toChain.getRpc()) ?? null;
```
priyajeet: request.toChain is HyperCore and not Arb...so its getting HyperCore rpc that doesn't exist. Can you check what this is ending up with?

in connect I think rpc's come from src/sdklegacy/config

kyleleighton: We are hitting the Arb rpc. [This PR adds](https://github.com/wormhole-foundation/wormhole-connect/blob/b5fd31f63a58f2c4112ff0a6a00e14c04744f8da/src/sdklegacy/config/MAINNET.ts#L34) the Hypercore rpc as Arb

Screenshot 2025-10-01 at 1 40 47 PM

priyajeet: For portal it will come from https://github.com/wormholelabs-xyz/rpc-proxy-worker/blob/main/src/index.ts
Will have to update there also.

Maybe better would be to force request the ARB rpc instead getRPC(Arbitrum) 

Something like https://github.com/wormhole-foundation/wormhole-connect/blob/e02307fadf4e4ed2105b5a16cefe6166c6487db4/src/hooks/useGetTokenBalances.ts#L187

kyleleighton: Nice. That seems to work: https://explorer.mayan.finance/tx/0x9f5aea58e3e61b7961337a2b443db6f4e78b84515e6feb5360ecd03cca171460

priyajeet (approved):

PR #3835, opened 2025-10-01
chore: update Mui & fix issues from it

bradleyplaydon:
## Updates
- Updated Mui package
- Fixed incorrect imports in new version of Mui

### DIFF
--- a/src/WormholeConnect.tsx
+++ b/src/WormholeConnect.tsx
@@ -1,7 +1,4 @@
 import * as React from 'react';
-// fixes "styled_default is not a function" error
-// https://github.com/vitejs/vite/issues/12423#issuecomment-2080351394
-import '@mui/material/styles/styled';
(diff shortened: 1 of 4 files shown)


src/WormholeConnect.tsx
```diff
@@ -1,7 +1,4 @@
 import * as React from 'react';
-// fixes "styled_default is not a function" error
-// https://github.com/vitejs/vite/issues/12423#issuecomment-2080351394
```
priyajeet, line 3: The bug isn't fixed, but assume its a mui v6 issue?

priyajeet (approved):

src/WormholeConnect.tsx
bradleyplaydon: I didn't get any build errors from removing it, so I assume so? It also no longer exists anymore?

PR #3839, opened 2025-10-01
feat: Asset picker design changes

emreboga:
This will apply the new designs for the asset picker and bridge header.
Fixes PROD-803

Screenshot 2025-10-02 at 7 17 07 PM
Screenshot 2025-10-02 at 7 16 57 PM
Screenshot 2025-10-02 at 7 15 06 PM
Screenshot 2025-10-02 at 7 15 15 PM


src/views/v3/Bridge/index.tsx
```diff
@@ -519,132 +490,102 @@ function Bridge(props: BridgeProps) {
 
   const bridgeContent = (
```
emreboga, line 491: I was thinking to move bridgeContent to a separate component (BridgeContent.tsx) but it's too tangled here I decided to do this in a separate work.

kev1n-peters (approved):

PR #3843, opened 2025-10-03
fix: transaction history inconsistencies

nascob:
### Updates 

- do not rely on decimals coming from whscan 
- filter out Portal Bridge attestation txs from history results

### DIFF
--- a/src/utils/whScanUtils.ts
+++ b/src/utils/whScanUtils.ts
@@ -1,12 +1,12 @@
 import type { WormholeScanTransaction } from 'hooks/useTransactionHistoryWHScan';
 
-export const isPortalBridgeAttestationTx = (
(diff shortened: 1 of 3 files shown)


src/utils/whScanUtils.ts
```diff
@@ -0,0 +1,12 @@
+import type { WormholeScanTransaction } from 'hooks/useTransactionHistoryWHScan';
+
+export const isPortalBridgeAttestationTx = (
```
priyajeet: Can you switch to function() syntax for this. I think hence forth for all top level functions we should use that style.

nascob: Done.

priyajeet (approved):

PR #3845, opened 2025-10-06
fix: Styling for chain icon and close and back icons is asset picker

emreboga:
Verified on Portal.

Screenshot 2025-10-06 at 6 43 34 PM


src/views/v3/Bridge/AssetPicker/ChainList.tsx
```diff
@@ -104,7 +104,6 @@ function ChainList(props: Props) {
         whiteSpace: 'nowrap',
       },
       chainIcon: {
-        border: '1px solid transparent',
```
emreboga, line 107: This was causing a shift and we don't need it to set the border radius.

kev1n-peters (approved):

PR #3849, opened 2025-10-08
feat: Enable Solana routes to HyperCore

nascob:
### Changes

This PR enables  Solana -> HyperCore routes

List of changes: 
- Update Mayan SDK to 11.3.0
- Remove EVM restrictions 
- Detect and switch wallet destination wallet 
- Send USDC Permit signatures into Solana routes

### DIFF
--- a/src/routes/mayan/MayanRouteBase.ts
+++ b/src/routes/mayan/MayanRouteBase.ts
@@ -649,11 +649,10 @@ export class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
       );
 
       if (request.fromChain.chain === 'Solana') {
-        // Uncomment and use this when we want to support HyperCore USDC deposits on SOL
-        // const solanaOptions = {
-        //   allowSwapperOffCurve: true,
-        //   usdcPermitSignature,
-        // };
+        const solanaOptions = {
+          allowSwapperOffCurve: true,
+          ...(usdcPermitSignature ? { usdcPermitSignature } : {}),
+        };
 
         const { instructions, signers, lookupTables } =
           await (this.isTestnetRequest(request)
@@ -663,7 +662,7 @@ export class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
                 destinationAddress,
                 null,
                 rpc,
-                { allowSwapperOffCurve: true },
+                solanaOptions,
               )
             : createSwapFromSolanaInstructions(
                 quote.details!,
@@ -671,7 +670,7 @@ export class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
                 destinationAddress,
                 null,
                 rpc,
-                { allowSwapperOffCurve: true },
+                solanaOptions,
               ));
 
         const payerKey = new PublicKey(originAddress);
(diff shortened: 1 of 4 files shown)


src/utils/hypercore.ts
```diff
@@ -63,17 +63,6 @@ export function validateHyperCoreTransfer<N extends Network>(
     };
   }
 
-  // Temporary restriction to EVM for now; SOL and SUI are coming soon
-  if (!isEvmChain(fromChain.chain)) {
```
kyleleighton: We still need to gate this, but now the source can be either EVM or Solana.

nascob: or SUI. It's good to be cautious - i'll add it back.

src/utils/hypercore.ts
```diff
@@ -134,7 +123,11 @@ export async function maybeGetHyperCorePermitSignature<N extends Network>(
     throw new Error('No wallet provider available for HyperCore permit');
   }
 
-  const wallet = walletProvider.getWallet(ARBITRUM, TransferWallet.SENDING) as
+  const walletToSwitch = isEvmChain(request.fromChain.chain)
+    ? TransferWallet.SENDING
```
priyajeet, line 143: What happens if no connected wallets? If you go to your wallet app and disconnect and start from clean slate? And also what happens if wallet is solana only, like turning off evm

src/routes/mayan/MayanRouteBase.ts
```diff
@@ -732,13 +731,15 @@ export class MayanRouteBase<N extends Network> extends routes.AutomaticRoute<
             quote.params.amount,
           );
 
-        const options: ComposableSuiMoveCallsOptions | undefined =
-          builtTransaction
+        const options: ComposableSuiMoveCallsOptions = {
+          ...(builtTransaction
```
priyajeet: Lets refactor this section (building of options) to be more readable. Currently ... ? : {}s are just making it hard to decipher whats going on. Can move it to a utility file under sui subfolder, like there is one for evm

src/utils/hypercore.ts
nascob: All good questions that I've been testing.

Q: What happens if I not not connected wallets?
A: The validation logic that's already in place asks you to connect a wallet. If it's not the right type of wallet - it rejects it (nothing seems to happen).

Q: If you go to your wallet app and disconnect and start from clean slate? 
A:  I don't understand this - nothing has changed on the wallet/network select logic - it's just the same as disconnecting a Solana/EVM wallet and starting over?

Q: And also what happens if wallet is solana only, like turning off evm
A: It does what it does now with the "Connect source/destination wallet" message in the CTA Button

src/routes/mayan/MayanRouteBase.ts
nascob: I agree - the syntax is crap - but all this is doing is merging defined parts. 
Idk about a util function but maybe this would make it more readable? 

```typescript 
const txPart =
  builtTransaction && {
    builtTransaction,
    inputCoin: { result: remainingAmountCoin },
  };

const permitPart = usdcPermitSignature && { usdcPermitSignature };

const options: ComposableSuiMoveCallsOptions = {
  ...txPart,
  ...permitPart,
};
```

src/utils/hypercore.ts
nascob: Done

src/routes/mayan/MayanRouteBase.ts
nascob: Refactored

kyleleighton:
LGTM. Nice job.

kev1n-peters (dismissed):

kev1n-peters (approved):

PR #3850, opened 2025-10-08
feat: Fee offsetting for Mayan and Executor routes (CCTP, NTT, Token Bridge)

emreboga:
This PR implements referrer fee offsetting to ensure users receive exactly the amount they request, after referrer fees are deducted. Fee offsetting calculates the exact additional amount needed so users receive their desired amount after fees are deducted:
- User enters: 100 USDC
- Connect calculates: +1.01 USDC offset
- Total sent: 101.01 USDC
- Fee deducted: 1.01 USDC
- User receives: 100 USDC
  
The feature is currently behind an experimental flag:
```
{
  network: 'mainnet',
  ui: {
    experimental: { feeOffsetting: true },
  },
  ...
}
```

You can see the offset amount right beneath the amount input.

Screenshot 2025-10-09 at 12 07 42 AM


src/utils/fees.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  // Other routes use deci-basis points (1 dbps = 0.001% = 1/100000)
+  return applyOffsetFormula(amount, feeDbps, 100000n);
+}
+
+/**
+ * Format fee display text
+ */
+export function formatFeeDisplay(
```
kev1n-peters: Doesn't look like this is called from anywhere?

src/utils/fees.ts
```diff
@@ -0,0 +1,178 @@
...
+  } else {
+    return undefined;
+  }
+
+  // Handle Mayan routes differently - they use getReferrerBps function
+  if (sdkRoute.rc.meta.name.startsWith('MayanSwap')) {
+    const mayanRoute = sdkRoute.rc as any;
+    if (mayanRoute.getReferrerBps && sourceToken && destToken) {
+      // Create a RouteTransferRequest object for Mayan's getReferrerBps
+      const request = {
```
kev1n-peters, line 106: Is there a reason to not use `routes.RouteTransferRequest.create()`?

src/utils/fees.ts
```diff
@@ -0,0 +1,178 @@
...
+  // Handle SDK routes with static config
+  const routeConfig = (sdkRoute.rc as any).config;
+
+  if (!routeConfig) {
+    return undefined;
+  }
+
+  let feeDbps = 0n;
+
+  // CCTP Executor routes use referrerFeeDbps directly
```
kev1n-peters: I think we should explicitly check the `sdkRoute.rc.meta.name` before pulling out the fee to be on the safe side in case new routes are added

src/utils/fees.ts
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Calculate the fee offset amount needed to achieve the desired output after fee deduction
+ * @param amount - The desired output amount (what user wants to receive)
+ * @param routeName - The name of the route
+ * @param sourceToken - The source token (required for token-specific fees)
+ * @param destChain - Optional destination chain (for Mayan routes)
+ * @param destToken - Optional destination token (for Mayan routes)
+ * @returns The additional amount to add so that after fee deduction, user receives the desired amount
+ */
+export function calculateFeeOffset(
```
kev1n-peters, line 75: I think we may want to refactor this in the future to abstract the fee BPS retrieval from the SDK route itself, rather than extracting it here. That would be a larger project to create a consistent interface across all routes.

src/utils/fees.ts
```diff
@@ -0,0 +1,178 @@
...
+      } else {
+        feeDbps = routeConfig.referrerFee.feeDbps;
+      }
+    } else {
+      feeDbps = routeConfig.referrerFee.feeDbps;
+    }
+  }
+  // Token Bridge Executor route uses referrerFee.referrerFeeDbps
+  else if (routeConfig.referrerFee?.referrerFeeDbps !== undefined) {
+    feeDbps = routeConfig.referrerFee.referrerFeeDbps;
```
kev1n-peters: The TB route also has per token overrides in `routeCOnfig.referrerFee.tokenFeeOverrides`.

src/utils/fees.ts
```diff
@@ -0,0 +1,178 @@
...
+  value: bigint,
+  decimals: number,
+  maxDecimals: number = 6,
+): bigint {
+  if (decimals <= maxDecimals) {
+    // No rounding needed if token has 6 or fewer decimals
+    return value;
+  }
+
+  const factor = BigInt(10 ** (decimals - maxDecimals));
```
kev1n-peters: `const factor = 10n ** BigInt(decimals - maxDecimals);` to avoid losing precision

src/utils/fees.ts
```diff
@@ -0,0 +1,178 @@
...
+ * @param decimals - The number of decimal places for the token
+ * @param maxDecimals - Maximum decimal places to keep (default 6)
+ */
+export function roundDownToDecimals(
+  value: bigint,
+  decimals: number,
+  maxDecimals: number = 6,
+): bigint {
+  if (decimals <= maxDecimals) {
+    // No rounding needed if token has 6 or fewer decimals
```
kev1n-peters: `// No rounding needed if the token has fewer decimals than the max allowed`

src/utils/fees.ts
emreboga: Good point!

src/utils/fees.ts
emreboga: Yup, I agree on moving this into SDK 👍

src/utils/fees.ts
emreboga: I wasn't aware of that one, let me try.

src/utils/fees.ts
emreboga: You mean a non-executor route with a similar `referrerFeeDbps` prop? Yeha, I can add a check for executor route.

src/utils/fees.ts
emreboga: Oh, I might have dropped it along the way. I'll remove it.

src/utils/fees.ts
kev1n-peters: Yeah, I think we should explicitly check the routes before pulling out the fees:
`TokenBridgeExecutorRoute`
`NttExecutorRoute`
`CCTPv2StandardExecutorRoute`
`CCTPv2FastExecutorRoute`

src/utils/fees.ts
kev1n-peters: If we use that one, we shuld pass in the toke decimals parameters to save on RPC calls (check out the create() implementation in the SDK)

emreboga: The issue is if we pass a Wormhole instance or create a RouteTransferRequest in calculateFeeOffset, there are issues such as creating a RouteTransferRequest requires async calls to get Wormhole context and chain contexts. The function needs to remain synchronous for use in the React component (FeeOffset.tsx). 
I'd like to keep things less complicated in this case. I can add more tests for guard rails.

src/utils/fees.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3850/commits/d93e2ae0ece385285fdd270b17d5af38f4bc738a

src/utils/fees.ts
emreboga: I'll add decimals to the request object.

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3850/commits/d93e2ae0ece385285fdd270b17d5af38f4bc738a

src/utils/fees.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3850/commits/d93e2ae0ece385285fdd270b17d5af38f4bc738a

src/utils/fees.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3850/commits/d93e2ae0ece385285fdd270b17d5af38f4bc738a

src/utils/fees.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3850/commits/d93e2ae0ece385285fdd270b17d5af38f4bc738a

src/utils/fees.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3850/commits/d93e2ae0ece385285fdd270b17d5af38f4bc738a

src/utils/fees.ts
```diff
@@ -0,0 +1,181 @@
...
+        },
+        destination: {
+          id: destToken.tokenId || {
+            chain: destChain || destToken.chain,
+            address: destToken.address,
+          },
+          symbol: destToken.symbol,
+          decimals: destToken.decimals,
+        },
+      } as any;
```
kev1n-peters: I don't love typecasing to `any` here, would it be possible to extend `RouteTransferRequest` with a new class and pass that instead here? `getReferrerBps()` may be modified in the future to call methods on the passed `RouteTransferRequest`

emreboga: I understand your concern that the link from this utility to the actual Mayan route implementation might be lost if we change the route and don't reflect that here. However that will not be much different than using `routes.RouteTransferRequest.create()` because it'll need a Wormhole instance and chain contexts for creation.
Let me think about this more.

emreboga: Do you think this would catch anything at build-time if the route interface changes:
```
    const request = {
        source: {
          id: sourceToken.tokenId || {
            chain: sourceToken.chain,
            address: sourceToken.address,
          },
          symbol: sourceToken.symbol,
          decimals: sourceToken.decimals,
        },
        destination: {
          id: destToken.tokenId || {
            chain: destChain || destToken.chain,
            address: destToken.address,
          },
          symbol: destToken.symbol,
          decimals: destToken.decimals,
        },
      } as routes.RouteTransferRequest<Network>;
```

kev1n-peters (dismissed):

kev1n-peters (dismissed):

kev1n-peters (approved):

PR #3851, opened 2025-10-09
fix: use theme font family in chain list

kev1n-peters:
This was causing the chain list font on monadbridge.com to look too small


emreboga:
This fix is included in https://github.com/wormhole-foundation/wormhole-connect/pull/3854

PR #3855, opened 2025-10-09
fix: Missing prop to calculate fee-offset for Mayan routes

emreboga:
Destination token was missing in FeeOffset component when calculating the offset for Mayan routes.

### DIFF
--- a/src/utils/fees.ts
+++ b/src/utils/fees.ts
@@ -75,7 +75,7 @@ export function calculateFeeOffset(
   route: SDKv2Route | string | undefined,
   amount: sdkAmount.Amount | undefined,
   sourceToken: Token | undefined,
-  destToken: Token,
+  destToken?: Token,
 ): sdkAmount.Amount | undefined {
   if (!sourceToken || !amount || !route || sdkAmount.units(amount) === 0n) {
     return undefined;
--- a/src/utils/fees.ts
+++ b/src/utils/fees.ts
@@ -68,7 +68,6 @@ export function applyOffsetFormula(
  * @param amount - The desired output amount (what user wants to receive)
  * @param routeName - The name of the route
  * @param sourceToken - The source token (required for token-specific fees)
- * @param destChain - Optional destination chain (for Mayan routes)
  * @param destToken - Optional destination token (for Mayan routes)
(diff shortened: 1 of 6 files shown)


src/utils/fees.ts
```diff
@@ -68,15 +68,13 @@ export function applyOffsetFormula(
...
  * @param sourceToken - The source token (required for token-specific fees)
- * @param destChain - Optional destination chain (for Mayan routes)
  * @param destToken - Optional destination token (for Mayan routes)
  * @returns The additional amount to add so that after fee deduction, user receives the desired amount
  */
 export function calculateFeeOffset(
   route: SDKv2Route | string | undefined,
   amount: sdkAmount.Amount | undefined,
   sourceToken: Token | undefined,
-  destChain?: string,
```
emreboga, line 79: No need for chain as we can get it from the token.

src/utils/fees.ts
```diff
@@ -114,7 +112,7 @@ export function calculateFeeOffset(
         },
         destination: {
           id: destToken.tokenId || {
-            chain: destChain || destToken.chain,
+            chain: destToken.chain,
```
bradleyplaydon, line 115: Nice

src/views/v3/Bridge/AssetPicker/FeeOffset.tsx
```diff
@@ -5,27 +5,26 @@ import InfoOutlineIcon from '@mui/icons-material/InfoOutline';
...
-  token?: Token;
-};
+import { useGetTokens } from 'hooks/useGetTokens';
 
 /**
  * Displays the fee offset amount that will be added to ensure users receive
  * exactly what they requested after protocol fees are deducted
  */
-function FeeOffset(props: Props) {
+function FeeOffset() {
```
bradleyplaydon, line 15: Just out of interest, does the parent no longer have the props?

emreboga: Parent was passing down the source token it got from Redux. We need to get the destination token from redux and while at it we can get the source as well and don't need them from parent.

bradleyplaydon (approved):

PR #3856, opened 2025-10-10
fix: Addressing follow up items from Asset Picker design changes

emreboga:
Fixes PROD-847
Note: The last item in the above ticket is addressed by https://github.com/wormhole-foundation/wormhole-connect/pull/3854


bradleyplaydon (approved):
LGTM :)

PR #3887, opened 2025-10-27
fix: Number formatting for different locales

emreboga:
Fixes BRI-49


src/views/v3/Bridge/AmountInput.tsx
```diff
@@ -47,12 +47,11 @@ const DebouncedTextField = memo(
 
     const onInnerChange: ChangeEventHandler<HTMLInputElement> = useCallback(
       (e: React.ChangeEvent<HTMLInputElement>) => {
-        const value = removeCommas(e.target.value);
-
-        if (!isValidDecimalInput(value)) {
+        if (!isValidDecimalInput(e.target.value)) {
```
emreboga, line 50: Validity should be checked with the locale formatted value, before removing commas.

priyajeet (approved):

PR #3889, opened 2025-10-28
enable gas drop off by default when recipient gas token

kev1n-peters:
Addresses BRI-45
Addresses BRI-48

48 is needed for 45 to work (user needs ability to toggle gas which was not exposed)

### DIFF
--- a/src/hooks/useAutoEnableGasDropoff.ts
+++ b/src/hooks/useAutoEnableGasDropoff.ts
@@ -14,12 +14,13 @@ export interface UseAutoEnableGasDropOffParams {
   destinationBalances: Balances;
   hasUserManuallyChangedGas: boolean;
   currentToNativeToken: number;
+  isFetchingBalances: boolean;
 }
 
 /**
  * Automatically enables gas drop-off when the destination
- * wallet has zero native token balance, unless the user has manually changed
- * the gas setting.
+ * wallet has zero native token balance for executor routes,
+ * unless the user has manually changed the gas setting.
  */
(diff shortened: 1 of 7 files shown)


src/views/v3/Bridge/index.tsx
```diff
@@ -64,6 +65,8 @@ function Bridge(props: BridgeProps) {
   const dispatch = useDispatch();
 
   const [showHistory, setShowHistory] = useState(props.showHistory ?? false);
+  const [hasUserManuallyChangedGas, setHasUserManuallyChangedGas] =
```
priyajeet, line 68: nit: useBoolean is available

src/hooks/useAutoEnableGasDropoff.ts
```diff
@@ -0,0 +1,80 @@
...
+  receivingWalletAddress,
+  destinationBalances,
+  hasUserManuallyChangedGas,
+  currentToNativeToken,
+  isFetchingBalances,
+}: UseAutoEnableGasDropOffParams): void {
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    console.log(route);
```
priyajeet: Was this intended to be left? Should it be .info?

priyajeet (approved):

src/views/v3/Bridge/index.tsx
kev1n-peters: did not know this, will look into using it in the future (not used in connect currently)

PR #3895, opened 2025-10-30
fix: Number formatting for intl locales

emreboga:
Number intl formatting between user display and stored values is a bit complicated. I've first renamed the util functions to better match their functionalities:
```
formatWithCommas() → formatNumberIntl()
removeCommas() → removeFormatting()
isValidDecimalInput() → isValidDecimalNumber()
```
- `formatNumberIntl` now formats a non-locale number string to locale number string.
- `isValidDecimalNumber` now makes sure to remove any formatting before validating. It can also work with both locale and non-locale numbers strings.
- We also make sure to use the validated value and re-format it before displaying.

### DIFF
--- a/src/utils/formatNumber.test.ts
+++ b/src/utils/formatNumber.test.ts
@@ -1,8 +1,8 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import {
-  formatWithCommas,
-  removeCommas,
-  isValidDecimalInput,
+  formatNumberIntl,
+  removeFormatting,
+  isValidDecimalNumber,
   formatMinAmount,
 } from './formatNumber';
 import { amount as sdkAmount } from '@wormhole-foundation/sdk';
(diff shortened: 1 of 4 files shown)


src/utils/formatNumber.test.ts
```diff
@@ -39,146 +39,144 @@ describe('formatNumber utilities', () => {
...
 
   describe.each(localeConfigs)(
-    'isValidDecimalInput - $locale',
+    'isValidDecimalNumber - $locale',
     ({ locale, decimal }) => {
       beforeEach(() => {
         mockNavigatorLanguage(locale);
       });
 
       it('should accept valid decimal numbers', () => {
```
priyajeet, line 129: Unless I missed it, we should add a test that adds groups to the number (in addition to decimals) being checked. This was the recent issue.

pt-PT "10 000,99" 
en-US "10,000.99"

10group000decimal99

src/utils/formatNumber.test.ts
```diff
@@ -1,8 +1,8 @@
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
 import {
-  formatWithCommas,
-  removeCommas,
-  isValidDecimalInput,
+  formatNumberIntl,
+  removeFormatting,
+  isValidDecimalNumber,
```
priyajeet: isValidFormattedDecimalNumber() might be more descriptive since the value passed in should/can be formatted

priyajeet: isValidFormattedNumber() might also work

src/utils/formatNumber.test.ts
emreboga: Let me add that and pt-PT locale as well.

emreboga: Actually, I'll replace Japanese with Polish. It has the same separators as US anyway.

emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3895/commits/701ee0fded2a85d666575793b6ae645dcf3829f4

src/utils/formatNumber.test.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3895/commits/701ee0fded2a85d666575793b6ae645dcf3829f4

priyajeet (approved):

PR #3899, opened 2025-11-03
fix: Number precision for large amounts

emreboga:
`parseInt` can't safely handle integers larger than Number.MAX_SAFE_INTEGER (9,007,199,254,740,991).
Using BigInt fixes it.

Fixes BRI-88

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,6 +162,7 @@
     "@typescript-eslint/eslint-plugin": "^8.38.0",
     "@typescript-eslint/parser": "^8.38.0",
     "@vitejs/plugin-react-swc": "^3.5.0",
+    "@vitest/coverage-v8": "^3.2.4",
     "@vitest/ui": "^3.2.4",
     "commitizen": "^4.3.1",
     "cz-conventional-changelog": "^3.3.0",
--- a/src/utils/formatNumber.ts
+++ b/src/utils/formatNumber.ts
@@ -39,15 +39,15 @@ export const formatNumberIntl = (value: string): string => {
   }
 
   const [integerPart, decimalPart] = value.split('.');
-  const intNum = parseInt(integerPart, 10) || 0;
+
+  // Use BigInt to handle large numbers without precision loss
+  const intNum = integerPart ? BigInt(integerPart) : 0n;
 
   const locale = getUserLocale();
   const { decimal } = getSeparators(locale);
 
-  // Format the integer part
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -162,6 +162,7 @@
     "@typescript-eslint/eslint-plugin": "^8.38.0",
     "@typescript-eslint/parser": "^8.38.0",
     "@vitejs/plugin-react-swc": "^3.5.0",
+    "@vitest/coverage-v8": "^3.2.4",
```
emreboga, line 165: Something we need to get the test coverage numbers when we run tests.

src/utils/formatNumber.ts
```diff
@@ -39,23 +39,24 @@ export const formatNumberIntl = (value: string): string => {
...
     useGrouping: true,
     minimumFractionDigits: 0,
     maximumFractionDigits: 0,
   }).format(intNum);
 
   // Append the locale decimal separator + any digits (or preserve trailing decimal separator)
-  if (decimalPart !== undefined || value.endsWith(decimal)) {
+  // Input is non-localized format, so check for '.' not locale decimal
+  if (decimalPart !== undefined || value.endsWith('.')) {
     return formattedInt + decimal + (decimalPart ?? '');
```
priyajeet, line 60: decimal ---> . ?

src/utils/formatNumber.ts
```diff
@@ -39,23 +39,24 @@ export const formatNumberIntl = (value: string): string => {
   }
 
   const [integerPart, decimalPart] = value.split('.');
```
priyajeet, line 41: What about bigInt of decimal part?

emreboga: That's less likely as you need to go 14 decimals to reach that, but I will add 👍

src/utils/formatNumber.ts
emreboga: Not sure if I understood. `decimal` is the localized separator. Value input is not localized, we check for ".", but the return will be localized so we use `decimal`.

src/utils/formatNumber.ts
emreboga: Oh wait, we don't need that. The reason we are doing BigInt parsing for integer part is because we need to format with group separators. There is no grouping in decimal part and we directly append the value (no parsing).

priyajeet (approved):

PR #3900, opened 2025-11-03
feat: Change Fogo token display names and show warning

emreboga:
Fixes BRI-17

Screenshot 2025-11-04 at 12 30 13 AM

Screenshot 2025-11-04 at 12 20 57 AM

### DIFF
--- a/src/config/mainnet/tokens.ts
+++ b/src/config/mainnet/tokens.ts
@@ -815,12 +815,14 @@ export const MAINNET_TOKENS: TokenConfig[] = [
   },
   {
     symbol: 'FOGO',
+    name: 'FOGO FUEL',
     icon: TokenIcon.FOGO,
     decimals: 9,
     tokenId: { chain: 'Fogo', address: 'native' },
   },
   {
     symbol: 'WFOGO',
+    name: 'FOGO',
     icon: TokenIcon.FOGO,
     decimals: 9,
     tokenId: {
--- a/src/views/v3/Bridge/index.tsx
+++ b/src/views/v3/Bridge/index.tsx
@@ -519,6 +519,16 @@ function Bridge(props: BridgeProps) {
     onConfirm,
   ]);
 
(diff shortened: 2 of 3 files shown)


src/config/mainnet/tokens.ts
```diff
@@ -815,12 +815,14 @@ export const MAINNET_TOKENS: TokenConfig[] = [
...
   {
     symbol: 'FOGO',
+    name: 'FOGO FUEL',
     icon: TokenIcon.FOGO,
     decimals: 9,
     tokenId: { chain: 'Fogo', address: 'native' },
   },
   {
     symbol: 'WFOGO',
+    name: 'FOGO',
```
kev1n-peters, line 825: Do we need to change the symbols or are they good as is?

src/views/v3/Bridge/index.tsx
```diff
@@ -609,6 +609,22 @@ function Bridge(props: BridgeProps) {
           {transactionError}
           {transactionInfo}
           <AmountValidationError validation={amountValidation} />
+          {destToken &&
+            destToken.symbol === 'FOGO' &&
```
kev1n-peters: Can we check the chain instead of symbol which isn’t guaranteed to be unique?

src/config/mainnet/tokens.ts
emreboga: Nothing is mentioned about symbols so far.

src/views/v3/Bridge/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3900/commits/3720970ecd0d6cc7e6110b83e1fcbd7e95dca232

evan-gray:
Closing as it seems the traditional names may stick.

emreboga:
Fogo reversed course on this after integrator pushback and will stick with the traditional approach.

PR #3901, opened 2025-11-03
feat: allow filtering of tokens based on source or destination list

priyajeet:


src/config/types.ts
```diff
@@ -96,6 +96,7 @@ export type IsRouteSupportedHandler = (
 export type IsTokenSupportedHandler = (
   token: Token,
   sourceToken?: Token, // The selected source token, if applicable
+  tokenType?: 'source' | 'destination',
```
priyajeet: Alternative could be a breaking change and move both sourceToken and tokenType nested inside another param, say context (or better name). Upside is that it makes the api more extensible.

```ts
export type IsTokenSupportedHandler = (
  token: Token,
  context: {
    sourceToken?: Token, // The selected source token, if     applicable
    tokenType?: 'source' | 'destination',
  },
) => boolean;
```

priyajeet: `tokenType` should probably be `tokenListType` in any case

src/utils/tokenListUtils.ts
```diff
@@ -177,12 +177,17 @@ export const applyTokenWhitelist = (
 export const applyCustomTokenSupport = (
   tokens: Token[],
   sourceToken?: Token,
+  isSourceToken?: boolean,
```
priyajeet: isSourceToken ==> isSourceList

kev1n-peters (dismissed):

kev1n-peters (approved):

PR #3913, opened 2025-11-10
chore: add Ink Sepolia

aapclark:


dvgui (dismissed):

kev1n-peters (dismissed):
Need to include a default Ink testnet RPC in src/sdklegacy/config/TESTNET.ts

kev1n-peters:
Also will need to wait until this is published so the executor addresses are available in the NTT SDK for Ink Testnet: https://github.com/wormhole-foundation/native-token-transfers/pull/736

aapclark:
> Need to include a default Ink testnet RPC in src/sdklegacy/config/TESTNET.ts

Fixed

kev1n-peters:
We should update the package.json to use version 4.0.4 of the NTT SDKs (which includes Ink testnet support -- update each package that's on 4.0.3 to 4.0.4 and run npm i afterwards)

dvgui (approved):

PR #3914, opened 2025-11-10
fix: handle coingecko 404 responses gracefully

kev1n-peters:
fixes seeing errors in the console related to coins not being found on coingecko, e.g.:
```
hook.js:608 Coingecko error Error: string message has invalid length for format hex
    at 6900-4003b15ccf3279d5.js:7:431211
    at h.stringToUint8Array (6900-4003b15ccf3279d5.js:7:431502)
    at new h (6900-4003b15ccf3279d5.js:7:430597)
    at c (6900-4003b15ccf3279d5.js:467:47180)
    at b.parseAddress (6900-4003b15ccf3279d5.js:1479:25102)
    at b.chainAddress (6900-4003b15ccf3279d5.js:1479:25203)
    at b.tokenId (6900-4003b15ccf3279d5.js:1479:25277)
    at 6900-4003b15ccf3279d5.js:483:442697
    at Array.map (<anonymous>)
    at 6900-4003b15ccf3279d5.js:483:442675
    ```

### DIFF
--- a/src/utils/coingecko.ts
+++ b/src/utils/coingecko.ts
@@ -82,7 +82,15 @@ const coingeckoRequest = async (
     signal: params?.abort?.signal,
     headers,
   })
-    .then((resp) => resp.json())
+    .then((resp) => {
+      if (!resp.ok) {
+        console.error(
+          `Coingecko request failed with status ${resp.status}: ${path}`,
+        );
+        return null;
+      }
+      return resp.json();
+    })
     .catch((err) => {
       console.error('Error fetching from Coingecko', err);
       return null;
@@ -136,7 +144,13 @@ export const fetchTokenPrices = async (
         params,
       )
         .then((data) => {
-          if (data['error'] !== undefined || data['error_code'] !== undefined) {
+          if (data === null) {
+            // Request failed (e.g., 404, network error)
+            resolve([]);
+          } else if (
+            data['error'] !== undefined ||
+            data['error_code'] !== undefined
+          ) {
             reject(data['error']);
           } else {
             resolve(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/utils/coingecko.ts [hunk 1]
```diff
@@ -82,7 +82,15 @@ const coingeckoRequest = async (
     signal: params?.abort?.signal,
     headers,
   })
-    .then((resp) => resp.json())
+    .then((resp) => {
```
priyajeet, line 85: This file is quite nested and is hard to read, but has a deterministic output. I think it would greatly benefit from claude auto-refactoring. Before refactoring we can get claude to write a whole bunch of tests, then get claude to refactor and tests should still work. Obvious fixes can be switching to async/await, moving coingecko call out for fetching native vs non native tokens, moving request building and response parsing/mapping out as functions, reduce loops after loops (for eg last few lines maps, filters, flats and still loops), etc.

src/utils/coingecko.ts
```diff
@@ -82,7 +82,15 @@ const coingeckoRequest = async (
...
+      if (!resp.ok) {
+        console.error(
+          `Coingecko request failed with status ${resp.status}: ${path}`,
+        );
+        return null;
```
priyajeet, line 90: Why not throw since we are already logging below and returning null

src/utils/coingecko.ts
```diff
@@ -136,7 +144,13 @@ export const fetchTokenPrices = async (
         params,
       )
         .then((data) => {
-          if (data['error'] !== undefined || data['error_code'] !== undefined) {
+          if (data === null) {
+            // Request failed (e.g., 404, network error)
+            resolve([]);
```
priyajeet, line 149: Why resolve and not reject like other errors below? Or inversely, why reject below and not return []. In the end all its checking if its fullfilled or not.

src/utils/coingecko.ts
```diff
@@ -186,24 +200,29 @@ export const fetchTokenPrices = async (
...
-          )
+          .then((data) => {
+            if (data === null) {
+              // Request failed (e.g., 404, network error)
+              resolve([]);
+            } else {
+              resolve(
+                nativeTokens
+                  .map((tokenId) => {
+                    const cgid = NATIVE_TOKEN_IDS[tokenId.chain];
```
bradleyplaydon: camelCase?

PR #3916, opened 2025-11-11
feat: Monad mainnet support

kev1n-peters:
The Monad Bridge route supports ETH and MON transfers between Ethereum and Monad via a frontend allowlist.

### DIFF
--- a/src/routes/monad/consts.ts
+++ b/src/routes/monad/consts.ts
@@ -13,8 +13,34 @@ export const TOKEN_DENY_LIST: Partial<Record<Network, TokenId[]>> = {
   ],
 };
 
+// Mainnet-only allowlist - only tokens in this list can be bridged on mainnet
+// Testnet does not use an allowlist, only the deny list above
+export const TOKEN_ALLOW_LIST: Partial<Record<Network, TokenId[]>> = {
+  Mainnet: [
+    // Monad tokens
+    Wormhole.tokenId('Monad', 'native'), // MON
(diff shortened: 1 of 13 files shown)


src/routes/monad/consts.ts
```diff
@@ -13,8 +13,34 @@ export const TOKEN_DENY_LIST: Partial<Record<Network, TokenId[]>> = {
   ],
 };
 
+// Mainnet-only allowlist - only tokens in this list can be bridged on mainnet
+// Testnet does not use an allowlist, only the deny list above
+export const TOKEN_ALLOW_LIST: Partial<Record<Network, TokenId[]>> = {
```
priyajeet, line 18: Instead of a hard code here, can passing in a [isTokenSupportedHandler()](https://github.com/wormhole-foundation/wormhole-connect/blob/b5116a8b16c9736e2a08a11dca4f3ee6b2a8383d/src/utils/tokenListUtils.ts#L182) also work? And if so, may have to decorate that callback with network too.

priyajeet (approved):

src/routes/monad/consts.ts
kev1n-peters: The rationale for baking the allowlist into Connect is to ensure that only ETH and MON are supported via the Monad Bridge route for every integrator. This way, not every integrator needs to set a isTokenSupportedHandler.This allowlist should have been implemented at the contract level, but it isn’t so we have to do it at the frontend level.

PR #3929, opened 2025-11-18
feat: add transaction history chain filtering

kev1n-peters:
Add support for filtering transaction history by specific chains via the `transactionHistoryChains` UI config option.

Wormholescan supports filtering at the API level. For Mayan and LiFi, we filter at the client level for simplicity and, as a result, only support limited history which is fine for now.

Addresses BRI-122

### DIFF
--- a/src/hooks/useTransactionHistoryLiFi.ts
+++ b/src/hooks/useTransactionHistoryLiFi.ts
@@ -1,5 +1,6 @@
 import { useCallback, useEffect, useState } from 'react';
 import { amount as sdkAmount } from '@wormhole-foundation/sdk';
+import type { Chain } from '@wormhole-foundation/sdk';
 import type { ChainId as LifiChainId } from '@lifi/sdk';
 
 import config from 'config';
@@ -60,6 +61,7 @@ type Props = {
   address: string;
   page?: number;
   pageSize?: number;
+  chains?: Chain[];
 };
 
 const ONE_WEEK = 7 * 24 * 60 * 60; // 1 week per page in seconds
@@ -80,7 +82,7 @@ const useTransactionHistoryLiFi = (
   const [isFetching, setIsFetching] = useState(false);
   const [hasMore, setHasMore] = useState(true);
 
-  const { address, page = 0, pageSize = 30 } = props;
+  const { address, page = 0, pageSize = 30, chains } = props;
 
   const resetTransactions = () => {
     setTransactions((current) => (current?.length === 0 ? current : []));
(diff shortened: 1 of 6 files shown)


kev1n-peters:
Addresses BRI-122

src/hooks/useTransactionHistoryWHScan.ts
```diff
@@ -442,10 +449,12 @@ const useTransactionHistoryWHScan = (
       setIsFetching(true);
 
       try {
-        const res = await fetch(
-          `${config.wormholeApi}api/v1/operations?address=${address}&page=${page}&pageSize=${pageSize}`,
-          { headers },
-        );
+        let url = `${config.wormholeApi}api/v1/operations?address=${address}&page=${page}&pageSize=${pageSize}`;
+        if (chainIds && chainIds.length > 0) {
+          url += `&includesChain=${chainIds.join(',')}`;
```
kev1n-peters, line 454: This is not in production yet

src/hooks/useTransactionHistoryLiFi.ts
```diff
@@ -263,8 +278,12 @@ const useTransactionHistoryLiFi = (
               return parsedTxs;
             });
 
-            // LiFi returns max 1000 results, if we get less than pageSize, no more data
-            if (resData.length < pageSize) {
+            // If filtering by chain client-side, disable pagination since we can't
```
emreboga, line 281: So we won't have pagination for lifi transactions? How is it going to work when merging with others transactions from Mayan or WHScan?

src/hooks/useTransactionHistoryMayan.ts
```diff
@@ -194,7 +207,11 @@ const useTransactionHistoryMayan = (
               });
             }
 
-            if (resData?.length < limit) {
+            // If filtering by chain client-side, disable pagination since we can't
```
emreboga, line 210: Does this mean when filtering is on we can only see the first page in tx history from all sources (mayan, whscan, lifi)?

emreboga (approved):

kev1n-peters:
We currently fetch only the first page of LiFi and Mayan transactions when chain filtering is enabled. This isn’t ideal, but given the time constraints and the limited filtering capabilities of the LiFi and Mayan APIs, this is the most pragmatic approach for now. Once those APIs support more robust filtering (similar to Wormholescan), we should revisit and improve this behavior.

When chain filtering is not enabled, LiFi and Mayan continue to paginate as usual. Wormholescan always supports full pagination regardless.

We decided not to display any warnings to users about this for the moment.

PR #3931, opened 2025-11-19
fix: more consistent layouts

evan-gray:
I tried not to go too far with styling chnges, but wanted a more consistent layout and appearance to maintain the "widget" style. Here are some before and after pics against a light blue background to make it obvious.

# Before
image
image
image

# After
image
image
image
image


src/views/v3/Redeem/index.tsx
```diff
@@ -397,20 +392,11 @@ function Redeem() {
       statusText = `Ready to claim on ${toChain}`;
     }
 
-    return (
-      <Typography
-        data-testid="redeem-view-status-header"
```
kev1n-peters, line 402: I think removing `data-testid` will break the e2e tests which look for this in the DOM (if you do a search all for `redeem-view-status-header` you can see)

kev1n-peters (approved):

PR #3934, opened 2025-11-19
feat: MTT tx history

kev1n-peters:


src/hooks/useTransactionHistoryWHScan.ts
```diff
@@ -301,6 +301,16 @@ const useTransactionHistoryWHScan = (
     [parseSingleTx],
   );
 
+  // Parser for Multi Token NTT transactions (appId === MTT)
```
priyajeet, line 304: Curious, do these ever require customization and why not adjust when needed? Seems we are just duplicating references/callbacks here and above and below.

priyajeet (approved):

src/hooks/useTransactionHistoryWHScan.ts
kev1n-peters: I'm not sure why these are duplicated, they don't need to be. I just followed the pattern instead of refactoring for now.

PR #3943, opened 2025-11-21
feat: Deduct a fixed amount for gas token on Max button click

emreboga:
- Gas reserve mechanism to prevent users from transferring their entire gas token balance, ensuring they retain enough to complete transactions.
- Refactored percent buttons into a standalone component and added unit test coverage for gas reserve functionality and percent button behavior.

Please see the loom for the new behavior.
https://www.loom.com/share/2f20e803af594bfa8bc9c192dd5b9022

### DIFF
--- a/src/utils/gasReserve.ts
+++ b/src/utils/gasReserve.ts
@@ -1,43 +1,8 @@
 import type { Chain } from '@wormhole-foundation/sdk';
 import { amount as sdkAmount } from '@wormhole-foundation/sdk';
 
+import config from 'config';
 import { getGasToken } from 'utils';
-
-/**
- * Gas reserve amounts to keep when clicking "Max" on gas tokens.
- * These reserves ensure users have enough gas to complete transactions.
- *
- * Values provided by team based on typical transaction costs:
- * - Ethereum: 0.01 ETH ($30 notional)
- * - L2s (Base, Op, Arb): 0.001 ETH ($3 notional)
- * - Other EVM L1s or cheap Move L1s: 0.01 native token
- * - Solana: 0.01 SOL ($1.3 notional)
- */
-const GAS_RESERVES: Partial<Record<Chain, string>> = {
-  // Ethereum mainnet - higher reserve due to higher gas costs
-  Ethereum: '0.01',
-
-  // L2s - lower reserve due to cheaper gas
-  Base: '0.001',
-  Optimism: '0.001',
-  Arbitrum: '0.001',
-
-  // Other EVM L1s - standard reserve
-  Bsc: '0.01',
-  Avalanche: '0.01',
-  Polygon: '0.01',
-  Fantom: '0.01',
-  Celo: '0.01',
-  Moonbeam: '0.01',
-  Monad: '0.01',
-
-  // Solana
-  Solana: '0.01',
-
-  // Move chains
-  Sui: '0.01',
-  Aptos: '0.01',
-};
 
 /**
  * Get the gas reserve amount for a given chain.
@@ -47,7 +12,9 @@ const GAS_RESERVES: Partial<Record<Chain, string>> = {
  * @returns The amount to reserve, or undefined if no reserve configured
  */
 export function getGasReserve(chain: Chain): sdkAmount.Amount | undefined {
-  const reserve = GAS_RESERVES[chain];
+  const chainConfig = config.chains[chain];
+  const reserve = chainConfig?.gasReserve;
+
   if (!reserve) {
     return undefined;
   }
--- a/src/utils/gasReserve.ts
+++ b/src/utils/gasReserve.ts
@@ -29,6 +29,7 @@ const GAS_RESERVES: Partial<Record<Chain, string>> = {
   Fantom: '0.01',
   Celo: '0.01',
   Moonbeam: '0.01',
+  Monad: '0.01',
 
   // Solana
   Solana: '0.01',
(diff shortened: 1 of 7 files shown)


src/utils/gasReserve.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Get the gas reserve amount for a given chain.
+ * Returns undefined if no reserve is configured for the chain.
+ *
+ * @param chain - The source chain
+ * @param decimals - The token decimals
+ * @returns The amount to reserve, or undefined if no reserve configured
+ */
+export function getGasReserve(
+  chain: Chain,
+  decimals: number,
```
emreboga: Just realized, we may not need this as we can get the gas token from chain and get its decimals.

src/views/v3/Bridge/AssetPicker/PercentButtons.tsx
```diff
@@ -0,0 +1,243 @@
...
+const GAS_RESERVE_INFO_MESSAGE =
+  'A small amount of the network token balance is reserved to cover the network cost of this transaction.';
+
+function PercentButtons(props: Props) {
+  const theme: any = useTheme();
+  const { route: selectedRoute } = useSelector(
+    (state: RootState) => state.transferInput,
+  );
+  const { sourceToken, destToken } = useGetTokens();
+  const [showInsufficientTooltip, setShowInsufficientTooltip] = useState<
```
emreboga, line 44: Could have named this var better. Maybe `showInsufficientGasTooltip`?

src/utils/gasReserve.ts
kev1n-peters: There is a `getGasToken()` method that returns `Token` with decimals property

src/utils/gasReserve.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Gas reserve amounts to keep when clicking "Max" on gas tokens.
+ * These reserves ensure users have enough gas to complete transactions.
+ *
+ * Values provided by team based on typical transaction costs:
+ * - Ethereum: 0.01 ETH ($30 notional)
+ * - L2s (Base, Op, Arb): 0.001 ETH ($3 notional)
+ * - Other EVM L1s or cheap Move L1s: 0.01 native token
+ * - Solana: 0.01 SOL ($1.3 notional)
+ */
+const GAS_RESERVES: Partial<Record<Chain, string>> = {
```
kev1n-peters: Missing some chains like Monad

src/views/v3/Bridge/AssetPicker/PercentButtons.tsx
```diff
@@ -0,0 +1,243 @@
...
+    [theme],
+  );
+
+  // Check if the selected token is the gas token for the chain
+  const isGasToken = useMemo(() => {
+    if (!sourceToken || !props.chain) {
+      return false;
+    }
+    try {
+      const gasToken = getGasToken(props.chain);
```
kev1n-peters, line 73: How can isSameToken or isSameToken throw

kev1n-peters:
I wonder if we should remove the percent buttons at some point and just use the toggle like the executor route does. We should probably even remove the AutomaticTokenBridge route at some point (the only route that actually uses the percent buttons for gas dropoff, and this route isn't even used much anymore (except for Sui I think which is going to have executor token bridge support soon I think)

src/utils/gasReserve.ts
emreboga: Yup, used that one in the utility function instead.

src/views/v3/Bridge/AssetPicker/PercentButtons.tsx
emreboga: It's getGasToken, not isSameToken that throws.

src/utils/gasReserve.ts
kev1n-peters: Do you think it would be better to put a `gasReserve` property on `ChainConfig` so we don't forget to update it when new chains are added? https://github.com/wormhole-foundation/wormhole-connect/blob/a2f9e7b1342b8c8012b3f8dba32d7fe0903a5fd5/src/config/types.ts#L229-L236

src/views/v3/Bridge/AssetPicker/PercentButtons.tsx
kev1n-peters: Do you know which chain this throws for? Each chain should have its gas token configured

src/utils/gasReserve.ts
emreboga: That's a good point! I was going to add that to the doc about adding a new chain but I think this is a better option.

src/views/v3/Bridge/AssetPicker/PercentButtons.tsx
emreboga: Fantom :) That's why I added that to the test cases. Is that a bug?
```
    it('returns undefined for Fantom (gas token not configured)', () => {
      const reserve = getGasReserve('Fantom');

      // Fantom has a reserve configured but no gas token in config, so should return undefined
      expect(reserve).toBeUndefined();
    });
```

src/utils/gasReserve.ts
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3943/commits/f1b64592289171a533fd5f7b590c3150a8f8fd56

kev1n-peters (approved):

PR #3952, opened 2025-11-24
feat: added routeSortPriority config

kev1n-peters:


src/hooks/useSortedRoutesWithQuotes.ts
```diff
@@ -106,6 +106,8 @@ export const useSortedRoutesWithQuotes = ({
 
   // Only routes with quotes are sorted.
   const sortedRoutesWithQuotes = useMemo(() => {
+    const sortPriority = config.ui.routeSortPriority || 'fastest';
```
kev1n-peters, line 109: The previous behavior was to sort by fastest, so we default to fastest when the config is undefined.

evan-gray (approved):

evan-gray (approved):
confirmed the `cheapest` logic is the same as `fastest` just in a different order and the `fastest` logic remains unchanged.

PR #3953, opened 2025-11-24
feat: remove auto connect destination wallet

kev1n-peters:


evan-gray (approved):
confirmed this appears to make me connect my destination wallet every time I refresh

PR #3955, opened 2025-11-24
feat: better slippage exceeded error message

kev1n-peters:
actual exception message: `error completing transfer Error: Simulation failed: {"InstructionError":[0,{"Custom":6001}]}`
image


evan-gray (approved):
hmm... this is probably safe since it's applied only to the mayan same chain swaps. unfortunately every custom error often starts at 6000, so it's not like _all_ error 6001 are this error, only from the Jupiter program, but this might be good enough for the time being.

PR #3957, opened 2025-11-24
feat: expose lifi routeTimingStrategies via config

kev1n-peters:


src/routes/lifi/LiFiRoute.ts
```diff
@@ -198,6 +198,9 @@ export class LiFiRoute<N extends Network>
       integrator,
       referrer: params.options.referrer,
       fee: feePercent,
+      // Wait up to 500ms for quotes
+      // https://docs.li.fi/guides/integration-tips/latency#selecting-timing-strategies
+      routeTimingStrategies: ['minWaitTime-500-0-0'],
```
evan-gray: just to be clear, this is a time-limited return (timeout) for 500ms. https://docs.li.fi/guides/integration-tips/latency#timing-strategy-examples

should this be something the Connect integrator can control? I would be nervous about hardcoding this behavior if performance degrades for certain chains or routes important to a particular integrator.

evan-gray (dismissed):

src/routes/lifi/LiFiRoute.ts
kev1n-peters: We can expose it as a config, that probably makes more sense

src/routes/lifi/LiFiRoute.ts
```diff
@@ -198,6 +198,7 @@ export class LiFiRoute<N extends Network>
       integrator,
       referrer: params.options.referrer,
       fee: feePercent,
+      routeTimingStrategies: this.config?.routeTimingStrategies,
```
emreboga, line 201: It omits routeTimingStrategies prop, if the value of `this.config?.routeTimingStrategies` returns undefined, right?

kev1n-peters: Yes

emreboga (approved):

PR #3958, opened 2025-11-24
fix: remove error logging when source and destination tokens are native

priyajeet:


src/config/utils.ts
```diff
@@ -155,7 +155,10 @@ export const validateDefaults = (
   }
 
   if (defaults.source?.token && defaults.destination?.token) {
-    if (defaults.source.token === defaults.destination.token) {
+    if (
+      defaults.source.token === defaults.destination.token &&
+      defaults.source.token !== 'native'
```
emreboga: Can we use?
`import { isNative } from '@wormhole-foundation/sdk';`

emreboga (approved):

PR #3968, opened 2025-12-01
feat: auto connect destination wallet

priyajeet:


kev1n-peters (commented):
I wonder if we could just use the `useConnectToLastUsedWallet` hook to achieve what this PR seeks to accomplish. In that hook, we could do:
```typescript
if (chainToPlatform(sourceChain) === chainToPlatform(destChain)) {
  await connectWallet(destChain, TransferWallet.RECEIVING, true);
}
```

src/contexts/wallet/WalletProvider.tsx
```diff
@@ -118,15 +127,8 @@ function WalletProvider({
   }, [dispatch, walletProvider]);
 
   const disconnectWallet = useCallback(
-    async (chain: Chain, type: TransferWallet): Promise<void> => {
```
priyajeet, line 121: I moved disconnect click in the UI to just call the provider and do disconnections. Instead of calling wallet.disconnect() and wait for an async "disconnect" event since that was causing issues.

src/contexts/wallet/WalletProvider.tsx
```diff
@@ -52,12 +52,21 @@ function WalletProvider({
...
 
         if (shouldDisconnect) {
-          wallet.disconnect();
+          walletProvider.disconnectWallet(chain, type);
         }
       };
 
       const handleDisconnect = () => {
-        dispatch(clearWallet(type));
+        if (wallet === walletProvider.getWallet(chain, type)) {
```
priyajeet, line 61: What likely is still broken is that if kill connections from the wallet app itself, it doesn't clear the UI. But we need to guard the clear wallet below somehow. Any better checks so that I only listen to disconnects from the app and not our manual one?

src/hooks/useConnectToLastUsedWallet.ts
```diff
@@ -1,38 +1,161 @@
...
+
+  // Check if both source and destination swapped
+  const sourceChanged = sourceChain !== prevSourceChainRef.current;
+  const destChanged = destChain !== prevDestChainRef.current;
+
+  const chainsSwapped =
+    sourceChanged &&
+    destChanged &&
+    sourceChain === prevDestChainRef.current &&
+    destChain === prevSourceChainRef.current;
```
priyajeet: This file essentially has 5 side effects doing similar stuff 
1. when source chain changes (connect last source wallet)
3. when dest chain changes (connect last dest wallet)
4. when source wallet changes (connect dest wallet)
5. when destination wallet changes (connect source wallet)
6. when source and destination swap (just swap)

src/utils/wallet/InternalWalletProvider.ts
```diff
@@ -31,48 +31,56 @@ let pendingConnect:
...
+  if (!existing && localStorageKey && walletName !== ReadOnlyWallet.NAME) {
+    localStorage.setItem(localStorageKey, walletName);
   }
+}
 
-  const localStorageKey = getLastUsedWalletKey(chain);
+function removeWalletfromLocalStorage(chain: Chain, type: TransferWallet) {
+  const localStorageKeyForType = getLastUsedWalletKey(chain, type);
 
-  const handleDisconnect = () => {
```
priyajeet, line 58: Closure issue when swapping wallets since these handlers still reference prior type

src/views/v3/Redeem/index.tsx
```diff
@@ -860,6 +861,7 @@ function Redeem() {
         <Button
           variant="primary"
           onClick={() => {
+            dispatch(clearTransfer());
```
priyajeet: Bug that surfaced but exists today, when going from redeem page to bridge page by clicking new transaction button, the redux state was stale.

src/contexts/wallet/WalletProvider.tsx
kev1n-peters: Would passing some metadata to handleDisconnect indicating where the disconnect was initiated from (our UI vs wallet sdk) be a possibility?

priyajeet: Yea that can help

src/hooks/useConnectToLastUsedWallet.ts
```diff
@@ -1,38 +1,161 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import type { Chain } from '@wormhole-foundation/sdk';
 import { TransferWallet } from 'utils/wallet';
 import useWalletProvider from 'hooks/useWalletProvider';
+import type { WalletData } from 'store/wallet';
 
-export const useConnectToLastUsedWallet = (
+function connectToChain(
```
kyleleighton: Any reason to not just make this a hook? It implements its own lifecycle logic, so I don't see the benefit in this case.

src/hooks/useConnectToLastUsedWallet.ts
```diff
@@ -1,38 +1,161 @@
...
+    sourceChain === prevDestChainRef.current &&
+    destChain === prevSourceChainRef.current;
 
+  // If chains are swapping we will not be reconnecting any wallets
+  // as the wallet provider is just swapping references
+  if (chainsSwapped) {
+    isSwappingRef.current = true;
+  }
+
+  // Effect runs when source chain changes but its not a swap
```
bradleyplaydon: So many useEffects to all call the same thing, was previous method with conditions not preferred?

src/utils/wallet/InternalWalletProvider.ts
```diff
@@ -31,48 +31,56 @@ let pendingConnect:
...
   };
   receiving?: {
+    chain: Chain;
     wallet: Wallet;
-    disconnectHandler: () => void;
   };
 } = {};
 
-function setWalletConnection(
+function setWalletInLocalStorage(
```
bradleyplaydon, line 43: much better haha

src/views/v3/Redeem/index.tsx
bradleyplaydon: Should this not be done by an effect?

bradleyplaydon (dismissed):

src/hooks/useConnectToLastUsedWallet.ts
```diff
@@ -1,38 +1,161 @@
...
+  const chainsSwapped =
+    sourceChanged &&
+    destChanged &&
+    sourceChain === prevDestChainRef.current &&
+    destChain === prevSourceChainRef.current;
 
+  // If chains are swapping we will not be reconnecting any wallets
+  // as the wallet provider is just swapping references
+  if (chainsSwapped) {
+    isSwappingRef.current = true;
```
kyleleighton: Should we mutate the `ref` during render? Can we move it into into an effect?

kyleleighton (dismissed):

src/contexts/wallet/WalletProvider.tsx
```diff
@@ -118,32 +128,32 @@ function WalletProvider({
...
-      }
+    (chain: Chain, type: TransferWallet) => {
+      walletProvider.disconnectWallet(chain, type);
       dispatch(clearWallet(type));
     },
     [dispatch, walletProvider],
   );
 
+  const clearWallets = useCallback(() => {
+    walletProvider.clearWallets();
```
priyajeet, line 139: We have this unfortunate case of that state about wallets is managed in redux but also in wallet provider and we weren't syncing before. This is to kind of sync it.

priyajeet: This is also for a clearing scenario and not particularly disconnections. Things need to be cleared when user is navigating between bridging <---> redeem as they are diff views and have their own isolated wallet connections.

src/AppRouter.tsx
```diff
@@ -37,16 +38,16 @@ const AppRouterContent = () => {
     const bridgeRoute = 'bridge';
     // reset redeem state on leave
     if (prevRoute === redeemRoute && route !== redeemRoute) {
+      clearWallets();
       dispatch(clearRedeem());
-      dispatch(clearWallets());
```
priyajeet, line 41: This was the prior mentioned state not being synced with wallets when going between views.

src/utils/wallet/InternalWalletProvider.ts
```diff
@@ -31,52 +31,56 @@ let pendingConnect:
...
 
-function getWalletConnection(type: TransferWallet): Wallet | undefined {
-  return walletConnections[type]?.wallet;
+function setWalletConnection(
+  chain: Chain,
+  type: TransferWallet,
+  wallet: Wallet,
+): void {
+  walletConnections[type] = { chain, wallet };
+  setWalletInLocalStorage(chain, type, wallet, true);
```
priyajeet, line 76: When setting wallet in local storage, also set the other side for auto-connection

src/utils/wallet/InternalWalletProvider.ts
```diff
@@ -180,8 +187,42 @@ async function connectWallet(
...
+  const connection = walletConnections[type];
+
+  if (!connection || connection.chain !== chain) {
+    return;
+  }
+
+  walletConnections[type] = undefined;
+  removeWalletfromLocalStorage(chain, type);
+
+  const otherType =
```
priyajeet, line 200: opposite of set wallet connection above, remove the other end also since we are requesting a disconnect and ideally wallets are page scoped.

src/hooks/useConnectToLastUsedWallet.ts
```diff
@@ -1,38 +1,111 @@
...
+    prevDestChainRef.current = destChain;
+    prevSendingWalletRef.current = sendingWallet;
+    prevReceivingWalletRef.current = receivingWallet;
+
+    if (hasChainsSwapped) {
+      swapWallets();
+      return;
+    }
+
+    // Connect to source chain if:
```
priyajeet: typo `Connect to source chain` should read `Connect to source wallet`

src/hooks/useConnectToLastUsedWallet.ts
```diff
@@ -1,38 +1,111 @@
...
+    // 1. Source chain changed, OR
+    // 2. Receiving wallet connected but no sending wallet (auto-connect other side)
+    const shouldConnectSource =
+      hasSourceChainChanged ||
+      (hasReceivingWalletChanged &&
+        receivingWallet?.address &&
+        !sendingWallet?.address);
+
+    // Connect to destination chain if:
```
priyajeet: same typo

src/contexts/wallet/WalletProvider.tsx
priyajeet: future, will backlog it, not a big enough use case where people are actively going into their apps and disconnecting often, and even when they do they are just leaving the app

src/utils/wallet/types.ts
```diff
@@ -68,6 +68,20 @@ export interface WormholeConnectWalletProvider {
     autoConnect?: boolean,
   ): Promise<Wallet | null>;
 
+  /**
+   * Disconnect a wallet connection.
+   *
+   * @param chain - The chain to connect to
+   * @param type - Whether this is for sending or receiving
+   * @returns Promise resolving to the connected wallet, or null if connection failed/was cancelled
```
kev1n-peters: This returns `void` now

kev1n-peters (approved):

PR #3969, opened 2025-12-01
feat: Emitting swap.input event when user swaps assets

emreboga:
This will be used by the integrator/host app to adjust any local state when source/destination assets are switched.

### DIFF
--- a/src/views/v3/Bridge/SwapInputs/index.tsx
+++ b/src/views/v3/Bridge/SwapInputs/index.tsx
@@ -9,6 +9,8 @@ import type { RootState } from 'store';
 import useWalletProvider from 'hooks/useWalletProvider';
 import { setAmount, swapInputs } from 'store/transferInput';
 import { setToNativeToken } from 'store/relay';
+import config from 'config';
+import { UserActions } from 'telemetry/types';
 
 function SwapInputs() {
   const dispatch = useDispatch();
@@ -16,9 +18,8 @@ function SwapInputs() {
   const theme: any = useTheme();
   const [rotateAnimation, setRotateAnimation] = useState('');
 
-  const { isTransactionInProgress, fromChain, toChain } = useSelector(
-    (state: RootState) => state.transferInput,
-  );
+  const { isTransactionInProgress, fromChain, toChain, token, destToken } =
+    useSelector((state: RootState) => state.transferInput);
 
   const styles = useMemo(
     () => ({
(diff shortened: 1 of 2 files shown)


src/views/v3/Bridge/SwapInputs/index.tsx
```diff
@@ -72,6 +73,14 @@ function SwapInputs() {
 
   const canSwap = !isTransactionInProgress && fromChain && toChain;
 
+  // Helper function to emit user action events
```
priyajeet: See https://github.com/wormhole-foundation/wormhole-connect/blob/development/src/telemetry/utils.ts

src/views/v3/Bridge/SwapInputs/index.tsx
```diff
@@ -84,7 +93,34 @@ function SwapInputs() {
 
     dispatch(swapInputs());
     dispatch(setAmount(''));
-  }, [canSwap, isTransactionInProgress, dispatch, swapWallets]);
+
+    // Emit select.* events to notify about the swap action
+    emitUserAction(UserActions.SelectSrcChain, toChain);
```
priyajeet: Another question would be, instead of having it here + also having it in asset picker, can it be done once on the parent file that has access to handleSourceChainChange, handleSourceTokenChange etc

emreboga: Swap component is pretty independent, we can keep it that way:
```
{/* Swap source/destination assets button */}
{!config.ui.hideSwapInputs && <SwapInputs />}
```

src/views/v3/Bridge/SwapInputs/index.tsx
emreboga: https://github.com/wormhole-foundation/wormhole-connect/pull/3969/commits/1d009ddbf0093a854dea379ce6518d3d6e270020

src/views/v3/Bridge/SwapInputs/index.tsx
priyajeet: Followup, would having this at the redux action level be better or does that open a can of worms. Since we have actions for setting tokens and chains and even one for swapping.

emreboga: I thought about that too. swapInputs() will be triggered from a single location only and that wouldn't make much of a difference tbh. Besides, I'd leave anything but state manipulation out of reducer functions.

priyajeet (approved):

src/views/v3/Bridge/SwapInputs/index.tsx
emreboga: I also thought maybe a middleware to handle this action outside of reducers, but then it'd be an overkill I think.

PR #3973, opened 2025-12-04
feat: add ConfigContext with Zustand store for instance-scoped configuration

DBosley:
## Summary

- Introduces React Context pattern (`ConfigContext`, `useConfig()` hook) for config access
- Adds Zustand store for selector-based subscriptions (`useConfigSelector()`) to enable granular re-renders
- Enables multiple WormholeConnect instances on the same page with isolated state
- Backwards compatible: singleton remains functional via legacy bridge during incremental migration

## Changes

| File | Change |
|------|--------|
| `src/contexts/ConfigContext.tsx` | ConfigProvider with Zustand integration, useConfig/useSetConfig hooks |
| `src/store/configStore.ts` | Zustand store with useConfigSelector/useConfigValue hooks |
| `src/config/index.ts` | Legacy bridge (updateLegacyStore) for setConfig() compatibility |
| `src/WormholeConnect.tsx` | Wraps children with ConfigProvider |
| `src/AppRouter.tsx` | Example migration to useConfig() |
| `package.json` | Add zustand as dev + peer dependency |

## Test plan

- [x] Build passes: `npm run build:lib`
- [x] All 425 tests pass: `npm run test:unit`
- [x] New tests verify Zustand store functionality (18 tests)
- [x] New tests verify ConfigContext integration (17 tests)
- [x] Pre-commit hooks pass

### Selector-based subscriptions
```tsx
// Only re-renders when network changes
const network = useConfigSelector((c) => c.network);

// Only re-renders when chains change  
const chains = useConfigSelector((c) => c.chainsArr);
```

### Multi-instance support
Each `<WormholeConnect>` widget gets its own isolated Zustand store.

### Legacy compatibility
`setConfig()` calls sync to the active Zustand store via `updateLegacyStore()`.

## Migration path

This PR establishes the foundation. Subsequent PRs will:
1. Migrate React components to `useConfigSelector()` for optimized re-renders
2. Add config parameter to utility functions
3. Remove singleton entirely


src/store/configStore.ts
```diff
@@ -0,0 +1,138 @@
...
+  return createStore<ConfigState>()(() => ({
+    config: initialConfig,
+  }));
+}
+
+/**
+ * Context for providing the store instance to the component tree.
+ * Components use this to access the correct store for their WormholeConnect instance.
+ */
+export const ConfigStoreContext = React.createContext<ConfigStore | null>(null);
```
priyajeet, line 37: Needed anymore?

src/store/configStore.ts
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Hook to access the config store from Context.
+ * @throws Error if used outside ConfigProvider
+ */
+function useConfigStore(): ConfigStore {
```
priyajeet, line 43: Not needed either

src/store/configStore.ts
```diff
@@ -0,0 +1,138 @@
...
+ * const chains = useConfigSelector((c) => c.chainsArr);
+ *
+ * // Select multiple values (re-renders when either changes)
+ * const { network, isMainnet } = useConfigSelector((c) => ({
+ *   network: c.network,
+ *   isMainnet: c.isMainnet,
+ * }));
+ * ```
+ */
+export function useConfigSelector<T>(
```
priyajeet, line 76: Can just be useConfig() inside `useConfig.ts`.
Consumers will just do `const config = useConfig();`

src/store/configStore.ts
```diff
@@ -0,0 +1,138 @@
...
+export function useConfigValue(): InternalConfig<Network> {
+  const store = useConfigStore();
+  return useStore(store, (state) => state.config);
+}
+
+/**
+ * Hook to get the store's setState function for updating config.
+ * Used internally by ConfigProvider.
+ */
+export function useConfigStoreApi(): ConfigStore {
```
priyajeet, line 98: Not sure what is this doing

src/store/configStore.ts
```diff
@@ -0,0 +1,138 @@
...
+ * Creates a new config store instance.
+ *
+ * Each WormholeConnect instance gets its own store, enabling multiple
+ * independent widgets on the same page.
+ */
+export function createConfigStore(
+  initialConfig: InternalConfig<Network>,
+): ConfigStore {
+  return createStore<ConfigState>()(() => ({
+    config: initialConfig,
```
priyajeet, line 29: Should break this down to different configs, so that consumers can react to specific changes on sub configs.

src/store/configStore.ts
```diff
@@ -0,0 +1,138 @@
...
+  return useStore(store, (state) => selector(state.config));
+}
+
+/**
+ * Hook to get the full config object.
+ *
+ * Prefer useConfigSelector() for better performance - this hook will
+ * re-render on ANY config change.
+ */
+export function useConfigValue(): InternalConfig<Network> {
```
priyajeet, line 89: duplicate, but instead might make sense to have useUIConfig() in useUIConfig.ts, useTokensConfig() in useTokensConfig.ts etc, each returns that slice of the state.

src/store/configStore.ts
priyajeet: Probably should also have setters for the config, so that the place where we build the config calls this.

src/store/configStore.ts
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Unregister the legacy store when ConfigProvider unmounts.
+ */
+export function unregisterLegacyStore(store: ConfigStore): void {
+  if (legacyActiveStore === store) {
+    legacyActiveStore = null;
+  }
+}
+
+/**
```
priyajeet, line 128: I think both the register and unregister can likely move inside the main store's methods? When you call the main store's setXXX method, it should set itself and additionally set legacy stuff.

PR #3993, opened 2025-12-23
feat: LiFi route request quote latency improvements

kev1n-peters:
Use /advanced/routes instead of the /quote endpoint, which should reduce request latency.

From LiFi:

The routeTimingStrategies setting is not a hard timeout. It only controls how long we wait for enough routes from providers before picking the best one. The /quote endpoint also builds the stepTransaction, which isn’t governed by that timing strategy, so total response time can still end up in the 2–4s range you’re seeing for these routes.

If you need tighter control over latency, we’d recommend calling /routes with your desired timing strategy, selecting a route as soon as you’re happy with one, and then calling /stepTransaction separately when you’re ready to execute. That lets you enforce your own 900ms cutoff on the client side if no acceptable route has arrived yet.

Addresses BRI-190

### DIFF
--- a/src/routes/lifi/LiFiRoute.ts
+++ b/src/routes/lifi/LiFiRoute.ts
@@ -207,7 +207,10 @@ export class LiFiRoute<N extends Network>
     };
 
     // Add timing strategies if configured
-    if (this.config?.routeTimingStrategies) {
+    if (
+      this.config?.routeTimingStrategies &&
+      this.config.routeTimingStrategies.length > 0
+    ) {
       routesRequest.options!.timing = {
         routeTimingStrategies:
           this.config.routeTimingStrategies.map(parseTimingStrategy),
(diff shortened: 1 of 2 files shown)


src/routes/lifi/LiFiRoute.ts
```diff
@@ -183,39 +184,73 @@ export class LiFiRoute<N extends Network>
...
-    if (normalizedParams.exchanges?.allow)
-      quoteRequest.allowExchanges = normalizedParams.exchanges.allow;
-    if (normalizedParams.exchanges?.deny)
-      quoteRequest.denyExchanges = normalizedParams.exchanges.deny;
-    if (normalizedParams.exchanges?.prefer)
-      quoteRequest.preferExchanges = normalizedParams.exchanges.prefer;
-
-    return getQuote(quoteRequest);
+    // Add timing strategies if configured
+    if (this.config?.routeTimingStrategies) {
```
emreboga: an empty array of routeTimingStrategies would still be truthy here, which may end up with routeTimingStrategies: []. Would this be an issue down the road?

kev1n-peters: I don't think that would cause any issues, but I'll add a length check

emreboga (approved):

PR #4000, opened 2026-01-07
fix: Exclude refunded txs

emreboga:


src/hooks/useTransactionHistoryMayan.ts
```diff
@@ -117,10 +124,8 @@ const useTransactionHistoryMayan = (
       return;
     }
 
-    // Transaction is in progress when it's not completed or refunded
-    const clientStatusLC = clientStatus?.toLowerCase();
-    const inProgress =
-      clientStatusLC !== 'completed' && clientStatusLC !== 'refunded';
+    // Transaction is in progress when it's not completed
+    const inProgress = status !== 'completed';
```
emreboga, line 128: No need to check for refunded here after the new check above.

priyajeet (approved):

PR #4042, opened 2026-09-12
fix: clarify per-token balance-fetch errors after RPC fallback

### LINKED ISSUE #4040: Clarify per-token balance-fetch errors after RPC fallback
## Context

In v6.0/current, `useGetTokenBalances` performs indexed balance fetching with a per-token fallback (lines 189–257). The catch at lines 249–251 currently logs:

```text
Failed to fetch balance for token X
```

This condition is a balance-retrieval/RPC failure. It is distinct from a verified successful balance read that shows the wallet has insufficient balance.

Issues #4030 and #4031 are examples of related user-visible balance/error ambiguity only; they are not asserted to have the same root cause.

## Requested change

Please clarify the wording for per-token balance-fetch failures and include structured chain and token context plus a safely normalized underlying error.

The diagnostic must not include:

- A full RPC URL
- Secrets
- Request headers
- A wallet address

Preserve the existing per-token fallback behavior and preserve failure visibility.

## Acceptance criteria

- An unavailable balance is never represented as an insufficient balance.
- A successfully fetched low balance remains distinct from an unavailable balance.
- Tests cover the clarified behavior, structured context, safe error normalization, and sensitive-data exclusions.

devtechedge:
Fixes #4040.

The per-token fallback in `useGetTokenBalances` previously logged `Failed to fetch balance for token X`, which could read as an insufficient balance rather than a retrieval failure. The catch block now logs a clarified message, `Balance fetch failed, balance unavailable (not confirmed insufficient)`, with structured network, chain, and token context.

The underlying error is normalized through a new `normalizeBalanceFetchError` helper in `utils/errors.ts`. It bounds the message to a single line of at most 200 characters and redacts RPC URLs, EVM and Solana wallet addresses, and credential-like values. Request headers are never read, and no wallet address is included in the log.

The fallback behavior itself is unchanged: a failed fetch still leaves no balance entry, so it remains distinct from a successfully fetched low balance.

Tests:
- `utils/errors.test.ts` covers safe normalization: message extraction, whitespace collapsing, URL and address redaction, credential redaction, and truncation.
- `hooks/useGetTokenBalances.test.tsx` covers the clarified log with structured context and sensitive-data exclusions, plus the distinction between an unavailable balance and a successfully fetched zero balance.

### DIFF
--- a/src/hooks/useGetTokenBalances.test.tsx
+++ b/src/hooks/useGetTokenBalances.test.tsx
@@ -1,0 +1,181 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { renderHook, waitFor } from '@testing-library/react';
+import { createMockToken } from 'utils/testHelpers';
+import useGetTokenBalances from './useGetTokenBalances';
+
+const WALLET_ADDRESS = '0x1234567890abcdef1234567890abcdef12345678';
+const RPC_URL = 'https://ethereum-rpc.publicnode.com';
+const API_KEY = 'sk-super-secret-key';
+
+const mockToken = createMockToken({
+  chain: 'Ethereum',
+  addressString: '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48',
+  symbol: 'USDC',
+  name: 'USD Coin',
+  decimals: 6,
+  key: 'Ethereum:USDC',
+});
+
+const mockWallet = {
+  type: 'Evm' as const,
+  address: WALLET_ADDRESS,
+  currentAddress: WALLET_ADDRESS,
+  error: '',
+  name: 'Test Wallet',
+};
+
+const getBalanceMock = vi.fn();
+
+vi.mock('config', () => ({
+  default: {
+    network: 'Testnet',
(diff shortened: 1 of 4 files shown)


devtechedge:
Hello - soft nudge on this balance-fetch error clarity whenever bandwidth allows.

The per-token fallback in `useGetTokenBalances` previously logged `Failed to fetch balance for token X`, which could read as an insufficient balance rather than a retrieval failure (`Fixes #4040`).

The catch block now logs a clarified message with structured network, chain, and token context, and routes the underlying error through a new `normalizeBalanceFetchError` helper in `src/utils/errors.ts` that bounds the message, redacts RPC URLs and wallet addresses, and never reads request headers.

Fallback behavior is unchanged: a failed fetch still leaves no balance entry, so it stays distinct from a successfully fetched low or zero balance.

`src/utils/errors.test.ts` and `src/hooks/useGetTokenBalances.test.tsx` cover normalization, redaction, truncation, and the unavailable-versus-zero distinction.

Happy to tweak the log wording, tighten redaction, or adjust the structured fields if maintainers want a different shape.

No pressure - just keeping it visible since the branch is still ready.

